PR #256, opened 2021-05-20
Skye on pc off

adamjermyn:
This PR would turn Skye on by default and PC off by default.

Currently there are 11 test cases which fail on this branch, so we should get them to pass before merging. I'm opening this PR to provide a place to have that debugging conversation.

Of those 11 failures, 7 are RSP failures which I think all die in the initial model builder.

The others are `hse_riemann`, `ns_h`, `ns_he`, and `example_astero`, all of which run for a decent while before failing.

I'll start by looking at `example_astero` and gradually work my way through these. If anyone has ideas/knows what's going on with other cases help would be very much appreciated.


adamjermyn:
Thanks @warrickball for fixing the seismology cases!

The two remaining blockers are ccsn_IIp and 20M_z2m2.

@evbauer would you be able to post a log from the 20M_z2m2 run that died last night? (because it timed out it doesn't show up on test hub)

In the meantime I'll start on ccsn_IIp...

evbauer:
Here's the log for `20M_z2m2_high_rotation` from `2f1524a`

[out.txt](https://github.com/MESAHub/mesa/files/6656857/out.txt)

adamjermyn:
Great, thanks! So this is actually going to finish eventually, it’s just
taking way too long. I’ll investigate more…

-Adam

adamjermyn:
We're nearly there. The current state of things is:
1. 15M_dynamo fails, but that's the only one. It fails because some reported quantities are (very slightly) out of bounds. I posted in the testing slack channel to see if anyone knows how bad this is.
2. I turned off Skye in `20M_z2m2_high_rotation` and `ccsn_IIp`. PC is off on these two, so at high temperatures and/or densities they're just using HELM. They now pass.
3. I think the issue in (2) with Skye might be blending from Skye -> HELM in places where we need the secondary electron/positron tables, and we're working on a fix for that so we can stay on Skye with no blends there, so we may be able to turn Skye back on at some point. This is one of those 'if only we had infinite precision we wouldn't have a problem' types of issues, not a physics issue.

adamjermyn:
I've fixed 15M_dynamo and merged `main` into this branch. So if it comes back clean on a ci optional run now I'll merge it back into main.

adamjermyn:
Thanks everyone for all the hard work on this! Merged and done.

PR #275, opened 2021-06-03
Rework approach to distinguish baryonic and gravitational mass

jschwab:
The MESA option use_mass_corrections allows the user to enable a set
of corrections that distinguish between baryonic and gravitational
mass.  The mass (m) and density (rho) in MESA are baryonic.  The
corrections are due to the composition-dependent mass excess, which is
provided for each cell by chem_lib as mass_correction.  The
gravitational (or inertial) mass of a cell is then given by
dm(k)*mass_correction(k) and the gravitational mass (m_grav) is
obtained by cumulatively summing the per-cell values from the center
to the surface.

The key corrections are as follows:

The momentum equation concerns the inertial mass, so mass corrections
should modify the dm present in this equation.  As was already being
done, the gravitational acceleration should use m_grav.

The Brunt-Vaisala frequency should be consistent with the momentum
equation.  Mass corrections enter through an overall multiplicative
term and through the definition of the gravitational acceleration.
Additionally, in the Brunt B expression, there is a new term that
reflects the difference between the gradient of the inertial mass
density and the baryonic mass density.

The total energy equation should have mass corrections present in the
KE and PE terms.  These terms come from v dot the momentum equation
and so require the same corrections.

The specific PE term contains m_grav.  Since the PE is evaluated at
the start, this requires storing m_grav_start.  Some changes in
remeshing routines were required to ensure that the values necessary
to calculate the PE were present and up to date.

MESA does not change m_grav during solver iterations.  (Doing so
consistently would require adding a new equation in the solver.)  When
m_grav is updated afterwards, this results in an energy conservation
error, because the specific potential energy changes without an
offsetting term in the total energy equation.  Because this reflects a
real inconsistency, the numerical energy conservation will worsen.
This change adds an info print out showing what rel_E_err is due to
this effect when use_mass_correction = .true. .

This also currently ignores changes in the mass_correction over a
timestep so that we don't need to add in its partials.

These corrections have not yet been thoroughly tested and
independently checked.  As such, the use_mass_corrections control is
relocated to controls_dev.

### DIFF
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -7566,21 +7566,6 @@
     use_gravity_rotation_correction = .true.
 
 
-      ! use_mass_corrections
-      ! ~~~~~~~~~~~~~~~~~~~~
-
-      !   UNTESTED.   EXPERIMENTAL FEATURE.   NOT FOR GENERAL USE.
-
-      ! Gravitational vs baryonic mass corrections.
-      ! If false, then no distinction between gravitational and baryonic mass.
-      ! If true, then gravitational mass is calculated using mass corrections.
-      ! Note: may need to wait for pre-ms model to converged before turning this on.
-
-      ! ::
-
-    use_mass_corrections = .false.
-
-
       ! use_dedt_form_of_energy_eqn
       ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~
       ! always_use_dedt_form_of_energy_eqn
--- a/star/defaults/controls_dev.defaults
+++ b/star/defaults/controls_dev.defaults
@@ -237,3 +237,44 @@
 
     conv_vel_use_mlt_vc_start = .true.
 
+
+! mass corrections
+! ================
+
+
+      ! use_mass_corrections
+      ! ~~~~~~~~~~~~~~~~~~~~
+
+      ! Gravitational vs baryonic mass corrections.
+
+      ! The Lagrangian coordinate (:math:`m`) in MESA is the baryonic
+      ! mass and the density (:math:`\rho`) is the baryonic mass
+      ! density.
+
+      ! If false, then no distinction between gravitational and baryonic mass.
+      ! If true, then the gravitational mass is calculated using mass corrections
+      ! and the momentum equation, total energy equation, and Brunt are modified.
+
+      ! The variable ``mass_correction`` is the quantity you multiply
+      ! the baryonic mass density by to get the gravitational mass
+      ! density:
+
+      !     (mass density) = (baryon density) * amu * mass_correction
+
+      ! Given the mass fractions in a cell, the value of
+      ! mass_correction is provided by the ``chem`` module.
+
+      ! MESA holds m_grav fixed during the newton iterations.  This results
+      ! in an energy conservation error, because the specific potential energy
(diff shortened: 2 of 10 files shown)


adamjermyn:
Reading through the changes now, some questions:
1. It looks like `total_energy_from_fixed_m_grav` counts the energy error due to neglecting `m_grav` changes through a step. Is the reason for accumulating this that we want to throw that into the energy conservation accounting (so we don't count it towards the energy error)? Or just to be able to report this piece?
2. The variable `mass_correction` is the thing you multiply the baryonic mass by to get the gravitational mass, right?
3. I don't understand what you're doing with `qhalf`. It looks initially like it's just 'store a quad precision version of 1/2` but then you're multiplying it by `mass_correction`...
4. I double-checked the brunt corrections. Here's what you've got:
```
! add term accounting for the composition-related gradient in gravitational mass
if (s% use_mass_corrections) then            
    delta_lnMbar = log(s% mass_correction(k-1)) - log(s% mass_correction(k))            
    s% brunt_B(k) = s% brunt_B(k) - chiRho_face*delta_lnMbar/delta_lnP/chiT_face
end if
```
The expression I have is:
Screen Shot 2021-06-14 at 10 41 18 AM

So the extra term we want is the one with `dlnMbar` and that agrees with what you have. But I'm puzzled as to why there isn't an overall factor of `Mbar` out front (i.e. multiply the final answer by `Mbar`). I have a vague recollection that we talked about this and concluded it shouldn't be there but I have no notes on why.

Otherwise everything looks good to me and I'm happy to have this be merged (and to maintain it going forward).

jschwab:
Thanks for the comments.  I pushed an updated version of the patch.

> The variable mass_correction is the thing you multiply the baryonic mass by to get the gravitational mass, right?

Yes, the `chem_lib` definition is 

```         
         ! mass_correction = sum(n(i)*W(i))/sum(n(i)*A(i)) --
         ! (mass density) = (baryon density) * m_u * mass_correction
```

> I don't understand what you're doing with qhalf. It looks initially like it's just 'store a quad precision version of 1/2but then you're multiplying it by `mass_correction`...

This was some attempt to save a few multiplies and a quad-precision variable.  But there's no reason to think that the benefit of doing that outweighs unclear code, so I re-wrote this more explicitly (and `qhalf` remains equal to 1/2).

> But I'm puzzled as to why there isn't an overall factor of Mbar out front (i.e. multiply the final answer by Mbar).

Seems like we agree on B.  I also agree N2 needs a factor of Mbar out front.  But I think that happens in the calculation of `rho_P_chiT_chiRho` at https://github.com/MESAHub/mesa/pull/275/files#diff-d386b142f64051604eee9c1c08834c9da4fa99467bb753e39727d9c805e73bcdR148

>  Is the reason for accumulating this that we want to throw that into the energy conservation accounting (so we don't count it towards the energy error)? Or just to be able to report this piece?

I don't know whether `total_energy_from_fixed_m_grav` is useful.  I added it so there would be some monitoring about the size of the error induced by this term.  The way I had been thinking about it, it didn't make sense to alter the energy accounting itself.  Happy to discuss that more and when we converge on an approach, I'll describe it in the defaults so the user knows to expect these messages.

adamjermyn:
Thanks for clarifying and refactoring `qhalf`. This all looks good to me, feel free to merge :-)

PR #278, opened 2021-06-18
Catch negative luminosity in `star_utils:get_phot_info`

warrickball:
`get_phot_info` can ultimately return values from either the outermost meshpoint, interpolation inside the model, or the innermost meshpoint. The first and last option include a catch
```f90
      L = max(1d0, s% L(1)) ! don't use negative L(1)
```
but the second option (interpolation) did not, which caused an FPE error in `T_tau_gradr`. This PR fixes that by adding the same catch before returning after interpolation. The FPE situation is [no worse](https://testhub.mesastar.org/whb%2Fnegative-L-FPE/commits/89da512) and a full run of the test suite is [passing](https://testhub.mesastar.org/whb%2Fnegative-L-FPE/commits/e05468d).

### DIFF
--- a/star/private/star_utils.f90
+++ b/star/private/star_utils.f90
@@ -1155,6 +1155,7 @@
                   v = s% v(k) + (s% v(k+1) - s% v(k))*(tau_phot - tau00)/dtau
                end if
                L = s% L(k) + (s% L(k+1) - s% L(k))*(tau_phot - tau00)/dtau
+               L = max(1d0, s% L(1)) ! don't use negative L(1)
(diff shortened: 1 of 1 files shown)


adamjermyn:
This might be a bigger discussion, but do we want to be catching this with
`max` or would it be better to say 'This is an invalid L, throw `ierr`.'?

-Adam

warrickball:
I agree that raising a non-zero `ierr` would be more sensible. I'm happy to open an issue for when someone has time.

adamjermyn:
Yes please :-)

-Adam

PR #287, opened 2021-06-25
Add environment variables for multiplying mesh_delta_coeff, time_delta_coeff, and max_model_number 

earlbellinger:
Introduces three new environment variables

```
$MESA_MESH_DELTA_COEFF_FACTOR
$MESA_TIME_DELTA_COEFF_FACTOR
$MESA_MAX_MODEL_NUMBER_FACTOR
```

and multiplies the relevant parameter by that factor after reading in the inlist. 

--- 
**Edit**: Has been replaced by one unified variable, named `$MESA_TEST_SUITE_RESOLUTION_FACTOR`.


star/job/run_star_support.f90
```diff
@@ -3822,7 +3826,69 @@ subroutine add_fpe_checks(id, s, ierr)
...
+         real(dp) :: time_delta_coeff_factor = 1
+         real(dp) :: max_model_number_factor = 1
+         character(len=20) :: mesh_delta_coeff_factor_str
+         character(len=20) :: time_delta_coeff_factor_str
+         character(len=20) :: max_model_number_factor_str
+         
+         include 'formats'
+         
+         ierr = 0
+         call GET_ENVIRONMENT_VARIABLE('MESA_MESH_DELTA_COEFF_FACTOR', mesh_delta_coeff_factor_str, STATUS=status)
```
rjfarmer, line 3848: We should be checking if status /=0

rjfarmer:
We should also add the variables to star/test_suite/each_test_run so the testhub can pick up that we are doing a convergence test @wmwolf @jschwab (and testhub would need to know to display a new icon).

If there any reason to use three new variables? Why not just a flag that multiples mesh, time, and model number by X? 

We should document in docs/source/reference/env_vars.rst the new environment variables.

earlbellinger:
> If there any reason to use three new variables? Why not just a flag that multiples mesh, time, and model number by X?

I have no opinion here. I shall let you all decide your preference and implement accordingly. I just figured that someone might want the flexibility to specify the factors on a run-by-run basis, perhaps to test with even stricter tolerances on longer timescales.


> We should document in docs/source/reference/env_vars.rst the new environment variables.

OK, I will add that once people comment on the above variable discussion

star/job/run_star_support.f90
earlbellinger: OK, I've implemented it! [I had thought it might have been fine for people to specify one of the env vars without caring about the others (so for example they could increase mesh_delta without caring about the others), so I had skipped it.]

earlbellinger:
> We should also add the variables to star/test_suite/each_test_run so the testhub can pick up that we are doing a convergence test @wmwolf @jschwab (and testhub would need to know to display a new icon).

Would this go in `do_one` or perhaps `check_run`? I see the `testhub_log` calls in those functions so I could mimic that (once the env vars are finalized)

wmwolf:
I suspect changes to the time and mesh resolutions may have confounding effects on the number of models, so I could see potentially some need for separating out the different "boost" factors. This might benefit from some real-world testing to see what we might expect when we double the time and space resolution. If it seems like this just doubles the number of model numbers or less in all cases, I say just reduce to a single environment variable.

The values of these variables (or just a flag that says they're on... however that ends up making sense) should indeed be added to the testhub logs. As soon as I know what the names of the keys and what types of values they will contain are, I will add them to the testhub's database and allow them to be scooped up with submissions.

orlox:
For simplicity I'd vote for a unified environment variable. A well-behaved test case should normally behave better in all respects with increased spatial/temporal resolution.

adamjermyn:
I like the idea of having a standardized way of changing the resolution in
different dimensions by fixed multiples, but I really don’t think this
should be an environment variable that’s available in standard (non test
suite) runs.

My concern here is that currently we have a nice abstraction where having
the inlists and run_star_extras (almost) guarantees reproducibility across
platforms. I say almost just because there might be saved starting models
or rn scripts.

Introducing environment variables that modify inlist parameters breaks that
abstraction. What MESA does then depends on variables that aren’t tied to
your run directory (they’re state variables of the terminal session). We
kind of already have this problem with the “skip optional” flag in the test
suite, but that’s limited to the test suite rn scripts so users almost
never encounter it.

So my suggestion would be to either make these specific to the test suite
(via test suite extras) or make them inlist variables that modify the base
parameters by a specified relative amount. We can always supply scripts
that modify these to make user convergence studies easier…

-Adam

orlox:
This is just meant to be a simple way to do test_suite runs with boosted resolution without any inlist modifying hassle, for general convergence testing we already have time_delta_coeff and mesh_delta_coeff. I think adding any further inlist options would defeat the point. It would be nice to encapsulate this to the test_suite, but I'm not sure there is a place for that in the test_suite_extras. We would need something operating after the controls are read and that is not available as far as I can see. Rather than adding stuff to test_suite_extras to cover this I'd be more inclined to keep Earl's implementation in run_star_support, but I would vouch to have a single variable, and name it in a way that makes it clear this is for the testsuite, like TEST_SUITE_RESOLUTION_FACTOR.

annethoul:
I assume that this feature would be for test_suite cases convergence tests only, not for regular use in MESA. Or is it something that users would be encouraged to use to check that their models are converged? I am a little worried about having a one-for-all test like that because, as always, convective boundaries are not well behaved. At some point if the resolution is increased too much the results will stop converging. How much is too much will have to be tested. A "blind" convergence test  is not something that can be fully trusted in all cases. It needs a careful eye.

rjfarmer:
Yes this should only be developers. That is why i suggest we do a fixed number (with only 1 variable), its less useful for users. The point is not to have something that works out what the converged model is. The point is be to easily see which cases fail when ran at a higher resolution, and there will be many failures, when we turn this on. After we know which cases fail we can go and manually check them, I'm sure many will fail just because we have a bad stopping criteria (which is useful to know) and some may fail because they depend on some non-converged physics (that we may decide we can ignore). But at the moment we don't know which cases are well behaved, or not. Nor do we now if a previously converged test case suddenly becomes unconverged.

orlox:
And if we identify cases where a doubling of resolution is really an issue, we can disable the environment variable in the rn script as a simple workaround. I would hope this is really a very isolated issue, but we never know.

adamjermyn:
If it’s a dev only environment variable can we print a warning when it’s
on? I’d like to make it extremely hard for users to use this without
knowing what’s going on…

-Adam

earlbellinger:
Thanks everyone for your useful feedback. I have reduced it to one variable and added a warning as follows: 
``` ***
 MESA_TEST_SUITE_RESOLUTION_FACTOR set to  0.50000000000000000
 ***
 Warning: This environment variable is for testing purposes
           and should be set to 1 during normal MESA use.
 ***
 Multiplying mesh_delta_coeff and time_delta_coeff by this factor,
 and max_model_number by its inverse:

    old mesh_delta_coeff =    1.0000000000000000
    new mesh_delta_coeff =   0.50000000000000000

    old time_delta_coeff =    1.0000000000000000
    new time_delta_coeff =   0.50000000000000000

    old max_model_number =         3000
    new max_model_number =         6000
 ***
```
If this is fine then I shall update the docs accordingly and push that in this PR too.

warrickball:
Sorry I'm relatively late to the discussion but wanted to chip in while it was still active. I agree it's a good idea to twiddle the resolution knobs and see what test cases break. I'm not hugely fussed about how it's done but this isn't personally how I would do it. It strikes me as something that will probably yield a lot of results when we first do it but very few after an initial pass.

If it were up to me, I'd write a script to change the inlists appropriately, run the test suite locally with `./each_test_run_and_diff` and inspect the results with `./report`. I'm not sure we have to add code to MESA, set up (and document) an extra environment variable and implement new TestHub features.

earlbellinger:
In light of Warrick's comments, I would welcome more discussion and opinions before I proceed. 

Personally, I am rather a fan of integrating this with the TestHub. I think we will benefit from running it on a semi-regular basis (at least prior to every major release), and it will be great for us to view the results on the web rather than restricting the discussion e.g. to slack, with someone tasked with perhaps manually running it, looking through the results and presenting it each time issues arise. If we don't integrate this with the TestHub, people may be left wondering what's still broken after some progress after the initial pass, whereas with the TestHub they can poll that info whenever they want. I also see this as being quite useful in light of the ongoing test_suite overhaul, so it seems likely that we will want to run these tests semi-frequently (rather than one-off). 

About writing scripts to adjust the inlists in the test_suite one-by-one, Pablo and Rob's suggestion to do it via an env var struck me as a fairly elegant way to avoid doing just that. Setting up an env var was clearly rather straightforward and comes with essentially no risk of it going wrong, unlike a script that likely has to do large amounts of file i/o, which one could imagine breaking on a case-by-case basis (for example, if the script gets interrupted). 

However, I say all this as a person with relatively little experience running into the testing, convergence, and bug fixing frustrations that I am trying to help mitigate, so it will be useful for others to chime in here.

rjfarmer:
We have to do this via the testhub. If we don't then it falls on one person to run the code, debug, and send all the logs out to people to fix. If its in the testhub that is all automated. Someone pushes a [ci converge] commit, the clusters run this (no ones desktop will handle this, if we double time/mesh that is atleast ~4x the runtime and ~2x the memory). Then with the testhub now uploading the output files on failure means everyone can easily look at which cases fail (I do not want to run this myself and have to manually send the logs of failing test cases to people).

This will only this is useful for us, if it can be run easily and semi-regularly and that it does not fall on one person to do anything more than say 'Lets run a convergence test'.

warrickball:
I'm persuaded. I think I'm being a spanner because I personally wouldn't mind writing a script to perturb the inlist parameters, submit it to our cluster, and tell people about the results.

rjfarmer:
Nothing is stopping people from doing more complicated testing (those scripts would be good candidates to get dropped into linters/). This is just the first pass brute-force approach to the problem. Lets not bite off more than we can handle and get nothing done.

adamjermyn:
I’m convinced! (And this definitely should go through testhub!)



-Adam

earlbellinger:
OK, great. I've updated the `env_vars.rst` docs, and also added some logic to the `do_one` function in `star/test_suite/each_test_run` to write this variable to the `testhub.yml` file. 

It will log 
    ```testhub_log "resolution_factor: $MESA_TEST_SUITE_RESOLUTION_FACTOR"```
when it's set to a value other than 1, and otherwise:
    ```testhub_log "resolution_factor: 1"```

I think touching these files triggered additional reviewers. Welcome to the party ;)

star/test_suite/each_test_run
```diff
@@ -253,6 +253,12 @@ if [ "$DO_ALL" == 1 ] || [ "$curr_id" == "$id" ] ; then
     else
         testhub_log "fpe_checks: true"
     fi
+    if [ -z "$MESA_TEST_SUITE_RESOLUTION_FACTOR" ] || 
+       (( $(echo "$MESA_TEST_SUITE_RESOLUTION_FACTOR == 1" | bc -l) )); then
```
jschwab, line 257: This second part of this test seems unnecessary.  Why bother?  If the env var is set to `1.0` it seems fine to go to the other branch and write that in the YAML.  It's a presumably floating point field, so it'll all be the same when it eventually hits the testhub DB.

jschwab:
The approach you all converged on seems like a reasonable thing to try to me.

If you did want to try and keep this power away from users, I think design-wise it would fine to put the `multiply_tolerances` code and its call in test_suite_extras.  The call to test_suite_startup happens in the extras_startup hook, which is a little later than where the code is now, but I think still early enough to work as intended.  (But even if that is eventually desired, it is a simple change, so doesn't seem like a reason to block on giving this a whirl and seeing what happens.)

star/test_suite/each_test_run
earlbellinger: Sure, I can remove it. My only thought was to save potential headache if the implementer on the other side only checked that "1" was the "off" state without accounting for e.g. "1.0000" being a value that could show up there. But I agree, it's not necessary.

orlox (approved):

Issue #285, opened 2021-06-21
Standards for EOS return values


adamjermyn:
I don't know precisely what this will look like, but I'd like to propose that we figure out some standards for what the EOS is allowed to return, and say that return values that violate those conditions must raise `ierr /= 0`. This would be the job of individual EOS's to enforce, not a blend-level enforcement.

A few standards that come to mind:
1. No EOS return quantity should be NaN.
2. No negative pressures.
3. No negative sound speeds.
4. No negative adiabatic indices.

Is there anything else we want to check? We can also make a module-level test that runs a few grids of (rho, T) at different compositions as an easy way of catching these...

fxt44:
for a GIGO model, no negative or NAN input quantities - notionally T, Rho, composition. on the output side of GIGO, maybe add no negative entropies and specific heats.

adamjermyn:
I've made a new branch (`eos_better_tests`) for this. It currently implements your suggestions on the input side. Working on the output side now.

adamjermyn:
Implemented output checks in this commit:
https://testhub.mesastar.org/eos_better_tests/commits/a249bd4

These checks all throw an `ierr /= 0` and return immediately.

adamjermyn:
Here are the current output checks:
```
         do k=1,num_eos_basic_results
            if (is_bad(res(k) .or. is_bad(d_dlnd(k)) .or. is_bad(d_dlnT(k))) then
               ierr = 1
               return
            end if
            do l=1,species
               if (is_bad(d_dxa(k,l))) then
                  ierr = 1
                  return
               end if
            end do
         end do

         if (res(i_grad_ad) < 0d0 .or. res(i_chiRho) < 0d0 .or. res(i_chiT) < 0d0 &
            .or. res(i_Cp) < 0d0 .or. res(i_Cv) < 0d0 .or. res(i_gamma1) < 0d0 &
            .or. res(i_gamma2) < 0d0 .or. res(i_gamma3) < 0d0) then
            ierr = 1
            return
         end if
```

fxt44:
another option on the input side is to only consider values within a valid range - no temperatures less than 10 K or larger than 1e13 K, no densities larger than 1e12 g/cc, abundances greater than 1,  and so on. might get false positives if the validity range is made too small. i also wonder how expensive all this checking will get. maybe its an option?

adamjermyn:
Added.

[...]

PR #289, opened 2021-06-28
Add tests to input/output of eosdt.

adamjermyn:
This PR introduces new tests to the input quantities for the EOS in eosdt and the returned outputs. These tests are:
1. No EOS return quantity may be NaN.
2. No negative pressures.
3. No negative sound speeds.
4. No negative adiabatic indices.
5. No negative specific heats.
6. No negative chi's.
7. No negative or NaN compositions.
8. No negative or NaN X or Z or abar or zbar.
9. No (logRho,logT) outside the plane (logT_lo,logT_hi,logRho_lo,logRho_hi), which I've set to match the range covered by HELM (though these could get out of sync if HELM changes down the road).
10. No negative Rho or T.

Once merged this resolves issue #285.

I'm currently waiting to merge pending results from [Testhub](https://testhub.mesastar.org/eos_better_tests/commits/12dec6c).


adamjermyn:
Exciting! Lots of cases fail. As you might expect these all look like they're cases where the EOS was either failing silently or somehow ignoring bad inputs. I'll start tracking them down.

fxt44:
GIGO to the rescue (or demise)!

adamjermyn:
Some of the failures are because the root finders for e.g. eosPT go out of bounds. Presumably beforehand they'd go out of bounds and then somehow hop back in? Here's an example, from rsp_BEP:
```
 eval1_mesa_Rho_given_PT failed           0
 P   239.82657328270977     
 T   17385.095374334280     
 rho_guess   2.1459613882639316E-010
 eval1_mesa_Rho_given_PT failed           0
 P   239.82657328270977     
 T   17384.921524246398     
 rho_guess   2.1459828479855314E-010
                                                    T_1           0    5.8593662214878832D+03
                                                   Tmax           0    1.7385095374334313D+04
                                            resid_T_min           0   -1.0000000000000280D+00
                                            resid_T_max           0    5.1944688649472917-315
 eval1_mesa_Rho_given_PT failed         149
 P   239.82657328270977     
 T   17384.921524246398     
 rho_guess   2.1459828479855314E-010
 failed in do_rsp_build
 failed in build_rsp_model
 star_create_RSP_model ierr          -1
 do_load1_star ierr          -1
 do_before_evolve_loop ierr          -1
 do_before_evolve_loop ierr          -1
```

Any thoughts on how to handle that?

fxt44:
sharing two experiences here, neither direct, but perhaps relevant. one, is during solution for the nuclear statistical equilibrium where mass and charge conservation are the constraint equations. here, the 2d root finder commonly explores solutions that are wildly away from sum X = 1, yet because the resulting differences are large, the root finder abandons that search direction. two, is during solutions under a constant pressure constraint. here, the root finder may well venture off the thermodynamic manifold where there is no solution for the (P,rho) pair. again, the root finder bails on that direction as the errors are large. bottom line is that in both cases i let the root findsersdo their thing without trying to constrain intermediate results. mileage may vary in mesa.

evbauer:
Any idea why this branch is getting checksum mismatches?
https://testhub.mesastar.org/eos_better_tests/commits/9c83e09

adamjermyn:
No idea. In principle all that's changed is that it raises ierr /= 0 in
`eosDT_get` when you query with invalid inputs (and likewise when EOS
returns invalid outputs). Maybe there's an FPE that we only avoided by dint
of allowing GIGO?

-Adam

adamjermyn:
Now that the TDC merge is in I merged `main` into this branch. We're still getting checksum mismatches and lots of failures. Looks like just lots of cases of `ierr` not being caught correctly by the routines that call the EOS.

I don't have the time/energy/interest to push forward on this, so I'm going to close out this PR. I'll leave the issue open as an enhancement in case someone else wants to take it on.

PR #292, opened 2021-07-02
[ci converge] increase max model number

earlbellinger:
Instead of max_model_number / test_suite_res_factor, let's divide by res_factor twice

### DIFF
--- a/star/job/run_star_support.f90
+++ b/star/job/run_star_support.f90
@@ -3855,7 +3855,7 @@
             write(*,*) "          and should be set to 1 during normal MESA use."
             write(*,*) "***"
             write(*,*) "Multiplying mesh_delta_coeff and time_delta_coeff by this factor,"
-            write(*,*) "and max_model_number by its inverse:"
+            write(*,*) "and max_model_number by its inverse twice:"
             write(*,*) ""
             write(*,*)    "   old mesh_delta_coeff = ",   s% mesh_delta_coeff
             s% mesh_delta_coeff = test_suite_res_factor * s% mesh_delta_coeff
@@ -3866,7 +3866,7 @@
             write(*,*)    "   new time_delta_coeff = ",   s% time_delta_coeff
             write(*,*)    ""
             write(*,*)    "   old max_model_number = ",   s% max_model_number
-            s% max_model_number = s% max_model_number / test_suite_res_factor
+            s% max_model_number = s% max_model_number / test_suite_res_factor / test_suite_res_factor
             write(*,*)    "   new max_model_number = ",   s% max_model_number
             write(*,*)    ""
          end if


evbauer:
I think this is a good idea, but it looks like there's some sort of bug with the max_model_number implementation.

Here, the output claims that max_model_number has increased form 1000 to 4000, but it stops the run at 2000:
https://logs.mesastar.org/86e407a4773244fe79ea62dfc077f490e39f48c9/cannon_conroy_intel/15M_dynamo/out.txt

And before you implemented this change, the old version said that max_model_number increased from 1000 to 2000, but then stopped at 1000.
https://logs.mesastar.org/cae38db8df56aab78d997af1db6924a37f1dfbae/cannon_conroy_intel/15M_dynamo/out.txt

earlbellinger:
Hmm, are you sure? It looks like each inlist has its own max_model_number, and they are each getting appropriately multiplied. For example, the stop at 2000 has a block at the top of its inlist saying: 
```
    old max_model_number =          500
    new max_model_number =         2000
```

evbauer:
Oh, thanks! I had just looked at the beginning of the file and the end, so I didn't notice the switch to a new inlist in between. Sorry for the false alarm!

earlbellinger:
I just noticed that this is still a draft pull request from months ago, so let's tidy that up. 

Since I had to remind myself its purpose: the main idea was that we want to occasionally increase the temporal and spatial resolution of all the tests by some factor (e.g. 2) to make sure they are semi-resilient to such a change. But some tests then hit the max model number, so we want to multiply the max model number by some equivalent factors.

wmwolf:
I'm trying to make the testhub more aware of these various ci flags. I'd like to have it be able to check if we are running a convergence test and to know what the factor is. The longer-term goal is to let testhub handle telling computers what optimal commit they should test and how they should test it.

Is the idea that someone could tack on `[ci converge 2]` and then get a factor of ~2 increase in space and time resolution? Or is it just that we pick some arbitrary number on a computer-by-computer basis to scale up all the resolutions with a more generic `[ci converge]`?

Also, I note that the latest documentation in developing/test_suite.html does not mention `[ci converge]`, while 21.12.1's version does... did this get deleted at some point?

earlbellinger:
> Also, I note that the latest documentation in developing/test_suite.html does not mention [ci converge], while 21.12.1's version does... did this get deleted at some point? 

That was my fault. I added [ci converge] to the docs today but accidentally pushed to the 21.12.1 branch instead of main. I'll fix it shortly.

earlbellinger:
> Is the idea that someone could tack on [ci converge 2] and then get a factor of ~2 increase in space and time resolution? Or is it just that we pick some arbitrary number on a computer-by-computer basis to scale up all the resolutions with a more generic [ci converge]?

Just my two cents: the former is an interesting suggestion and something that could be done, but at least at the moment I think the latter is probably all that we need. The main idea with this flag is seeing which tests break when you change the temporal and spatial resolution a little bit. (As a side note, I could be wrong but I think only Evan has this implemented at the moment, and so his opinion would particularly be valued here.)

wmwolf:
Leaving it as a flag is great. I just wanted to know the intended use so the testhub knows how to scrape data. For instance, `[ci optional]` can (in theory) accept an additional integer a la `[ci optional 20]` which will only run optional inlists for the first 20 (or whatever) test cases, presumably in `star`.

Due to the unpredictability of how long these cases could take, I don't see broad adoption of the flag in our future, so I think keeping things simple is the best route.

Issue #270, opened 2021-05-28
`make_co_wd` doesn't run in the test suite, but other test cases rely on it for starting models.


adamjermyn:
I count six test cases that say their initial models come from `make_co_wd`. Grep output:
```
wd_he_shell_ignition/inlist_he_shell_ignition
9:      load_model_filename = 'co_wd_0.96M.mod' ! from make_co_wd, 13738, initial_mass = 6.0d0

wd_acc_small_dm/inlist_wd_acc_small_dm
9:      load_model_filename = 'co_wd_1.00M.mod' ! from make_co_wd, 13738, initial_mass = 6.4d0

wd_c_core_ignition/inlist_relax_mass
4:      load_model_filename = 'make_co_wd_6M.mod' ! WD from 6Msun (ZAMS) via make_co_wd

wd_stable_h_burn/inlist_wd_stable_h_burn
9:      load_model_filename = 'co_wd_1.00M.mod' ! from make_co_wd, 13738, initial_mass = 6.4d0

gyre_in_mesa_wd/inlist_pulse_wd
7:      load_model_filename = 'co_wd_0.85M.mod' ! from make_co_wd, 13738, initial_mass = 4.7d0

wd_cool_0.6M/inlist_wd_cool_0.6M
11:      ! comes from final.mod produced by make_co_wd as of version 14731
```

For starters we should maybe be a little worried that these haven't been regenerated in a long time. But more importantly, `make_co_wd` isn't run as part of the test suite, so we have no idea if it works or not.

I ran into this on the `Skye_on_PC_off` branch, where `wd_c_core_ignition` is showing a failure that I was worried was due to a starting model generated with a different EOS (I'm now less sure that this is the problem). Along the way I went to try regenerate the starting model using `make_co_wd` and couldn't. At least on that branch on commit `2c865da` it fails on an all-inlist run with the following tail:
```
        366   8.351401  1.629E+04   2.826401   3.238048   0.605815   0.001019   0.000000   0.000000   0.317390  19.734221   1384     21
-3.7561E+00   8.089088  -0.792815   2.891026   2.066940 -99.000000   0.604797   0.000000   0.765029   0.020402   5.317551     12
 4.3220E+08   6.282687   0.216866   2.449357 -99.000000  -8.102218   0.577366   0.208947   0.001936   0.003286  0.000E+00  max increase
                               rel_E_err    6.2863672357018316D-10
                       log_rel_run_E_err       -7.1783317924390762

[...]

Issue #197, opened 2021-03-10
Large number of retries in 1M_thermohaline


jschwab:
After 6cbc77584ba42b5b812e5b299476b230d7008e42 , the `1M_thermohaline` test went from 1027 steps, 0 retries to 1735 steps, 142 retries.  At present, it is sitting at 1535 steps, 283 retries.

adamjermyn:
This improved somewhat when I turned on TDC for this case, and has stayed a little bit better (now 1764 steps with 154 retries). I'm looking into whether better time-step controls can help further.

adamjermyn:
I believe commit `9ffa01f` [testhub](https://testhub.mesastar.org/TDC_on_default/commits/9ffa01f) resolves this by tightening the `dX` timestep controls.

adamjermyn:
Resolved with PR #293.

PR #293, opened 2021-07-08
Big TDC merge

adamjermyn:
This PR updates the time-dependent convection feature in a number of ways:
1. Now available as an option for `MLT_option`. Enable TDC by setting `MLT_option='TDC'`. TDC is only ever used if selected with `MLT_option='TDC'`.
2. Old TDC switching options based on time-step size are removed. All switching is done by checking whether the physical assumptions of MLT are appropriate. If they are, we use `Cox` and otherwise we use `TDC`. Note that `TDC` reduces mathematically to `Cox` in the limit where MLT's assumptions are satisfied (slowing changing structure relative to turnover/acceleration time).
3. Numerous TDC solver improvements that allow it to provide robust solutions across the test suite.
4. TDC is disabled during relaxation and re-enabled afterwards. Empirically this leads to much better behavior during relaxation.
5. TDC is disabled in the first step of a run unless the run provides sets the flag `have_mlt_vc` to `.true.`. This amounts to a promise that the contents of `mlt_vc` reflect the results from previous evolution and not just "filled with zeros because nothing was provided". This is important because filling `mlt_vc` with zeros when not physically called for will cause TDC to freak out and force tiny time steps to compensate ("you handed me a convection zone supposedly near thermal equilibrium but all the speeds are zero!!!").
6. This PR turns on TDC in most (but not all) test cases and turns it on by default in MESA/star. The set of test cases was chosen to give good coverage of cases where TDC likely matters (e.g. `dev_to_cc_20`) and cases where it's irrelevant, while still providing coverage of other MLT options.

In addition this PR contains some changes unrelated to TDC that cropped up in all the testing I did along the way:
1. Lots of FPE fixes, some in HELM and some in mlt_get_results.
2. Some HELM call structure changes to eliminate unused variables.
3. Fixed a partials bug in RTI by rewriting it in auto_diff.

At the moment after a merge from `main` this branch (`TDC_on_default`) passes on all but one test case (`dev_to_cc_12`). That test case currently has TDC disabled and it still fails for reasons that are inscrutable to me. I would greatly appreciate some help in looking into it, because I've now spent more than two weeks staring at this case and trying things, and it's the only thing blocking this merge. You can see the latest failed log [here](https://logs.mesastar.org/3d8a829a30a2b65b1faee8f156668fcc78d90f3b/cannon_conroy_intel/dev_to_cc_12/out.txt).

The failure occurs in `inlist_to_cc_header`, which is the second-to-last inlist in the test. The termination code is `max_model_number`, whereas it should be `fe_core_infall_limit`. It appears that the time-step gets driven down to ~1d-3 seconds (versus ~1d2 seconds on `main`) and the model just doesn't evolve far enough to reach `fe_core_infall_limit`. The proximate cause of this is a series of retries in quick succession starting on step 2741 of the same inlist, excerpt below:
```
 retry: delta_lgT_cntr hard limit    1824    2741
       2741   9.566228   3121.197  -6.137870  11.761932  12.000000   8.674446   0.000000   0.000000   0.280038   5.248099   1824    237
 1.4506E+02   9.566228   2.913227   7.935733  10.609541 -99.000000   3.325554   0.000000   0.000009   0.019865   2.818593      6
 6.3146E+04   8.233551   4.758432  11.761867  17.787098  -8.668121   1.938093   0.000000   0.000000   0.585900  0.189E-03         retry
                                 Fe_core        0.0000000000000000
                          fe_core_infall        0.0000000000000000
                               rel_E_err    3.0139015288970127D-10
                       log_rel_run_E_err       -5.9769953845428132
                     dt_div_max_tau_conv        0.0101890421964826

       2742   9.573473   3121.197  -6.105979  11.748492  12.000000   8.674446   0.000000   0.000000   0.280038   5.128534   1817    237
 1.4506E+02   9.573473   2.913227   7.934177  10.610756 -99.000000   3.325554   0.000000   0.000009   0.019865   2.763147      6
 6.3291E+04   8.230446   4.758432  11.748426  17.772674  -8.668121   1.938093   0.000000   0.000000   0.586684  0.189E-03          hold
                                 Fe_core        0.0000000000000000
                          fe_core_infall        0.0000000000000000
                               rel_E_err    1.2846383902085779D-10
                       log_rel_run_E_err       -5.9767539647153356
                     dt_div_max_tau_conv        0.0095885342343731

       2743   9.578471   3121.197  -6.185848  11.749375  12.000000   8.674446   0.000000   0.000000   0.280038   5.046625   1816    237
 1.1605E+02   9.578471   2.913227   7.932919  10.608423 -99.000000   3.325554   0.000000   0.000010   0.019865   2.731908      6
 6.3407E+04   8.228208   4.758432  11.749309  17.778100  -8.668121   1.938093   0.000000   0.000000   0.583866  0.189E-03      lgT_cntr
                                 Fe_core        0.0000000000000000
                          fe_core_infall        0.0000000000000000
                               rel_E_err    1.6619538194637069D-10
                       log_rel_run_E_err       -5.9765340617087022
                     dt_div_max_tau_conv        0.0346354744084729

       2744   9.575113   3121.198  -6.079917  11.742047  12.000000   8.674446   0.000000   0.000000   0.280038   5.092951   1821    237
 9.2838E+01   9.575113   2.913227   7.931987  10.608884 -99.000000   3.325554   0.000000   0.000010   0.019865   2.753005      7
 6.3500E+04   8.228155   4.758432  11.741980  17.764824  -8.668121   1.938093   0.000000   0.000000   0.583410  0.189E-03      lgT_cntr
                                 Fe_core        0.0000000000000000
                          fe_core_infall        0.0000000000000000
                               rel_E_err    3.9780075371441945D-12
                       log_rel_run_E_err       -5.9764126638280715
                     dt_div_max_tau_conv        0.0406953229383509

       2745   9.596451   3121.198  -6.448802  11.746953  12.000000   8.674446   0.000000   0.000000   0.280038   4.768625   1823    237
 8.4008E+01   9.596451   2.913227   7.931196  10.608569 -99.000000   3.325554   0.000000   0.000010   0.019865   2.599970      7
 6.3584E+04   8.221780   4.758432  11.746887  17.754377  -8.668121   1.938093   0.000000   0.000000   0.587301  0.188E-03      lgT_cntr
                                 Fe_core        0.0000000000000000
                          fe_core_infall        0.0000000000000000
                               rel_E_err    5.7116415346976825D-13
                       log_rel_run_E_err       -5.9763016541989185
                     dt_div_max_tau_conv        0.0279300172978451

 retry: delta_lgT_cntr hard limit    1823    2745
       2745   9.571867   3121.198  -6.321800  11.759400  12.000000   8.674446   0.000000   0.000000   0.280038   5.139655   1823    238
 6.3006E+01   9.571867   2.913227   7.931402  10.607934 -99.000000   3.325554   0.000000   0.000012   0.019865   2.766009      6
 6.3563E+04   8.228352   4.758432  11.759336  17.756731  -8.668121   1.938093   0.000000   0.000000   0.587299  0.188E-03         retry
                                 Fe_core        0.0000000000000000
                          fe_core_infall        0.0000000000000000
                               rel_E_err    8.2705770262738297D-09
                       log_rel_run_E_err       -5.9729377939966763
                     dt_div_max_tau_conv        0.0185648008972751

       2746   9.559986   3121.198  -6.204312  11.718049  12.000000   8.674446   0.000000   0.000000   0.280038   5.321120   1841    238
 6.3006E+01   9.559986   2.913227   7.930862  10.607446 -99.000000   3.325554   0.000000   0.000012   0.019865   2.843787      7
 6.3626E+04   8.230638   4.758432  11.717978  17.740033  -8.668121   1.938093   0.000000   0.000000   0.588596  0.188E-03          hold
                                 Fe_core        0.0000000000000000
                          fe_core_infall        0.0000000000000000
                               rel_E_err    1.1605699228021003D-12
                       log_rel_run_E_err       -5.9728743685147503
                     dt_div_max_tau_conv        0.0144992566531454

 retry: delta_lgT_cntr hard limit    1841    2746
       2746   9.578221   3121.198  -6.113616  11.727059  12.000000   8.674446   0.000000   0.000000   0.280038   5.040251   1841    239
 4.7254E+01   9.578221   2.913227   7.931005  10.608075 -99.000000   3.325554   0.000000   0.000013   0.019865   2.718730      7
 6.3610E+04   8.226332   4.758432  11.726990  17.741768  -8.668121   1.938093   0.000000   0.000000   0.588815  0.188E-03         retry
                                 Fe_core        0.0000000000000000
                          fe_core_infall        0.0000000000000000
                               rel_E_err    1.0182203546459458D-12
                       log_rel_run_E_err       -5.9728916627428790
                     dt_div_max_tau_conv        0.0007903040152960

       2747   9.558849   3121.198  -6.390823  11.730069  12.000000   8.674446   0.000000   0.000000   0.280038   5.337689   1831    239
 4.1282E+01   9.558849   2.913227   7.930593  10.607891 -99.000000   3.325554   0.000000   0.000014   0.019865   2.857434      5
 6.3652E+04   8.230684   4.758432  11.730000  17.741437  -8.668121   1.938093   0.000000   0.000000   0.587525  0.188E-03      lgT_cntr
                                 Fe_core        0.0000000000000000
                          fe_core_infall        0.0000000000000000
                               rel_E_err    2.7909316004782299D-08
                       log_rel_run_E_err       -5.9615944986007374
                     dt_div_max_tau_conv        0.0106792435058055

 retry: delta_lgT_cntr hard limit    1831    2747
       2747   9.555884   3121.198  -6.531701  11.740553  12.000000   8.674446   0.000000   0.000000   0.280038   5.384714   1831    240
 3.0961E+01   9.555884   2.913227   7.930703  10.608342 -99.000000   3.325554   0.000000   0.000015   0.019865   2.879200      7
 6.3641E+04   8.231407   4.758432  11.740486  17.742798  -8.668121   1.938093   0.000000   0.000000   0.587253  0.188E-03         retry
                                 Fe_core        0.0000000000000000
                          fe_core_infall        0.0000000000000000
                               rel_E_err    6.4699724534475005D-12
                       log_rel_run_E_err       -5.9728429224137498
                     dt_div_max_tau_conv        0.0072838391623541
```
This just keeps going with ` retry: delta_lgT_cntr hard limit` until the time-step reaches 1d-4 seconds, at which point it spends the next thousand+ steps recovering. Note that this failure is with the default inlist copied from `main`! No TDC at all!

I've tried a number of things to resolve this, including:
1. Removing `delta_lgT_cntr hard limit`.
2. Changing `thermohaline_coeff` to `100` away from the default of `0`.
3. Disabling MLT++.
4. Tightening `delta_lgRho_limit`.
6. Tightening `dX_nuc_drop_limit`.
7. Tightening `varcontrol_target`.
8. Fiddling with the `_tol_max_residual` controls.
9. Disabling `convergence_ignore_equL_residuals`.
10. Disabling `make_gradr_sticky_in_solver_iters`.

In most cases I can avoid the specific issue that crops up above, but at the cost of later similar issues, or in some cases issues where the evolution becomes substantially different (e.g. enabling thermohaline mixing).


adamjermyn:
I should also mention: this PR resolves issue #270 (making `make_co_wd` work and re-enabling it).

rjfarmer:
I think the problem in dev_to_cc_12 is much earlier at model around 1490 when we switch from inlist_to_lgTmax_header to inlist_to_cc_header. At this point we have center convective silicon burning region but it does not grow in size. Maybe there is an issue with the model loading?

rjfarmer:
It takes alot of retires at 1498

adamjermyn:
Hmmm that is suspicious. Thanks! I'll go dig into that.

adamjermyn:
Commit `9ffa01f` resolves issue #197.

adamjermyn:
There was a bug in photo/model loading, but it only affects some flags that this test case doesn't care about. Still looking.

adamjermyn:
Some flags were being set incorrectly. I've fixed that now, but none of that appears to have mattered.

It looks like the convection speeds might not be saving/loading correctly. They are being saved and loaded, but sum(mlt_vc) changes by about 10% from the end of lgTmax to the beginning of cc.

adamjermyn:
Ok model *saving* is working correctly. Checking loading...

adamjermyn:
Nevermind loading is working just fine. False alarm (was loading the wrong model).

adamjermyn:
My last commit seems to have broken things, looks like an issue with initial model building. Will investigate.

adamjermyn:
I've gone through the flags and loading/saving logic in detail and it all looks fine. It also does the right things and interacts with the `relax` routines correctly. So now I'm looking at the test cases in more detail to see what might be happening...

adamjermyn:
`5M_cepheid` was hitting trouble where the post-relax state was still evolving rapidly relative to the turnover time. This means the convection speeds change a lot. Because it's convective near the surface, Teff and logL are sensitive to convection. There's a timestep limit on motion in the HR diagram, so what happened is that a cell would turn radiative, suddenly changing the thermal balance near the surface, pushing it too far in the HR diagram, and driving down the timestep. The cell would turn back to being convective, rinse and repeat. Different cells change each time (even with remeshing turned off). The process never settles down, presumably because the star never evolves far enough to stop evolving rapidly.

I think the higher-level issue is that small errors returned by the `star` solver are enough to change the surface convection properties substantially. And because TDC has time dependence there's hysteresis in those properties on short time-scales, so you can get a 'pulsation-like' process driven by solver noise.

There are two ways around this. The first is to take longer time steps. Just uncommenting the delta_HR control in the starting inlist fixes this by letting star take long enough time-steps that TDC reduces to MLT, the hysteresis goes away, and the star proceeds onto the MS. The other option is to just disable TDC in the starting inlist. I've done the latter because I didn't feel comfortable loosening tolerances on this case without knowing more about it.

On to the next test case.

adamjermyn:
`dev_to_cc_20` and `dev_to_cc_80` and `ccsn_IIp` were all failing due to a bug swapping gradL and gradA in TDC. That's fixed now. They might still fail, but it'll have to be for other reasons :-)

adamjermyn:
Based on local testing I'm expecting everything to pass on the optional run in `5f89624` ([Testhub](https://testhub.mesastar.org/TDC_on_default/commits/5f89624)). This follows a recent (this morning) merge *from* `main`, so if the test suite passes I'll plan to merge unless I hear objections.

PR #301, opened 2021-08-03
Fixed name in loading script for running.rst

alexji:
I am running with mesa-r15140 and encountered an error while following the tutorial on this page (copy-pasting the suggested `inlist_load` script).
```
At line 609 of file ../private/star_job_ctrls_io.f90
Fortran runtime error: Cannot match namelist object name load_model_filename
```

This PR fixes the name.

### DIFF
--- a/docs/source/using_mesa/running.rst
+++ b/docs/source/using_mesa/running.rst
@@ -249,7 +249,7 @@ do this your inlist might look like:
 
      ! start a run from a saved model
      load_saved_model = .true.
-     load_model_filename = '15M_at_TAMS.mod'
+     save_model_filename = '15M_at_TAMS.mod'
    
      ! display on-screen plots
      pgstar_flag = .true.


jschwab:
Alex, you win the prize for being the first external MESA pull request!

I think there is a version confusion issue here though.  On `main`, the star_job option `saved_model_name` has been replaced with `load_model_filename` everywhere.

On ReadTheDocs, the docs are versioned and it looks to me like we are displaying the correct instructions for each versions.

https://docs.mesastar.org/en/r15140/using_mesa/running.html#loading-a-model
https://docs.mesastar.org/en/latest/using_mesa/running.html#loading-a-model

alexji:
I see! I misunderstood then thinking that "latest" should correspond to r15140; what should it correspond to, since I don't see a newer version (and r15140 is what the installation links point to)?

jschwab:
The `latest` version of the docs corresponds to the `main` branch on GitHub, which is the MESA development version.  (The docs are dynamically generated, so correpond to whatever the HEAD commit is on `main`).

The ReadTheDocs default version (i.e., what you get if you go to docs.mesastar.org) should be r15140, which is the most recent release.

So r15140 is the right version of the code to be using and we just have to make sure folks are getting the matching docs.  If there's a spot to note some of this that would have eliminated this confusion, let me know and we'll try to make it clearer.

alexji:
I see! If it helps, I found the "latest" page through two different ways: through googling "mesa stellar evolution docs" and from the old MESA sourceforge documentation (e.g. http://mesa.sourceforge.net/star_job_defaults.html).

Since there is a large set of links pointing to "latest", I think a good way to avoid confusion would be adding some text on the homepage (https://docs.mesastar.org/en/latest/) that says "this is the documentation for the latest development version of MESA. For for the latest release documentation (version rXXXXX) see (link)".
It will be a bit annoying to update this each time you release a version, but that seems easier than trying to fix all the other links.

Closing the PR!

jschwab:
Aha!  Thanks for the pointer.  I fixed the link on http://mesa.sourceforge.net/star_job_defaults.html to point to the r15140 version.

PR #311, opened 2021-08-24
Update code and documentation to consistently present eps_grav

jschwab:
The implementation of eps_grav underwent many cleanups and
improvements over the past few years.  This updates the code and docs
so that they agree about the meaning of the available options and the
equations being solved.

### DIFF
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -7527,8 +7527,112 @@
 
     min_magnitude_brunt_B = -1d99
 
+
+
+
+
 ! structure equations
 ! ===================
+
+      ! use_dedt_form_of_energy_eqn
+      ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~
+      ! always_use_dedt_form_of_energy_eqn
+      ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+      ! ::
+
+    use_dedt_form_of_energy_eqn = .false.
+    always_use_dedt_form_of_energy_eqn = .true.
+
+
+      ! no_dedt_form_during_relax
+      ! ~~~~~~~~~~~~~~~~~~~~~~~~~
+      ! dedt_eqn_r_scale
+      ! ~~~~~~~~~~~~~~~~
+      ! ::
+
+    no_dedt_form_during_relax = .true.
+    dedt_eqn_r_scale = 1d8
+
+
+      ! always_use_eps_grav_form_of_energy_eqn
+      ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+      ! If true, use the ``eps_grav`` form of the energy equation.
+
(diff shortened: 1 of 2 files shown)


warrickball:
A few thoughts because I've long found these options confusing myself. It looks like `use_eps_grav_form_of_energy_eqn` corresponds to eq. (65) in MESA IV, which is the "MESA default" according to Table 4. If that's right, shall we document that explicitly?

Also, what happens if both options are set to `.false.`? e.g. if I set
```f90
    use_dedt_form_of_energy_eqn = .false.
    use_eps_grav_form_of_energy_eqn = .false.
```
If the option is a choice (i.e. one and only one must be true), shouldn't we use a string? e.g.
```F90
    energy_eqn_form = 'dedt' ! or 'eps_grav'
```
Finally, isn't the name `'eps_grav'` fundamentally confusing because the `'dedt'` form of the equation has an ε_grav term too? That is, `'eps_grav'` doesn't really distinguish any form of the equation whereas `'dedt'` does because that's only in eq. (63).

jschwab:
This whole situation has definitely been confusing on both the physics and the software side.

> It looks like use_eps_grav_form_of_energy_eqn corresponds to eq. (65) in MESA IV, which is the "MESA default" according to Table 4. If that's right, shall we document that explicitly?

Yes, the form we use is MESA IV, equation (65) + the composition term (which was dropped there).  Most of the other options in Table 4 have now gone away.

> If the option is a choice (i.e. one and only one must be true), shouldn't we use a string?

Yeah, you're right, it is probably better to either reduce to either a single boolean or a string option (the latter more easily allowing for other choices in the future).

> Finally, isn't the name 'eps_grav' fundamentally confusing because the 'dedt' form of the equation has an ε_grav term too? That is, 'eps_grav' doesn't really distinguish any form of the equation whereas 'dedt' does because that's only in eq. (63).

I think 'eps_grav' is OK, though I welcome alternative naming suggestions.  Consistent with the name, the MESA variable `eps_grav` only gets defined in the 'eps_grav' version.

Your confusion may be because the 'dedt' form mentioned in MESA IV (i.e. Table 4, eq. 63) is not currently what is meant by dedt form.  That gets superseded by MESA V, Section 3 and is also MESA VI eq 76 (aka MESA V eq. 9).

In MESA V, we called them dedt and dLdm forms, which was terribly confusing because both versions have a dL/dm term.

jschwab:
I just pushed a commit with changes in this direction.  Now you set `energy_eqn_option = 'dedt'` or `energy_eqn_option = 'eps_grav'`.

I'll write a changelog entry and some more docs once we converge on an approach.

jschwab:
I am 100% against using 'dLdm'.  We made a mistake, but that doesn't mean we have to keep making it :)

Yeah, I think 'dedt' and 'eps_grav' is probably good enough, if accompanied by accurate docs.  I'll flesh the docs a bit more and write a changelog entry and we can see what folks think.

PR #322, opened 2021-09-20
Moved local turbulent theories into a new `turb` module.

adamjermyn:
MLT/TDC/thermohaline/semiconvection are all local turbulent theories that only need data from a single cell or face (typically variables defined on that element + gradients there). This PR implements splits them off into a separate module and calls them via a `turb_support` module in `star` (formerly `mlt_get_results`).

My hope is that this will keep the internals of these routines separate from `star` and reduce the (apparent) complexity of `star` by reducing the number of files one needs to grok to follow what's going on there.

Renaming to `turb` from `mlt` reflects that we consider more than just MLT in evaluating turbulent mixing.

In principle `conv_premix` and `predictive_mixing` could be pulled in too, but thematically those feel like they belong in `star` (at least to me), because they require digging around in the internals of the star across many cells.

Some of the names of functions across the interface can probably be improved (feel free to make suggestions!).


rjfarmer:
Few things:

- Missing entry in CODEOWNERS
- Copyright statements are 2 years out of date (I know we do't update the old files but when making new files we should make them properly)
- How does this interact with the other_mlt_results hook?
- It does need some (working) tests before merging

warrickball:
I'm testing this branch with `ifort` in case there are any special compilation issues for crazies who don't use the SDK.

As an aside, I like this modularisation. AFAIK there was a time (10398 and before?) when `mlt` was a separate module. You might be able to look there for inspiration of what tests were run before. But I imagine you might also cook up a better suite of unit tests on your own anyway.

adamjermyn:
Looks like `ifort` fails to compile. Not sure why it goes on to report results for the test cases if it didn't manage to compile though...

I'm not quite sure how to interpret the compilation error:
```
LIB_TOOL libstar.a
MAKEDEPF90
TEST_COMPILE ../src/run_star_extras.f90
TEST_COMPILE ../../../star/job/run_star.f90
TEST_COMPILE ../src/run.f90
LOADER ../star
ld: ../../make/libstar.a(turb_support.o): in function `turb_support_mp_get_results_':
turb_support.f90:(.text+0x960): undefined reference to `tdc_mp_set_tdc_'
ld: turb_support.f90:(.text+0xa7b): undefined reference to `mlt_mp_set_mlt_'
ld: turb_support.f90:(.text+0xb50): undefined reference to `thermohaline_mp_set_thermohaline_'
ld: turb_support.f90:(.text+0xc58): undefined reference to `semiconvection_mp_set_semiconvection_'
make: *** [makefile_base:38: ../star] Error 1
```

Is there a `.mod` file missing in `ifort`?  I mostly copy-pasted the `auto_diff` `makefile_base`. The install line reads:
```
install:
        @$(CP_IF_NEWER) turb.mod $(MESA_DIR)/include
        @$(CP_IF_NEWER) $(LIB) $(MESA_DIR)/lib
ifneq (,$(findstring ifort,$(FC)))
        @$(CP_IF_NEWER) mlt.mod $(MESA_DIR)/include
        @$(CP_IF_NEWER) tdc.mod $(MESA_DIR)/include
        @$(CP_IF_NEWER) semiconvection.mod $(MESA_DIR)/include
        @$(CP_IF_NEWER) thermohaline.mod $(MESA_DIR)/include
        @$(CP_IF_NEWER) turb.mod $(MESA_DIR)/include
endif
```

warrickball:
Sorry, I've been working on this. It was because the makefile that got copied didn't include the new `turb` module. I killed those runs once started and have resubmitted, and it compiled OK. The tests are still running (I've "only" been allocated 72 cores) and the results will be noisy but there isn't any reason to worry yet.

adamjermyn:
I've added a unit test to check the scaling of vc in MLT in the efficient limit.

I wasn't sure what other unit tests would be useful here. Ideally we want to check things that we know analytically, but that's a pretty limited set (especially for e.g. semiconvection and MLT).

One other I suppose is to check that TDC actually solves the TDC equations, but there's a check inside TDC itself, and it issues an error if it fails.

I looked at the unit tests in the old (10398) mlt module. They're ... not super helpful. They're making sure that it reproduces exact numerical results from specific examples where it was evaluated. A test like that failing doesn't really tell you much, but I'm happy to throw more like that in...

None of that seems to be merge-blocking though (we're in a strictly better position now re: testing than we are on `main`).

So @warrickball let me know what happens with `ifort`. If we can get that working I'll merge and we can take it from there.

warrickball:
> I looked at the unit tests in the old (10398) mlt module. They're ... not super helpful. They're making sure that it reproduces exact numerical results from specific examples where it was evaluated.

That doesn't surprise me much. I think a lot of the module unit tests are like that (which is probably something we should improve...). I agree that reproducing some analytic results is better but there probably should be one or two evaluations of known results. I know in e.g. `kap` this encumbers us with updating the test output for minor precision changes but it at least flags when the output of those routines is changing unexpectedly, if only slightly.

> So @warrickball let me know what happens with ifort. If we can get that working I'll merge and we can take it from there.

Everything except `c13_pocket` passed in [9351f9c](https://testhub.mesastar.org/turb_mod/commits/9351f9c). `c13_pocket` will hopefully finish soon, narrowly inside the 8-hour max runtime in my cluster scripts. That said, I just tried to build 098e723 and have a failure to match output because `ifort` and `gcc`'s unformatted floats are slightly different. e.g. at lines 59, 69 and 71 in `turb/test/src/test_turb.f90`, the `write(*,*)` statements give different output under the different compilers:
```
6,8c6,8
<  vc at 1d5 Lsun   3965420.37858666     
<  vc at 1d8 Lsun   39765811.7067806     
<  Ratio:   10.0281452936281     
---
>  vc at 1d5 Lsun   3965420.3785866620     
>  vc at 1d8 Lsun   39765811.706780650     
>  Ratio:   10.028145293628064     
```
I'll commit some explicit formatting in the next ~30 minutes if you don't object.

I'll draw your attention to the [appropriate part of the coding style guide](https://docs.mesastar.org/en/latest/developing/code_style.html#formatting). ;-)

warrickball:
`c13_pocket` finished successfully and I've committed the explicit formats to the `turb`'s tester. If this passes then I'm happy to merge.

To reply to one of your earlier comments,
> Not sure why it goes on to report results for the test cases if it didn't manage to compile though...

Presumably a scripting error somewhere on my part. In order to separate the compilation report from the individual tests, they are run by different scripts. Failure to compile *should*, as you say, stop the test cases from running but I've clearly made a mistake somewhere! That only a few tests ran in this failed state is because I manually cancelled everything that hadn't run.

adamjermyn:
Looks good, and passes! Merging.

rjfarmer:
Well i guess github never showed the other suggestions I had:

I don't think we need the make/mkb file?

Why does turb not follow every other module by having a lib and def file?

Also every other module puts a wrapper code in lib to call into private code rather than just importing everything

adamjermyn:
Indeed it didn't!

I'll look into the first two.

On the third: The private files each export just the routine that needs to be used in the interface. I could write a wrapper on the public side but I think that would just complicate things... is there an advantage to adding these wrappers?

rjfarmer:
The point is people should be able to look at the public interface (and potentially hook into) without a huge fear that its going to change or have to understand the internal details. Finding things in private/ means you have to dig much deeper into the internals to understand what can and can't be called (sure you made them public but now i have to go to every file to work out what I can/can't call). 

Consistency with the rest of mesa also has some value.

adamjermyn:
That makes perfect sense, thanks. I'll move the appropriate routines directly into the interface.

FWIW `auto_diff` also uses this import model, but there I think it's a little more excusable (the `auto_diff` code isn't meant to be human-edited, and the number of things that would need to be in the public routines is rather large).

adamjermyn:
Regarding `lib` and `def`, that's just because there's (currently) nothing to `def`. Happy to make a `def` file if/when there is.

PR #330, opened 2021-09-30
Restore double_bh and wind_fed_bhhmxb

orlox:
This is my long overdue first pull request, so please let me know if I'm doing anything bad in here.

This is just a small amount of changes meant to bring double_bh and wind_fed_bhhmxb back to life. Changes have been also made so that dev_ppisn actually does something, so it is still now in working order. Ran the test suite on this a couple of days ago and everything is green.

https://testhub.mesastar.org/test_case_fixes/commits/5ea03a8


orlox:
ppisn is not passing yet, changes are only so that it actually runs a bit rather than crash from the get-go

evbauer:
Looks good, so I think this is ready for a "squash and merge". Thanks for working on these. Should we continue to leave this branch open after merging for work on other test cases?

orlox:
Plan to keep working on the branch to restore the remaining ones on my Todo list. Git-ignorant question, if this is merged into main will it appear as a single commit? Mostly don't want the crazy rebasing mess I made to end up there.

evbauer:
Yes, that's what the "squash" part of "squash and merge" is for. I just went ahead and did that.

Issue #331, opened 2021-10-05
`num` test failures beyond bit-for-bit differences with intrinsic math


warrickball:
If you try to build MESA with `USE_CRMATH = NO` in `utils/makefile_header`, some of the differences in `num`'s test results are surprisingly large. You'll need to skip `auto_diff`'s tests with `touch auto_diff/skip_test`. 

e.g. for the Cash–Karp method, the expected output is
```
                                             calculated           1    1.7633919298826672D+00
                                             calculated           2   -8.3553976467637270D-01
```
whereas we get
```
                                             calculated           1    1.7634625993645798D+00
                                             calculated           2   -8.3545009921766566D-01
```
which differs already at the ~0.001 level. There are also various similarly large differences when calling the implicit solvers with numerical Jacobians (though differences seem to remain small with analytic Jacobians).

I'm personally not worried by differences within a few orders of magnitude of machine precision (say, <1e-10) but I don't understand why these are so different. There are very few calls to functions that are replaced by CRMATH and they seem to be confined to adjusting stepsizes anyway.

rjfarmer:
I guess the first question is usually do we even use cash-karp? There's a lot in num and mtx we don't use anymore (or maybe ever)

warrickball:
Cash–Karp isn't the only one but FWIW
```
$ grep cash_karp star/p*/*
star/private/hydro_rotation.f90:         call cash_karp_work_sizes(1,liwork,lwork)
star/private/hydro_rotation.f90:            call cash_karp( &
```
That call is in `eval_fp_ft`.

adamjermyn:
Numerical jacobians will be computed with finite differences, which loses
~most precision right off the bat. Small differences then get magnified
quickly...

-Adam

fxt44:
the jacobian "only" guides the solution (e.g., how big a step can be taken for some accuracy criteria) and should not change the solution obtained. yes?

warrickball:
> Numerical jacobians will be computed with finite differences, which loses ~most precision right off the bat. Small differences then get magnified quickly...

On one hand, I totally agree. That's is also where I expect the problem is. On the other hand, even the finite difference Jacobians appear to only use things like +, -, *, / etc, rather than more troublesome functions like log, exp, pow, sin, cos, etc. I don't think CRMATH replaces those basic operations, so I expect the Jacobians to only differ to within a few orders of machine precision. Will that really be amplified?

[...]

PR #332, opened 2021-10-06
Removes old prescription for calculating moment of inertia (equivalent to setting `fitted_fp_ft_i_rot = .true.`).

adamjermyn:
Suggested by @orlox in #331, as we were only keeping the old prescription around in case the new one ran into trouble. Given that it's been two years since the change (MESA 11701) it may be time to remove it.

@orlox Could you take a look and make sure I didn't mess anything up in the process? I'll wait for you to sign off and for testhub to clear before merging.

### DIFF
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -3911,12 +3911,10 @@
       ! If false, use slightly more complex expression
       ! that takes into account finite shell thickness.
       ! In practice, there doesn't seem to be a significant difference.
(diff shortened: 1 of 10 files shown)


adamjermyn:
Oh and we need to decide what to do with `simple_i_rot_flag`, which no longer does anything...

adamjermyn:
Ok, restored. I've updated the comments for `simple_i_rot_flag` to remove reference to the fits. Just waiting on the tests now...

PR #333, opened 2021-10-07
Ideal eos

adamjermyn:
Introduces a new EOS made of ideal ion + radiation for use as a backplane.

@evbauer and @orlox maybe take this for a spin and let me know how it works? I'll also be looking at testhub before merging of course...

### DIFF
--- a/star/test_suite/12M_pre_ms_to_core_collapse/standard_after_core_c_burn.mod
+++ b/star/test_suite/12M_pre_ms_to_core_collapse/standard_after_core_c_burn.mod
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:9f0f5305940f763e49a725456bca10f5a78acafe34c5a5710e24797f7e8afba9
+oid sha256:8618ec8f68fd3dbbeb5d97e1fef6c618419590248c2fb7272000f5f86c03c186
 size 952091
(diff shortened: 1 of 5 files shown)


orlox:
Also for this composition there is a funny logP coming from HELM in the dragons region, but I guess that's unrelated to the implementation of the ideal EOS. Here near the edge of HELM lnP drops to very negative values.

eos_plotter
eos_regions

fxt44:
the funny P is coming from helm's coulomb corrections. it would be better, i think,  if skye covered the dragons region.

adamjermyn:
Yeah 🙁 . The only reason we use HELM there in fact is that Skye goes bad...

fxt44:
didn't the latest cms eos cover this dragon range? i recall aaron looked into it.

evbauer:
I was just in the process of writing a message about this. I think the failure in `make_o_ne` is fairly benign...

evbauer:
Oh, yes, those certainly should be zero for an ideal EOS. Was it just giving uninitialized garbage for those?

adamjermyn:
I would guess that you're hitting a kap boundary that low in temperature...

-Adam

adamjermyn:
I'll get the extra fraction info in `star` on Monday if no one beats me to it. Then I guess it's just down to what testhub says and what we need to clean up...

Issue #169, opened 2021-01-27
Reduce IO operations by using flush instead of repeated open/close.


adamjermyn:
This was suggested by Nick Carriero of the Flatiron Scientific Computing Core. Nick wrote to me with:

> One quick thing to do would be to use flush rather than close, if the goal is to ensure data is written to disk. The metadata ops involved in closing and then reopening a file are much "heavier" than just reading from or writing to a file.

This came up because I was running many jobs on a cluster and the number of ops became an issue. Nick looked into it, noticed that 3/7 of those ops were involved in opening and re-closing `history.data`, `profiles.index`, and `pgstar.dat`, each of which gets opened then closed once per step.

His suggestion is to open these files once at the start of a run, then use the Fortran `FLUSH` intrinsic after our write statements are all done to ensure that all data is written to disk. This can happen once per step, and we just close the file handle at the very end.

I don't think this is urgent, and I have a workaround for my use case (just locating all of the files on a local scratch system then copying them back when the run finishes), but I didn't want to forget it because it seems like a good idea for a rainy day.

PR #339, opened 2021-11-02
Flush files rather than close() them

rjfarmer:
This works on #169

@adamjermyn  can you have you cluster people see if this helps?

### DIFF
--- a/star/private/history.f90
+++ b/star/private/history.f90
@@ -39,10 +39,6 @@
          do_get_data_for_history_columns, write_history_info
 
 
-      logical, parameter :: open_close_log = .true.
-
-
-
       contains
 
 
@@ -251,7 +247,7 @@
          end if
 
          i0 = 1
-         if (write_flag .and. (open_close_log .or. s% model_number == -100)) then
+         if (write_flag) then
             if(.not. folder_exists(trim(s% log_directory))) call mkdir(trim(s% log_directory))
 
             fname = trim(s% log_directory) // '/' // trim(s% star_history_name)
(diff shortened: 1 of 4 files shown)


rjfarmer:
Better hold off as I seem to have broken most test cases https://testhub.mesastar.org/rf%2Fflush-io/commits/68a1d0f

warrickball:
I figured some of the failures might be `cannon`'s filesystem playing up so I also submitted it to `bluebear_ifort`. Most of the failures seem to be consistent across both systems. Many seem to happen because some output (looks like `history` data?) ends up being written to `stderr` and I think TestHub regards *anything* on `stderr` as a failure.

rjfarmer:
I was actually worried this casued cannons failures...

There are at least 2 issues one is some history data gets outptted to terminal and two there is an issue with the pgstar.dat file

rjfarmer:
Let see whats broken now https://testhub.mesastar.org/rf%2Fflush-io/commits/171cd41

rjfarmer:
I'm closing this for now as it needs a bigger restructure than i was expecting

matthiasfabry:
Do we have any idea why history data ended up on stderr with these changes? They seem sensible, or are there Fortran subtleties we'd be running into here?

fxt44:
maybe one thread is still writing/flushing while another thread is closing?

matthiasfabry:
That seems weird. Do multiple threads need to write to the same file? Also Rob's changes do not mention `close` anymore. In any case, could this be tested with `!$OMP CRITICAL`'s?

PR #342, opened 2021-11-12
Purge conv_vel equation

orlox:
This removes the remains of the old convective velocity equation, leaving MESA with a single option to do time dependent convection, TDC. It also moves the code that restores the *_old values to a single point in the code (set_current_to_old) to fix an issue on restore_mesh_on_retry.


fxt44:
woohoo! i take it you got tdc working with ppisn ...

orlox:
Not really. TDC is showing some strange behavior that I shared with Adam (final convective velocities changing very significantly under machine precision variation of inputs, though this is with the peculiar conditions of ppisn). ppisn can be made to work by shutting of convection during the pulse, just need to tune things so the test is not fragile. Without convection the behavior of the pulse changes, but I think that will have to do for the moment.

rjfarmer:
Could we get a changelog entry for this? Also with what ever our recommendation going forward is, for short timescale convection.

What happens with old mod files that have conv_vel as a column?

fxt44:
thanks pablo.

orlox:
Added to the release notes to say that TDC is the recommended method now. Regarging saved models with conv vel information, that is still read but would be rewritten by MLT, don't know if TDC takes that information from saved models (@adamjermyn ? Maybe you also want to add more detail about what TDC is exactly to the release notes?). Other than that the changes are coming all green on the test suite.

rjfarmer (approved):

PR #356, opened 2021-12-13
Fixes for rotation (Radec's and Mathieu's issues)

orlox:
These changes do two things:

- Include the effect of the Brunt on the calculation of rotational diffusion coefficients when using the Schwarzschild criterion. This is the problem Radec reported with large differences in rotational mixing between Ledoux and Schwarzschild.
- Restore time smoothing of the ST diffusion coefficients. This solves Mathieu's issue.

Attached is a comparison of a model of a 30 Msun star at 0.1 critical (adapted from Mathieu's example) ran with 15140 and HEAD of the pm/rotation_fixes branch. I also run the simulations with and without time smoothing, and it gives me consistent results. To do the time smoothing one needs to interpolate the diffusion coefficient from the old mesh in the previous step, I wrote down a simple linear interpolant in rotation_mix_info for that purpose. If anyone wants to switch that to use the built in interpolants, please go ahead (but be sure that the attached results are still reproducible).

comparison
[final_tests.tar.gz](https://github.com/MESAHub/mesa/files/7705015/final_tests.tar.gz)

### DIFF
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -3918,10 +3918,10 @@
     smooth_D_SSI = 0
     smooth_D_ES = 0
     smooth_D_GSF = 0
-    smooth_D_ST = 3
-    smooth_nu_ST = 3
-    smooth_D_omega = 3
-    smooth_am_nu_rot = 3
+    smooth_D_ST = 0
+    smooth_nu_ST = 0
+    smooth_D_omega = 0
+    smooth_am_nu_rot = 0
 
 
       ! ST_time_smooth_frac
@@ -3932,7 +3932,7 @@
 
       ! ::
 
-    ST_time_smooth_frac = 0.2d0
+    ST_time_smooth_frac = 0.5d0
 
 
       ! simple_i_rot_flag
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -3932,7 +3932,7 @@
 
       ! ::
 
-    ST_time_smooth_frac = 0.8d0
+    ST_time_smooth_frac = 0.2d0
 
 
       ! simple_i_rot_flag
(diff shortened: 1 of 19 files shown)


fxt44:
nice work pablo! would a test suite case be useful to help avoid this in the future?

orlox:
I think we already have one, 15M_dynamo. Perhaps taking that breakage more seriously as a sign of an issue could have sorted this out. But that's a bit hard to do when target values in test cases are missing jusitification, so I think that's the way to improve on that situation.

fxt44:
ok. as i tweaked 15m_dynamo to be more robust (basically it was grossly under-resolved), i'm game for tweaking it again to have it more adequately cover this case.

orlox:
Perhaps would be a good idea to ground the model on something published showing the impact of the dynamo. That way when referring to why we think the target values are good we can just say they're taken from X paper +- some tolerance.

fxt44:
ok. any suggestion for a published article?

adamjermyn:
Looks good to me.

orlox:
@fxt44 , Heger, Woosley & Spruit 2005 looks like a good one (https://ui.adsabs.harvard.edu/abs/2005ApJ...626..350H/abstract). In particular their tables 1 and 2 have numbers we can compare directly to. Would need some tuning to match their model assumptions though, at a first glance I do not see, for example what mass loss rate they use, which are pretty relevant to the final answer.

There's also Sujis et al. (2008) (https://ui.adsabs.harvard.edu/abs/2008A%26A...481L..87S/abstract). One can try to use their table 1 and see if we can reproduce those values of angular momentum. 

Jim also did calculations with vanilla TS dynamo on top of his own. One could try to reproduce the only TS values he has in Fig 3 of his 2019 paper (https://ui.adsabs.harvard.edu/abs/2019MNRAS.485.3661F/abstract).

fxt44:
thanks!

PR #357, opened 2021-12-16
TDC solver improvements

adamjermyn:
The motivation for this PR is the observation that when your timesteps are too long and you're transitioning from convective to radiative there are multiple solutions to the equations that TDC solves.

The equation TDC solves is given below, and is `Q-\Gamma=0`:
933374CA-3ED9-4AD5-A782-AB40AFFBB6E9 (1)
Plotting this gives:
BD7B675F-26CF-4846-B7FF-7C4DCFC3C800 (1)
Basically drawing a line of fixed `\Gamma` through the `Q` curve shows that depending on `\Gamma` you can have 1, 2, or 3 solutions. For `\Gamma > 0` there is always only one solution (above, panel on the right). For negative `\Gamma` there can be up to three solutions. When there are three, one of them always has convective velocity `w=0`. Here `y` is the superadiabaticity. 

There isn't a super principled way around this, other than using smaller time-steps. Rough guidance is that you want the time-step to be less than a dynamical time divided by the radiative sub-adiabaticity, but that's a very rough guide.

The existence of multiple solutions is not just a physical problem, it's also a numerical problem, because there are no formal guarantees about which solution the TDC solver will pick (when there are multiple). Thanks to @orlox for an awesome bug report of a minimal and reproducible example where the solver flip-flops based on parameters being nudged around at machine precision.

This PR is aimed to address that problem. It does this by forcing the solver to identify all three branches (when they exist) and then picking the one in which the convective velocity decays most slowly.

This doesn't eliminate discontinuities in the solution (think of finding the smallest real solution to a cubic, that's not far off from what's happening conceptually) but it does mean that when there's a jump it's usually a smaller jump in the structure of the star.

Currently there are three failing test cases
```
1M_pre_ms_to_wd
ccsn_IIp
hb_2M
```
and one not-yet-finished blocking merge (`c13_pocket`).

I'm starting on `ccsn_IIp` now, and if anyone wants to look at the others that'd be much appreciated.


orlox:
I won't have much time to help sort out those test cases, but one workaround would be to revert 1M_pre_ms_to_wd and hb_2M to Cox if we want a full-green test-suite.

adamjermyn:
1M_pre_ms_to_WD isn't even using TDC now...

adamjermyn:
I've reverted hb_2M to Cox. If we care later we can put it back on TDC.

orlox:
don't know if you reproduced locally the error in 1M_pre_ms_to_WD, but cannon segfaulted on that one

https://logs.mesastar.org/dfd4f4f20f49af0ff347c738b1cae9b2e057216e/cannon_conroy_intel/1M_pre_ms_to_wd/err.txt

maybe it was a hiccup?

adamjermyn:
@evbauer The log from `1M_pre_ms_to_wd` on Cannon is a bit confusing. It looks like it ran identically to commit `e8d7003` until the `to_wd` inlist, and then it hits a funny termination code:
```

 model saved to end_agb.mod
save photos/x967 for model 12967
termination code: envelope_mass_limit

                 runtime (minutes), retries, steps        3.96         3     12967
__________________________________________________________________________________________________________________________________________________

       step    lg_Tmax     Teff     lg_LH      lg_Lnuc     Mass       H_rich     H_cntr     N_cntr     Y_surf   eta_cntr   zones  retry
  lg_dt_yrs    lg_Tcntr    lg_R     lg_L3a     lg_Lneu     lg_Mdot    He_core    He_cntr    O_cntr     Z_surf   gam_cntr   iters  
    age_yrs    lg_Dcntr    lg_L     lg_LZ      lg_Lphoto   lg_Dsurf   CO_core    C_cntr     Ne_cntr    Z_cntr   v_div_cs       dt_limit
__________________________________________________________________________________________________________________________________________________

      12967   8.228543   3663.543   3.110331   3.233228   0.535305   0.009719   0.000000   0.000000   0.302227  19.613621   2153      3
 2.5066E+00   7.997187   2.014839   2.479656   2.002908  -6.021734   0.525586   0.000000   0.622823   0.020079   5.385947      6
 1.2396E+10   6.117848   3.239976   2.078906 -99.000000  -8.735073   0.485000   0.355228   0.017872   1.000000  0.000E+00          hold
                               rel_E_err    2.7939705898930813D-13
                       log_rel_run_E_err       -8.0322499174891870

DATE: 2021-12-15
TIME: 16:51:46
finished inlist_to_end_agb_header
run inlist_to_wd_header
DATE: 2021-12-15
TIME: 16:51:46
DATE: 2021-12-15
TIME: 16:52:00

******************************************************************
failed to create final.mod when running inlist_to_wd_header
******************************************************************
```
Do you know what this is about?

evbauer:
Yeah, that's confusing. Pablo is probably right. This one should run identically to `main` anyway right? I'll have Cannon rerun that commit.

PR #358, opened 2021-12-20
Jws/add s13 be7 wk li7 rate

rjfarmer:
This switches the 7Be(e-,nu)7Li rate to that of Simonucci et al 2013, as REACLIB does not handle the lowT rate (at all).


rjfarmer:
Testhub link https://testhub.mesastar.org/jws%2Fadd-s13-be7_wk_li7-rate/commits/82c3eca

PR #364, opened 2022-01-19
adding sentence about ending inlist file with carriage return

abostroem:
I spent a few hours trying to debug starting the inlist for starting from an existing model only to finally discover the issue was that my file didn't end with a blank line (I'm sure this is obvious to regular fortran users). The inlist modeled in the documentation doesn't include this return and I'm not sure how to get the RST to render it, so I added a sentence to try to head off this issue.


evbauer:
Thanks! This looks helpful. This is definitely the sort of thing that can trip people up. Perhaps just change "carriage return (to an empty line)" to "blank new line", just to avoid using a term that some users might be unfamiliar with.

PR #366, opened 2022-01-20
fixing typos in docs 

earlbellinger:


earlbellinger:
Sorry, I somehow didn't realize I wasn't the only commit here, so you can decline this if desired

evbauer:
In general, we don't want to merge the release version of the docs over into `main` because is has some release-specific numbers in it. But obviously we do want your fixes! In this case I think I'd just do it this way (from `main`):
```
git checkout release/r21.12.1 docs/source/developing/contributing.rst
```
And then just commit that onto `main`.

earlbellinger:
Thanks! That was very helpful. For what it's worth I needed to do 

`git checkout origin/release/r21.12.1 docs/source/developing/contributing.rst`

which I realized using 

`git branch -a`

to list all the branches. I'll close this pull request.

PR #367, opened 2022-02-07
Add a fast test of astero's simplex optimisation

warrickball:
This PR adds a crudely set up run of `astero`'s `simplex` optimisation, purely to cover more code with the tests. The "observed" data is based on a solar-mass star around 1 Gyr and the tolerances very loose so that each set of parameters evaluated quickly (< 1 min). The test is limited to 7 evaluations and takes about 5 minutes on my machine.

I tidied up a little bit of relevant `astero` documentation while I was at it.

The new test is [passing](https://testhub.mesastar.org/whb%2Fastero-fast-simplex-test/commits/14e20bb) and [documented](https://github.com/MESAHub/mesa/blob/whb/astero-fast-simplex-test/astero/test_suite/fast_simplex/README.rst).


rjfarmer (commented):
Why not have  a results_folder option? Then we can make it with a call to mkdir (from utils_lib). That way no one needs to be making directories themselves.

warrickball:
> Why not have a results_folder option? Then we can make it with a call to mkdir (from utils_lib). That way no one needs to be making directories themselves.

Agreed! But that's for another PR.

rjfarmer:
Well every other output type in mesa makes its own folders, wouldn't want astero to be left out on its own :)

warrickball:
@rjfarmer I've merged this so I can now more confidently start tidying up a bunch of things (including folder creation) in subsequent PRs when I have some more MESA time soon.

PR #369, opened 2022-02-08
White Dwarf Carbon-Oxygen phase separation

evbauer:
This PR implements an inlist option that can enable carbon-oxygen phase separation upon crystallization in cooling white dwarf cores, along with the associated mixing and energy release that influences the cooling timescale.

This includes two new mixing types to mark zones that are crystallized (`mixing_type = -1`) and zones that are mixed by fluid instability due to composition changes in the liquid above the crystallized core (`mixing_type = phase_separation_mixing`). I have not yet added these mixing types to the pgstar plots such as the Kippenhahn window. There are already many mixing types and the legend is getting rather crowded, so it may be better not to add these two cases that will very rarely show up.

This PR also includes some changes to the default evaluation of the Brunt B term which should make it more efficient. Previously calls to `do_brunt_B` contained `nzlo` and `nzhi` arguments that were usually ignored, because the subsequent call to `do_brunt_B_MHM_form` dropped those arguments and always evaluated B over all zones in the star. For phase separation, this added quite a lot of expense due to the iterative evaluation of the fluid instability outside the crystallizing core that relies on the Brunt B term. In this PR, `do_brunt_B_MHM_form` now respects the `nzlo` and `nzhi` arguments, so that repeated calls to `do_brunt_B` for a specific region where abundances have been updated are much less expensive.

I intend to keep this branch open after merging to continue development, but C/O phase separation appears to be sufficiently stable to try merging into `main` so others can start using it.


fxt44:
nice!

rjfarmer:
Can we get a changelog entry?

warrickball:
> This includes two new mixing types to mark zones that are crystallized (`mixing_type = -1`) ...
Why not a positive integer? Do we skip mixing if `mixing_type <= 0`?

evbauer:
We don't explicitly check something like `mixing_type <=0` as far as I know, but it seems like we could. All the positive mixing types correspond to some amount of mixing, so I thought it might be nice to use negative integers to designate something we're labeling as not doing any mixing (for some more specific reason than `no_mixing = 0`). Happy to change if that runs afoul of some other convention I'm unaware of though.

PR #376, opened 2022-02-24
Simple fix to 're' sometimes not finding the most recent photo

simonguichandut:
I was having some problems with `re` not being able to automatically find the most recent photo.  The fix was simple, and should not change anything for people who did not have these issues.

- ` \ls` to force the "non-aliased" `ls` command.  I like to alias `ls=ls -S` in my bash_profile to always sort by file size, but this will mess-up the search for the most recent photo with the `-t` option.

- Add the `-p` option to add a "/" to directories, then `grep -v /` to ignore those and only list files. I sometimes store photos for backup in e.g. `photos/runA/x500`, so I just want to ignore these directories and only look at files to find the most recent photo.

- Printing `($photo)` in case of problems to help debugging.

### DIFF
--- a/star/work/re
+++ b/star/work/re
@@ -3,7 +3,7 @@
 photo_directory=photos
 
 function most_recent_photo {
-    ls -t "$photo_directory" | head -1
+    \ls -tp "$photo_directory" | grep -v / | head -1
 }
 
 if [ $# -eq 0 ]
@@ -15,7 +15,7 @@ fi
 
 if [ -z "$photo" ] || ! [ -f "$photo_directory/$photo" ]
 then
-    echo "specified photo does not exist"
+    echo "specified photo ($photo) does not exist"
     exit 1
 fi
 


rjfarmer:
Thanks,

I'm guessing we would actually want to add to the script

````
shopt -u expand_aliases 
````

So we stop all aliases being expanded, not just ls

fxt44:
shopt within a bash shell.

rjfarmer:
I would suggest we have instead:

````bash
#!/bin/bash

shopt -u expand_aliases 

photo_directory=photos

function most_recent_photo {
    ls -tp "$photo_directory" | grep -v / | head -1
}

if [ $# -eq 0 ]
then
    photo=$(most_recent_photo)
else
    photo=$1
fi

if [ -z "$photo" ] || ! [ -f "$photo_directory/$photo" ]
then
    echo "specified photo ($photo) does not exist"
    exit 1
fi

echo "restart from $photo"
if ! cp "$photo_directory/$photo" restart_photo
then
    echo "failed to copy photo ($photo)"
    exit 1
fi

date "+DATE: %Y-%m-%d%nTIME: %H:%M:%S"
./star
date "+DATE: %Y-%m-%d%nTIME: %H:%M:%S"
````

PR #378, opened 2022-03-04
Pre compute some powers of (integer) Z

rjfarmer:
Many places in the EOS (and screening) compute powers of the charge Z
but given there is only a small number of Z's we can just precompute
the terms (which are saved into a new math_def module).

This saves about 5% runtime on a test case I ran.


adamjermyn:
The Skye side looks good to me!

Issue #328, opened 2021-09-24
State is unclear in the `star` pointer.


adamjermyn:
## Intro

I've been trying to understand what makes `state` so tricky in MESA/`star`. This is my attempt at expressing what's wrong, and what we might do about it.

Everything here has some caveats, the biggest of which is a 'so far as I know' that should be attached to every statement about MESA.

## Current State

There are at least six different kinds of state that live in `star_data`:
1. Metadata. Flags and configuration, mostly defined in `star_data_step_input.inc`.
2. The 'ground truth' structure variables. E.g. `xh` (structure), `xa` (composition), `mlt_vc` (for TDC).
3. Starting values. Used for evaluating time derivatives.
4. Copies of structure variables. e.g. `s%lnR` is a copy of `s%xh(i_lnR)`. Presumably this is for convenience and code clarity. Note that these are copies and not pointers.
5. Intermediate results. For instance results of `eos` calls or intermediate calculations like `csound` and `z53bar`. 
6. Outputs that go in history or profiles or pgstar windows or which are used for debugging.

There are also at least three different problems we routinely hit with state:
1. Sometimes an output becomes an intermediate result when a piece of code is written that relies on it. These may need updating at different points in time, resulting in inconsistencies like #66.
2. Different kinds of state need to be updated at different times when `star` performs iterations, starts a step, finishes a step, retries, model writing/reading, etc. This was a major source of bugs in the development of TDC, and even determining when a particular piece of state gets updated is challenging/time-consuming.
3. It is not clear which intermediate results depend on which other intermediate results, and what those dependencies look like. This makes it easy for things to get out of sync. This came up in developing `eps_mdot`, where at one point `s%R` and `s%lnR` got out of sync.

## Proposed Solution

I am not sure of the best solution here, but the best I've come up with is to the data inside `star_data` into six sub-types:
1. `meta`
2. `struct`
3. `start`
4. `ease`
5. `inter`
6. `output`
These correspond to the different kinds of state above.

A few examples:
- `s%v_flag` becomes `s%meta%v_flag`
- `s%xa` becomes `s%struct%xa`
- `s%lnR` becomes `s%ease%lnR`
- `s%mlt_vc_old` becomes `s%start%mlt_vc`
- `s%eps_mdot` becomes `s%inter%eps_mdot`
- `s%center_ye` becomes `s%output%center_ye`

[...]

PR #380, opened 2022-03-09
Move pgstar data intro its own derived type

rjfarmer:
As part of #328 this moves all pgstar inlist options (as well as its state) into a new pgstar_controls derived type. Inside star_type this is then stored as s% pg. As part of this organization there is now also a star_pgstar.f90 file which should now be holding all the other pgstar related variables and parameters.


fxt44:
wow, very nice rob.

Issue #375, opened 2022-02-23
`add_atmosphere_to_pulse_data` does not work properly with an Eddington atmosphere 


earlbellinger:
Running MESA with `add_atmosphere_to_pulse_data = .true.` causes 

```
save LOGS/profile1.data atm: eval_T_tau_uniform logT from eval_data                       NaN
```

This is evidentally caused by the atmosphere module expecting `Teff` as an input. Since it's not set, it's NaN. 

If one sets `atm_T_tau_opacity = 'varying'` then it crashes like so:

```
 atm: Call to eos_proc failed in eval_fcn
 Call to dopri5 failed in eval_data_try: idid=          -5
 atm: Call to eos_proc failed in eval_fcn
 Call to dopri5 failed in eval_data_try: idid=          -5
 atm: Call to eos_proc failed in eval_fcn
 Call to dopri5 failed in eval_data_try: idid=          -5
 atm: Call to eval_data failed in atm_t_tau_varying
 atm: Call to eos_proc failed in build_data
 atm: Call to dopri5 failed in build_T_tau_varying: idid=          -5
 failed in build_atm
 star_export_pulse_data ierr          -1
 do_saves terminate_normal_evolve_loop ierr          -1
 terminate_normal_evolve_loop ierr          -1
 do_evolve_one_step ierr          -1
```

Minimal inlist to produce these bugs:
```
&star_job
    steps_to_take_before_terminate = 2
/

&eos
/ ! end of eos namelist

&kap
  Zbase = 0.02
/

&controls
    add_atmosphere_to_pulse_data = .true.
    !atm_T_tau_opacity = 'varying'
    write_pulse_data_with_profile = .true.
/
```

warrickball:
I'll self-assign and work on this when I have some time after 7 March unless someone else wants to pick it up first. It's almost certainly a hangover from the interfaces changing `Teff` from an output *from* `atm` to an input *to* `atm`. It can probably fixed by mimicking the calls to get the atmospheric boundary condition during the evolution.

mjoyceGR:
I think @adamjermyn and I might have found either this or a related problem back in October or November (?) (not volunteering myself to work on it, just a note)

warrickball:
Should be fixed by #381, which I just merged.

PR #381, opened 2022-03-09
Fix `build_atm` to take Teff as input

warrickball:
Some time ago, `Teff` changed from an output from `atm` to an input to `atm`, because `atm` doesn't necessarily know where the photosphere of a star is. Many of these changes weren't reflected in the `build_atm` routines that are used to compute the atmospheric structure that's appended to the pulsation formats (e.g. FGONG), as noted in #375. This PR fixes the immediate issues and completes the transition. It also adds a new test case `check_pulse_atm` that writes the FGONG file, then reads it back in and checks that the atmosphere structure matches the selected T(τ) relation.

In the future I'll extend the test to check other formats, too, but for now it only checks FGONG.

I'll leave this open for a few days for comments.

### DIFF
--- a/atm/private/atm_t_tau_uniform.f90
+++ b/atm/private/atm_t_tau_uniform.f90
@@ -261,7 +261,7 @@ contains
   ! opacity
 
   subroutine build_T_tau_uniform( &
-       tau_surf, L, R, M, cgrav, kap, Pextra_factor, tau_outer, &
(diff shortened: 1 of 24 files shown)


mjoyceGR:
not close to merging it yet, but just a heads up that I have a new test_suite case that this will likely affect as well (test_suite/starspots).

PR #387, opened 2022-03-17
Move controls to their own derived type.

rjfarmer:
This moves all controls to s% ctrl. This will break (almost) everyone's
run_star_extras.f90. So there is a python script in linters/update_ctrls.py
which will do the transformation for you. Though the script does not
work when s was renamed (say in the binary module).

This works on #328


rjfarmer:
Testhub https://testhub.mesastar.org/rf%2Fctrl/commits/d0c56dd

rjfarmer:
After chatting with adam we are going to hold on merging this until we do move the other variables into their on struct (and update the python script to handle them). That way we can limit the number of times we break things.

Issue #377, opened 2022-03-02
1M_thermohaline fails with TDC


adamjermyn:
`1M_thermohaline` fails with TDC as of commit `91aecec`.

This is surprising because:
1. It passes with Cox.
2. The time-steps are 100kyr+, which should be long enough that TDC and Cox agree to near machine precision.

There are some thermohaline errors at the beginning of a run with TDC, but @evbauer tells me this shouldn't be an issue.

The various rotational mixing processes seem to give very different answers when TDC is on versus off. I'm not sure if that's the cause or effect here. In any case, this seems to make life difficult for the solver trailing the burning layer in the core, so the timestep gets pushed down when TDC is on.

One thought I had is that  thermohaline and semiconvection both set non-zero convection speeds. TDC then sees these and thinks “Ah there must have been convection here last step!” and in principle that can change the answer it returns, though I think that’s rather unlikely to cause this given how long the time-steps are compared with the convective decay time…

If this is the cause though it’s not obvious we want to change it. It’s worse for making 1M_thermohaline pass, but it might be more physical…

Anyway that seemed unlikely, so I added some instrumentation to `turb_support.f90` to look for differences:
```
         call set_TDC(&
            conv_vel_start, mixing_length_alpha, s% alpha_TDC_DAMP, s%alpha_TDC_DAMPR, s%alpha_TDC_PtdVdt, s%dt, cgrav, m, report, &
            mixing_type, scale, chiT, chiRho, L, r, P, T, rho, dV, Cp, opacity, &
            scale_height, gradL, grada, conv_vel, D, Y_face, gradT, s%tdc_num_iters(k), ierr)

         call set_MLT("Cox", mixing_length_alpha, s% Henyey_MLT_nu_param, s% Henyey_MLT_y_param, &
                        chiT, chiRho, Cp, grav, Lambda, rho, P, T, opacity, &
                        gradr, grada, gradL, &
                        Gamma2, gradT2, Y_face2, conv_vel2, D2, mixing_type, ierr)

         if (abs(conv_vel%val - conv_vel2%val) > 1d2 .or. abs(Y_face - Y_face2) > 1d-3) then
            write(*,*) 'Diff',k,s%m(k)/Msun,conv_vel%val, conv_vel2%val, 0.5d0*(conv_vel%val - conv_vel2%val)/(conv_vel%val + conv_vel2%val), &
[...]

PR #388, opened 2022-03-17
Turn TDC on by default.

adamjermyn:
Now that TDC reduces to Cox*, we can turn it on by default and so cleanly handle cases where a model goes from not needing TDC to needing it.

This PR turns TDC on by default and then selectively disables it in test cases where that caused problems (see #377). It also incorporates the `gradr_factor` machinery into TDC, so that it follows that logic the same way as MLT.

I would also be happy incorporating this PR but *not* turning TDC on by default, in which case the point is the `gradr_factor` fixes in TDC.

*The only remaining discrepancy between TDC and Cox that I'm aware of is in the treatment of the pressure scale height near the center of the model, where Cox-in-MESA uses an alternate scale height prescription and TDC continues to use (implicitly) `h = P / rho g`. In #377 there's some discussion of what would be involved in removing this discrepancy, and we can do that work if it turns out to matter...


adamjermyn:
If the latest commit on this branch (`72f3b29a`) passes on Testhub I think this is good to merge.

adamjermyn:
`make_o_ne_wd` was the last holdout and passes, so we're all good to merge.

PR #389, opened 2022-03-18
Cleanup up rate selection

rjfarmer:
Series of improvements to the rates code:

- Removes the ability to set which_rates. Now we default to NACRE and REACLIB but don't try to blend between them.
- Adds a new other_rate_get hook for modifying existing reactions
- Adds a other_net_derivs hook that allows for arbitrary reactions (formally you get access to dydt so you can change the composition however you want)
-  Reactions can now be set via an inlist option filename_of_special_rate
- Ship some of the rates removed when I removed the which rates options as rate tables
- Add NACREII rates as extra rate tables that can be loaded

Todo:
- Fix test suite
- Add back FL3a using other_net_derivs

Also merging will be fun as this will break @earlbellinger work on the rate output and is also will be broken by #387. So we need to merge earl's work first, then this, then the star ctrl changes.


fxt44:
interesting that parallelism seems to be a culprit.

rjfarmer:
I think it's more that things get moved around in memory differently now and we avoid whatever was causing the memory corruption.

warrickball (approved):
I haven't looked closely at the rates changes but the new `example_astero` parameters look fine.

warrickball:
Several tests have [failed](https://testhub.mesastar.org/rf%2Frates2/commits/1a870f2) on `bluebear_ifort`. I won't be able to look into this until April 25. If a few more systems could run the tests, we can see whether it's one machine acting up or a real bit-for-bit problem.

fxt44:
i'm rolling on 4 machines now ...

evbauer:
I just pushed an empty `[ci optional]` commit so that hopefully we can get a more apples-to-apples comparison between cannon and bluebear.

evbauer:
It looks like cannon agrees with bluebear about four of the reported failures with `[ci optional]`.
https://testhub.mesastar.org/rf%2Frates2/commits/f80496c

rjfarmer:
Everything's passing now:

https://testhub.mesastar.org/rf%2Frates2/commits/30e7518
https://testhub.mesastar.org/rf%2Frates2/commits/5c998c6

PR #390, opened 2022-03-18
Enabled Uq in TDC.

adamjermyn:
Uq is the turbulent viscosity term in RSP & TDC. Currently it's only on in RSP, but all the logic is there. This PR enables it for TDC. If this passes on Testhub I think it can be merged.


adamjermyn:
Nevermind that broke tons of things.

PR #393, opened 2022-03-25
fix to binary b_fm timestep control 

matthiasfabry:
the binary timestep control b_fm might be unjustly tripped if the donor was switched.

If the donor is switched at the start of the step, b% env_old is not updated to reflect the envelope mass of the new donor.


matthiasfabry:
Here a simple testcase to reproduce the problem

[fm_fix_example.zip](https://github.com/MESAHub/mesa/files/8361285/fm_fix_example.zip)

PR #399, opened 2022-04-08
Caplan diffusion coefficients for WD interiors

evbauer:
Implements diffusion coefficients of Caplan, Bauer, & Freeman (2022)
https://ui.adsabs.harvard.edu/abs/2022MNRAS.tmpL..33C/abstract

By default sets these to be the coefficients used in strongly coupled matter (e.g. white dwarf interiors approaching crystallization), with a smooth transition from the Stanton & Murillo coefficients over Gamma=5-10. Caplan coefficients fully on for Gamma>10.


star/private/diffusion_support.f90
```diff
@@ -1666,6 +1681,132 @@ subroutine lam_SM(nc,m,rho,T,Z,nd,lam_eff,kappa)
...
+        omegap = sqrt(pi4*ni_tot*Zbar*Zbar*qe*qe/(Abar*amu)) ! plasma frequency
+        ai = pow(3d0/(pi4*ni_tot),one_third)
+        kappa = ai/lam_e
+
+      end subroutine kappa_CBF
+
+      ! crlibm does not include error function, so implement an
+      ! erf fit good to 2.5e-5 (Abramowitz and Stegun 1964)
+      ! in terms of other bitsafe functions.
+      real(dp) function bitsafe_erf_fit(x)
```
warrickball, line 1797: Do we want to somehow implement the bit-safe error function in a more basic module (e.g. `math`) so it's available to other parts of MESA and in `run_star_extras.f90`?

warrickball:
This needs a changelog entry, especially because it changes the default.

star/private/diffusion_support.f90
evbauer: That sounds potentially reasonable, but perhaps we can raise this as a rainy-day issue after merging this? Also, is 2.5e-5 accurate enough to use as a general drop-in replacement for `erf`? It's certainly fine for this particular application, but I'm not sure I'd want to add this to math in a way that users might not realize they're not using an approximation.

rhdtownsend:
Have we established that the Fortran intrinsic error function is not bitsafe?

On Apr 12, 2022, at 3:26 AM, Warrick Ball ***@***.******@***.***>> wrote:



@warrickball commented on this pull request.

________________________________

In star/private/diffusion_support.f90<https://github.com/MESAHub/mesa/pull/399#discussion_r848133966>:

>
+        ! Compute kappa
+        ni_tot = 0d0
+        do i = 1,nc
+           ni_tot = ni_tot + nd(i)
+        end do
+        omegap = sqrt(pi4*ni_tot*Zbar*Zbar*qe*qe/(Abar*amu)) ! plasma frequency
+        ai = pow(3d0/(pi4*ni_tot),one_third)
+        kappa = ai/lam_e
+
+      end subroutine kappa_CBF
+
+      ! crlibm does not include error function, so implement an
+      ! erf fit good to 2.5e-5 (Abramowitz and Stegun 1964)
+      ! in terms of other bitsafe functions.
+      real(dp) function bitsafe_erf_fit(x)


Ah, that's fair. As a user, I would expect a more accurate expression in math or elsewhere.

evbauer:
I'm not 100% sure, but I think the argument is that since `erf` is a relative of `exp`, and we need an `exp` entry in `crlibm`, we should expect that `erf` would also not be bitsafe. And @adamjermyn indicated that he may have had some more direct experience with bit-for-bit issues using `erf` in his work with Alex Dittmann?

warrickball:
> Have we established that the Fortran intrinsic error function is not bitsafe?

Yes. I tried calculating erf with random reals with gfortran and ifort and got bit-level differences about once per hundred arguments.

rhdtownsend:
I thought our bit-for-bit concern was about portability across platforms, not across compilers — no?

On Apr 12, 2022, at 1:22 PM, Warrick Ball ***@***.******@***.***>> wrote:



Have we established that the Fortran intrinsic error function is not bitsafe?

Yes. I tried calculating erf with random reals with gfortran and ifort and got bit-level differences about once per hundred arguments.

warrickball:
> I thought our bit-for-bit concern was about portability across platforms, not across compilers — no?

I thought our bit-for-bit concern was about portability across compilers _and_ platforms...

evbauer:
It looks like discussion on this has died down, and we may be ready to move the `erf` discussion to a separate github issue and merge this PR. Any objections to going ahead and merging as long as [this](https://testhub.mesastar.org/Caplan_diffusion/commits/3cc34dd) comes back clean?

PR #401, opened 2022-04-17
Add a numerical integrator

rjfarmer:
This is based on simpsons 3/8 rule but runs with adaptive step sizes
with both a relative or absolute tolerance control


adamjermyn (approved):
Looks good to me. I didn't check the rule in detail but did look at the test output, the integrator, and the testing functions, and that all looks great.

PR #405, opened 2022-04-21
Add neutrino capture rates

rjfarmer:
Start on adding neutrino capture rates.

Its slow as molasses going uphill but its a starting point.

Also no idea yet if its actual spitting out sensible numbers

### DIFF
--- a/clean
+++ b/clean
@@ -16,6 +16,7 @@ rm -rf data/kap_data
 rm -rf data/chem_data
 rm -rf data/colors_data
 rm -rf data/rates_data/cache
+rm -rf data/neu_data
 
 rm -rf data/rates_data/cache/*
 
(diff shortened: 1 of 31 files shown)


fxt44:
cool. what's the target application(s) ?

rjfarmer:
Core collapse nucleosynthesis, at least initially. Idea being that the user specifies some neutrino flux(space,time) distribution (so we ignore the full complexity of neutrino transport) and then we see what happens.

fxt44:
fluorine is fun! where are the artes coming from?

rjfarmer:
I'm following what skynet does (their appendix C)

fxt44:
ahh, hence the added integration routine.

rjfarmer:
Yup, though that I think will need replacing already with something better. It needs to take alot of sub steps to even begin to converge on a value

fxt44:
well, there are fermi functions in there, so not unlike in an eos. a tuned quadrature scheme may do well. alternatively tabulate off-line rather than on the fly.

rjfarmer:
You know after plotting the rates, they are smooth enough that we can just pre-compute some polynomial fits in python and load the coefficients into mesa. 

Much simpler and also alot easier to get the derivatives out.

fxt44:
way easier! probably accurate 'enough".

PR #406, opened 2022-05-06
Refactor parameters and "variables" (i.e., observed constraints) in `astero`

warrickball:
The `astero` module sets up an optimization problem to find some stellar parameters that best fit data for some observed quantities, which I'll refer to as *variables*, because that's what the code calls them (more in a minute). The current version allows tables of mode frequencies and some "spectro" (really non-seismic) data for Teff, logg, [Fe/H], logL, logR, Z/X, Rcz and Ysurf (where I hope the symbols' meanings are obvious), as well as space for three user-defined variables. The parameters are mass, initial Y, initial FeH, MLT α, overshoot f_ov and up to three user-defined parameters.

The main purpose of this PR is to refactor the hardcoded parameters and variables into user-extensible lists that are as short or long as necessary up to some generous maximum sizes (currently 100). This refactor has two advantages:
1. it deletes a lot of code that was duplicated for the fixed parameters and variables; and
2. it makes `astero` more versatile. (Want to fit 12 parameters to 15 custom non-seismic constraints? No problem!)

The parameters and variables are now set in the routines `set_my_vars` and `set_my_params` in `run_star_extras.f90`. The default implementation preserves the old behaviour (though I renamed `FeH` to `M_H`) and falls back to any history column if no user-defined implementation is found.

Along the way, I also cleaned up the test cases' input files, which were largely duplicating controls not used in the tests, and old documentation that's floating around with it.

Some legacy controls (e.g. `Z_div_X_solar` and stuff to set Y based on Z) persist to keep the old parameters and variables working.

I'll write a changelog entry once the discussion points below are resolved (or no-one stops me from doing what I want to).

In the long run, I hope to write module-level documentation for `astero` but that needn't be associated with this PR. I've tried to update the documentation in the defaults but comments are welcome.

# Discussion points

Apologies if this is an abuse of PRs rather than raising issues but this is a good time to address interface changes.

### 1. What should we call the "variables"?

These are the numbers that usually correspond to observed non-seismic properties. The choice "variables" is inherited from the custom options that are already in the code but I think that's a bad choice because there are many different ways things can vary. My personal choice is to call them "constraints", which is what I hope to implement in the next few working days if no-one stops me.

### 2. Dependent parameters

The implementation in this PR (inherited from the existing implementation of user controls) assumes that each parameter can be set independently but there might be some weird cases where this isn't true. For example, the old controls used an initial Y and [Fe/H] to set the initial X and Y abundances. So, working through parameters one by one, first we set the initial Y. Then, to set the initial X using [Fe/H], the code inspects the Y that's just been set, combines that with [Fe/H] and `Z_div_X_solar` to derive X and uses that. There's a possibility that someone will cook up a pair of variables that, because they don't map neatly onto MESA `star` controls, have to be set simultaneously.

I'm inclined to leave things as they are in this PR and see if anyone actually gets tripped up by this. There are various ways of hacking around it but they all strike me as complicating this code for a fringe case.

### 3. Derived types?

The parameters and constraints each have a fixed set of properties. e.g. parameters have: a name; whether they're being optimized; their initial, minimum and maximum values; and a spacing for the `grid_search` option. Part of me is inclined to bundle these up into a derived type so the user would set something like
```f90
    param(1)%name = 'initial_mass'
    param(1)%vary = .true.
    param(1)%first = 1.0
    param(1)%min = 0.5
    param(1)%max = 1.5
    param(1)%delta = 0.01
```
instead of
```f90
    parameter_name(1) = 'initial_mass'
    vary_parameter(1) = .true.
    first_parameter(1) = 1.0
    min_parameter(1) = 0.5
    max_parameter(1) = 1.5
    delta_parameter(1) = 0.01
```
Although using types seems intuitive to me, there's little practical gain, and perhaps a bit of user confusion. I raise it now only because if it's worth doing, it's worth doing *now* while we're breaking the interface anyway.

For completeness, I envision the constraints/variables being set by something like
```f90
    constraint(1)%name = 'luminosity'
    constraint(1)%target = 1.0
    constraint(1)%sigma = 0.001
    constraint(1)%include_in_chi2_spectro = .true.
```


wmwolf:
Very cool! Regarding the discussion points, I think "constraints" is a valid choice. My first guess was "observables". When I think of "constraint" in this context, I think of limits on the parameters rather than the data you are trying to match, but I can see your intent, and I don't think that "constraints" is too confusing.

The derived types idea is tidy and neat (and I prefer it), but do we do anything else like this elsewhere in MESA? It seems like the second option is more commonly found today (see overshoot controls). Perhaps with changes to state in our future, we'll be getting more comfortable relying more on derived types, so I'm open to being convinced. Also, is the context of your example a `run_star_extras.f90` file, or are we allowed to set type attributes in namelists? I thought that namelists could only be of the format `NAME = VAL` or `NAME(INDEX) = VAL`, but that may jube bias because we don't use types inside of namelists currently.

warrickball:
> My first guess was "observables".

I've at least once used `astero` to create a model that matched the properties of another model, which wasn't "observed" even though I don't think it would've been confusing to call them observables.

> ... are we allowed to set type attributes in namelists?

I did some basic tests that imply that this can be done in namelists, which is the context I had in mind.

> The derived types idea is tidy and neat (and I prefer it), but do we do anything else like this elsewhere in MESA?

Not in inlists, as far as I could tell.

wmwolf:
> > My first guess was "observables".

> I've at least once used `astero` to create a model that matched the properties of another model, which wasn't "observed" even though I don't think it would've been confusing to call them observables.

Maybe "targets"? Ok, I'll stop.

warrickball:
The "variables" are now "constraints", since I think that makes most sense.

I'm not going down the derived type road now. There are a few too many special cases for the constraints (e.g. `delta_nu`, `numax`, `age`) for the interface to come out clean enough to warrant the change.

So this is now quite close to ready to merge, in my opinion. The tests are passing locally but I'll wait for TestHub to be all green.

Issue #76, opened 2019-10-24
Develop coding style guide


jschwab:
It would be useful to have a coding style guide.  This could cover information about formatting as well as style (do loops vs arrays).

matthiasfabry:
I'd like to bump this by saying lots of `.f90` files aren't formatted free-form style, but rather fixed-form style.
I'd argue free-from dramatically improves readability, e.g. look at `star/public/star_def.f90`:
```
      module star_def
      
      use star_data_def

      
      implicit none      
      
      
      end module star_def
```
vs.
```
module star_def

   use star_data_def

   implicit none

end module star_def
```
I might be overly fanatic about this, but I feel the code base could benefit from this. I can try a bulk reformat if this is something we'd like to see happen.

rhdtownsend:
I wholeheartedly approve of such a change; much of the current source code is (visually) an abomination.

rjfarmer:
@matthiasfabry if you are going to insist on whole codebase changes please only change the Fortran files. I've seen your [free-format-formatting](https://github.com/MESAHub/mesa/tree/free-form-formatting) branch and you are changing every file including those that aren't Fortran files. It's going to be difficult to merge if you insist on changing every file at once. Instead, go for one module at a time so there is even a slight chance we can review changes before merging.

Please also don't change the code in the stella folder. No one here can fix things if you break something in it.

matthiasfabry:
yes, that makes sense from a review standpoint. I can restart and make a folder of branches.

Also good to know `stella` is off limits

matthiasfabry:
I've created formatting/astero, /auto_diff, /binary and /const for you to review. I've tried to be as careful as possible for the formatter to only touch `.f90` files.

evbauer:
The fortran code in `auto_diff` is actually generated by python scripts from the `python` directory in that module, so we should generally avoid direct edits to the `auto_diff/private` files. If you need to make edits to `auto_diff`, you should edit python files and then regenerate the fortran as necessary.

I guess this isn't particularly well documented either, so maybe we should try to add some documentation somewhere to help developers navigate editing `auto_diff` if necessary. I'll admit I'm not very clear on that at the moment.

earlbellinger:
Here's some documentation on auto_diff:
https://adamjermyn.com/posts/auto_diff_mesa/

Best wishes,
Earl

evbauer:
[...]

PR #408, opened 2022-05-18
Add `ierr` output to `star_lib:star_get_history_output` (and maybe similar procedures)

warrickball:
I recently made use of `star_lib`'s `star_get_history_output` function, which has the call signature
```f90
    real(dp) function star_get_history_output(s, name)
```
If it fails to find a history column with the requested `name`, it returns `-HUGE(0.0d0)`. This PR changes the call signature to
```f90
    real(dp) function star_get_history_output(s, name, ierr)
```
where `ierr` is optional and, if included, returns `ierr = 1` if the column wasn't found so that the caller can more easily handle a failure.

This isn't the only procedure where I can't see a robust way of handling errors. Others that caught my eye are
```f90
    real(dp) function star_get_history_output_by_id(id, name) ! closely related
    real(dp) function star_get_profile_output(s,name,k)
    real(dp) function star_get_profile_output_by_id(id, name, k)
```
I'm happy to be dissuaded from changing these too but I think we want to discuss it. We [started discussing `ierr` more generally](https://github.com/MESAHub/mesa/issues/76#issuecomment-718084612) a bit in issue #76 but it was a digression there.

Obviously, I'd personally like to have `ierr` in these routines. Some routines don't have `ierr` but have unambiguously garbage output, which is also fine by me. e.g., `star_history_specs` returns indices < 0 if it didn't find something.

If anyone really doesn't like impure functions (i.e., functions that modify arguments), I can change these to subroutines.

Or I can unpack the machinery in `star_get_history_output` into my own routine and handle the failing case there. As it is, implementing `ierr` as an optional argument doesn't impede the purely functional way of doing things.

### DIFF
--- a/astero/test_suite/astero_adipls/src/run_star_extras.f90
+++ b/astero/test_suite/astero_adipls/src/run_star_extras.f90
@@ -79,7 +79,8 @@
             !    val = s% delta_Pg
             ! fall back to history column if user doesn't define name
             case default
-               val = star_get_history_output(s, name)
+               val = star_get_history_output(s, name, ierr)
+               if (ierr /= 0) call mesa_error(__FILE__, __LINE__)
          end select
 
       end subroutine set_constraint_value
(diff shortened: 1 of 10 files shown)


rjfarmer:
They should all get changed (or all not changed) as the by_id are just there so its easier to hook into MESA without needing to handle s.

Otherwise i think this is a good idea.

PR #414, opened 2022-06-07
Fix to b% fm to account for the donor switching during the step

matthiasfabry:
the binary timestep control b_fm might be unjustly tripped if the donor was switched.
If the donor is switched at the start of the step, b% env_old is not updated to reflect the envelope mass of the new donor.

made b% env an array so both stars can have their envelope stored


matthiasfabry:
Test package highlighting the issue
[fm_fix_example.zip](https://github.com/MESAHub/mesa/files/8854181/fm_fix_example.zip)

PR #420, opened 2022-06-08
Make cumulative energy error add absolute value of error from each step

evbauer:
It seems misleading to allow positive and negative energy errors from subsequent steps to cancel out when adding up the cumulative energy error, so I propose that we always add the absolute value of energy error from each step. This helps some plots I'm working on for the MESA VI energy section make more sense.

@rsmolec I had to relax the tolerance on the `rel_run_E_err` in the run_star_extras for `rsp_Type_II_Cepheid` for it to pass with this new calculation of cumulative energy error.

### DIFF
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -11355,6 +11355,17 @@
 
     warn_when_large_rel_run_E_err = 0.01d0
 
+      ! absolute_cumulative_energy_err
+      ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+      ! cumulative energy error is the sum of the absolute value of per-step errors.
+      ! Set this to .false. to allow positive and negative errors to cancel
+      ! when integrating over multiple steps.
+
+      ! ::
+
+    absolute_cumulative_energy_err = .true.
+
 
       ! warn_when_stop_checking_residuals    obsolete
       ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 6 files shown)


evbauer:
The per-step energy errors already represent total energies summed over the whole model. So we're allowing negative error from one solution to cancel out positive error from another solution, and those errors probably aren't even coming from the same regions of the star. Actually, in some cases that cancellation in the cumulative error can happen thousands of steps apart when the star is in completely different phases of evolution.

Negative or positive, I see it as error either way, and it seems very unlikely in general that a later error could precisely compensate for an earlier error in any physically meaningful way. If there's an exception I'm not thinking of here, it's not hard to implement a flag as you suggest, but right now I'm not seeing a good case for this.

fxt44:
ok. allow me to rephrase. my suggestion is add, not replace.
both metrics have their merits, so to me its about adding capability instead of removing capability.

one can imagine starting on a thermodynamic manifold, taking one timestep,and wondering if on average one is still on the same manifold beacause of conservation laws. now take many timesteps. one may want to know, on average over mass and time, by how much (and why!) the final manifold is different that the starting manifold as measured by the energy. for these averages, one may want the positive and negative energy errors. i stress the discussion in this paragraph should be independent of a decision on adding or removing/replacing software capability.

evbauer:
Thanks, that definitely helps me see what you're getting at. What you describe does sound well-defined and physically meaningful, but seems quite different from my conception of what "energy error" means in the context of MESA, where we've generally described it as saying something about the quality of solutions produced by our evolution steps.

Happy to implement an inlist flag to give an easy option to do things either way. I'd just argue that the default should be the new behavior I'm advocating for here, since that's more consistent with the way we've described energy error in instrument papers. Does that sound ok to you?

fxt44:
i'm just advocating for options. fine by me to change the default option!

evbauer:
This now comes with the requested user option to change the definition, and is [passing the test suite](https://testhub.mesastar.org/absolute_energy_error/commits/4f4779a). Ok to go ahead and merge?

PR #422, opened 2022-06-15
Add wrapper for accessing various rate outputs

rjfarmer:
No access to screened rates yet as that needs call to screen_net to
hanlde all the special rates (otherwise we could just use screen_pair)
and thats hard.


earlbellinger:
I think I've implemented this into the profile and history outputs now. It at least seems to match what you got :-) Please let me know if anything looks off!

earlbellinger (approved):

rjfarmer:
Closing this as #450 now works much better for doing this

PR #427, opened 2022-07-12
Op mono and radlev

evbauer:
Merging branch from Joey's fork onto a MESAHub branch for testing. Once it passes testing, we can merge that branch into main.

### DIFF
--- a/controls.defaults
+++ b/controls.defaults
@@ -1,0 +1,11771 @@
+! ========
+! controls
+! ========
+
+
+! specifications for starting model
+! =================================
+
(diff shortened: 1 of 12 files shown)


fxt44:
test suite rolling on 1151259 ...

evbauer:
You can hold off on testing at the moment. Those first couple commits only have some files uploaded in the top level directory, so the changes aren't actually applied at that stage. I did some work in the next few commits to move everything into place and apply the changes, but it looks like we're still missing a few minor things before we're ready to compile and run.

fxt44:
right ...

PR #428, opened 2022-07-15
Mombarg OP Mono Routines

evbauer:
Work in progress to merge Joey's changes for OP Mono and radiative levitation. Not yet passing the test suite because we need to upload a supplemental data file like we have for the other OP Mono data. Once that file is on zenodo, we can point to it in the documentation.


fxt44:
does one need OP_mono_master_grid_MESA_emesh.txt to run a meaningful test suite?

evbauer:
Yes. Should be uploaded to zenodo soon.

warrickball:
I think `ifort` is tripping up because the `COMPILE_NO_CHECKS` option means `ifort` assumes short lines based on the `.f` extension so errors because it can't see the ends of some lines. I've just checked that renaming the file to `.f90` and removing its special treatment in the makefile got me a successful build with `ifort`. I can push that change if no-one objects. It seems like a simpler solution than adding flags to the makefile headers or tinkering with the offending long lines in `op_eval_mombarg.f`. I assume these choices are just a hangover of copying from `op_eval.f`.

evbauer:
It looks like we're still missing some stuff to call the opacity part of the new OP Mono routines, but I think we at least have the radiative acceleration part working.

evbauer:
I think we're ready for merge now, pending a passing test suite. It would be good to get a few machines on that, especially a mac. @fxt44 could you test `1.4M_ms_op_mono` to make sure is runs on one of your macs?

fxt44:
2c174fc 1.4M_ms_op_mono passes on my m1 laptop:

warrickball:
`ifort` testing isn't usually a merge blocker but be aware that `bluebear_ifort` is still offline. I'm hoping it comes back up this afternoon for `ifort` testing and that the backlog clears quickly.

warrickball:
[`radiative_levitation`](https://testhub.mesastar.org/Mombarg_OPmono/commits/2c174fc/test_cases/star/radiative_levitation) failed on `bluebear_ifort`. There weren't any compiler warnings from the new `op_eval_mombarg.f90` file. I'm trying a local install with SDK to see if I've set things up wrong.

evbauer:
I had turned it on by default earlier in the branch to stress test, but the `'mombarg'` routines are really meant for the main sequence, so I switched it to default to the old version of the routines (`'hu'`), with the option to turn on `'mombarg'` available who know when it's the right application.

Issue #425, opened 2022-07-05
CPM is broken since 15140


rjfarmer:
Appears to be related to the removal of lnpgas

evbauer:
From Anne on slack:
[convcorecpm.pdf](https://github.com/MESAHub/mesa/files/9047868/convcorecpm.pdf)

rjfarmer:
Do we have a test case that specify tests that the core grows with ledoux? That would a be helpful first step to make sure we keep this working (when we have a fix)

evbauer:
Some notes that hopefully help clarify what's going on.

- Setting `conv_premix_fix_pgas = .false.` in 12778 breaks convective premixing. This means that the EOS updates internal to CPM need to be done at fixed PT rather than fixed DT.
- What breaks here is `gradr` getting set improperly. Doing an EOS call at fixed DT changes the pressure of the newly mixed cell after attempting to move the CPM convective boundary. Then when `gradr` is recalculated it uses that wrong pressure, and can end up below `grada` and therefore non-convective. But then the subsequent Newton iterations adjust the pressure profile back into the right configuration, and at the end of the step you end up seeing `gradr > grada` at the convective boundary.
- We don’t need to reintroduce `lnPgas_flag` to fix this, but we do need to reintroduce EOS calls at fixed PT for the cells touched by CPM. We need to change the logic that's commented out around [here](https://github.com/MESAHub/mesa/blob/main/star/private/conv_premix.f90#L1369) so that it no longer relies on `lnPgas_flag` for doing EOS calls at fixed PT for the relevant cells.

BucheleL:
@annethoul Here's a rough draft of a test suite case to check this for the future. It uses the starting model from the MESA summer school lab and works similarly to the existing hb_2M test suite case. The insist specifies the range of convective core mass that is acceptable when the central H abundance is 0.5. The extras_after_evolve routine checks whether the model ends up in this range. The values currently given in the inlist were pulled from the plot shared by Evan in this thread. 
[CPM_LEDOUX_test.tar.gz](https://github.com/MESAHub/mesa/files/9394879/CPM_LEDOUX_test.tar.gz)

PR #429, opened 2022-07-19
CPM Fix

evbauer:
Potentially addresses #425.

### DIFF
--- a/star/private/conv_premix.f90
+++ b/star/private/conv_premix.f90
@@ -1344,7 +1344,6 @@ contains
 
     logical, parameter :: TRACE_UPDATE_MODEL = .FALSE.
 
-    logical  :: lnPgas_flag
     integer  :: ierr
     integer  :: kf_t
     integer  :: kf_b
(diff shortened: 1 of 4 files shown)


warrickball:
Can we get a test case to make sure this doesn't break in the future? If not in this PR, we should open an issue to increase the chance that it does happen at some point.

annethoul:
Hi Warrick,

Yes, we discussed this in yesterday’s zoom with Bill Rob and Evan.
I will create a test case that tests CPM. 
However, the order in my priorities now is:
1)  finish preparing my summer school labs and lectures
2)  read mesa6 and comment the sections
3) attend the mesa6 writing session
4) attend the Summer School
5) thoroughly test the cpm fix !!
6) Create and test the test case

Therefore, the test case will probably not be ready before the end of August...

Anne

evbauer:
We may need to merge this PR sooner to close the release-blocking aspect, then open another issue reminding us that a test case is still needed.

PR #432, opened 2022-07-28
Starspots

mjoyceGR:
This Pull Request merges the starspots test_suite case into the main code base.


evbauer:
Let's add in documentation for the `README.rst` file to replace the template text, then I think this should be ready to merge.

evbauer (approved):

mjoyceGR:
Squash & merge with @evbauer's approval

star/test_suite/starspots/src/run_star_extras.f90
```diff
@@ -0,0 +1,413 @@
...
+         s% how_many_extra_profile_columns => how_many_extra_profile_columns
+         s% data_for_extra_profile_columns => data_for_extra_profile_columns  
+
+         s% how_many_extra_history_header_items => how_many_extra_history_header_items
+         s% data_for_extra_history_header_items => data_for_extra_history_header_items
+         s% how_many_extra_profile_header_items => how_many_extra_profile_header_items
+         s% data_for_extra_profile_header_items => data_for_extra_profile_header_items
+
+
+         if (s% use_other_surface_PT) then
```
warrickball, line 81: The `use_other_...` controls get handled by the main codebase, right? I think this is unnecessary. If it's been added because MESA _wasn't_ handling the logic, then that's worth investigating in the main code.

star/test_suite/starspots/src/run_star_extras.f90
```diff
@@ -0,0 +1,413 @@
...
+         extras_start_step = 0
+
+         !! to use constants as defined in MESA, import const_def and call
+         !! variables by their names there
+
+         ! set PB_i here and then other_mlt will know about it, because 
+         ! quantities set here are carried over the course of a timestep, 
+         ! AKA all newton iterations
+         
+         sb_sigma = boltz_sigma !5.67d-5 !!cgs units
```
warrickball, line 129: Why not just replace `sb_sigma` with `boltz_sigma` instead of defining another variable? (Same goes for `R_gas_constant = cgas` below.)

warrickball:
I realise I've commented after the merge with some things that can be tweaked. Happy to do these myself later. None actually affect any output so they're not important.

star/test_suite/starspots/src/run_star_extras.f90
evbauer: Yes, good point. Should be safe to delete this extra logic. I'd say feel free to make that change.

mjoyceGR:
fine with me to make those tweaks yourself

PR #436, opened 2022-08-10
added const secday, which is useful for binary stuff

matthiasfabry:
as the title says.

### DIFF
--- a/const/public/const_def.f90
+++ b/const/public/const_def.f90
@@ -96,7 +96,6 @@
       real(dp), parameter :: dayyer = 365.25d0 ! days per (Julian) year
       real(dp), parameter :: secyer = secday*dayyer ! seconds per year
       real(dp), parameter :: secday = 24*60*60  ! seconds in a day
-      real(dp), parameter :: secyer = secday*dayyer ! seconds per year
       real(dp), parameter :: ly = clight*secyer ! light year (cm)
 
 ! inexact but very well measured physical constants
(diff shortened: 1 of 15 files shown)


const/public/const_def.f90
```diff
@@ -94,7 +94,9 @@ module const_def
       real(dp), parameter :: au = 1.49597870700D13 ! (cm) - exact value defined by IAU 2009, 2012
       real(dp), parameter :: pc = (3.600D3 * rad2a) * au ! (cm) parsec, by definition
       real(dp), parameter :: dayyer = 365.25d0 ! days per (Julian) year
-      real(dp), parameter :: secyer = 24*60*60*dayyer ! seconds per year
+      real(dp), parameter :: secyer = secday*dayyer ! seconds per year
```
warrickball, line 97: Duplicate of line 99, no?

warrickball:
On balance, I think this is a sensible change. I expect most people recognise factors like `24*60*60` and `86400` as the number of seconds in a day but using `secday` is just that little bit clearer and slightly less error prone (e.g. `.../24*60*60` instead of `.../(24*60*60)`). I'd suggest we wait until early next week to give other devs a chance to respond.

If this is worth changing, you might want to make some substitutions now. Grepping for `"\*60\*"` or `"\*60d0\*"` in the source (including test cases) turns up some easy replacements.

orlox:
A bit of a tangential thing, but how would people feel about changing `secyer` to `secyear`? Guess it would break a bit of external code, but in a pretty trivial way that compilers should be able to suggest how to fix.

fxt44:
or even the more obvious second_per_year. neither here nor there, but the secyer naming is from ~2005.

evbauer:
Personally, I feel like a "fix" like `secyear` will just get us into [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding)  territory, so I'd prefer just to leave it.

(But I fully approve of `secday`.)

rjfarmer (approved):

PR #438, opened 2022-08-12
Fix for conductive opacity edge cases

evbauer:
Fixes a bug discovered by the students at MESA Summer School 2022.

331b515fe3e7d34be1453c81fac8b2f6ad07e95c introduced a bug by changing where the conversion from thermal conductivity to conductive opacity occurs. This change missed some `zlog` edge cases because they included `return` statements that caused the conversion at the end to be skipped. This manifested as `NaN` values for the `kap` derivatives because they never got set, and the returned opacity was also wrong (though not `NaN`). These `NaN`s in the derivatives propagated through to the overall opacity derivatives even when conductive opacities should be irrelevant.

The most relevant edge case where this occurs is for a pure hydrogen composition, though in principle it could also occur for a very high `zbar` (>60).

### DIFF
--- a/kap/private/condint.f90
+++ b/kap/private/condint.f90
@@ -194,57 +194,51 @@
          
          if (zlog <= logzs(1)) then ! use 1st
             call get1(1, logK, dlogK_dlogRho, dlogK_dlogT, ierr)
-            kap = exp10(logK)
-            return
-         end if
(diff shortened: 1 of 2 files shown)


fxt44:
changes look good to my eyes.

PR #439, opened 2022-08-12
Change in check_jdot_mb_conditions and binary_controls

jaiken17:
qrad_core and jdot_mb_max_qrad_core are misleading variable names since check_jdot_mb_conditions (binary/private/binary_jdot.f90) is actually checking that the convective core is not too large (ie that there actually is a radiative core). The documentation is also updated to reflect what the check_jdot_mb_conditions subroutine is actually doing.

### DIFF
--- a/binary/defaults/binary_controls.defaults
+++ b/binary/defaults/binary_controls.defaults
@@ -957,7 +957,7 @@
 
     jdot_mb_min_qconv_env = 1d-6
     jdot_mb_max_qconv_env = 0.99d0
-    jdot_mb_max_qrad_core = 1d-2
+    jdot_mb_max_qconv_core = 1d-2
     jdot_mb_qlim_for_check_rad_core = 1d-3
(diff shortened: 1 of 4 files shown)


evbauer:
Ok, looks like we have a compilation failure. Feel free to push a commit that fixes it and we can run the compilation test again.

jaiken17:
I will try to fix the compilation error later today.

jaiken17:
I have now fixed the compilation error, and can confirm it compiles on my machine and runs one of the binary tests. I have not run the full binary test suite since I am not set up to do so at the moment.

evbauer:
Great! Looks good to me. We'll just wait for whenever Pablo (@orlox) has a chance to look at it, and we can merge it once he approves.

rjfarmer:
@orlox any chance you could have a look at this? It seems @evbauer happy so can we just merge it?

Issue #440, opened 2022-08-12
rate error in MESA


200k33p3r:
During the MESA summer school 2022, I tried to run minilab1 task3 from Rob's lab using 'approx21_cr63.net' as the network and the code crashed which appears to be an issue with a bad reaction somewhere in MESA. See file attached.

[cr63_out.log](https://github.com/MESAHub/mesa/files/9328773/cr63_out.log)

maygpedersen:
This error occurred at my table during the summer school and I was able to reproduce it on my own laptop

PR #444, opened 2022-08-15
Start on fixing #440

rjfarmer:
Its seems with approx21_cr63.net we make a lot more free neutrons than
normal. So EOS and screening need to handle having zbar<1

Still have issues with e<0 in pack_for_export. @adamjermyn thoughts on what's best to do? Error out or add some checks earlier to skip things, if mostly made of neutrons.


rjfarmer:
Seems to have fixed approx21_cr63 with the chem_id fix. So no need to worry about negative e's now.

adamjermyn (approved):
Looks good to me. Thanks for handling this!

Issue #163, opened 2021-01-24
adipls should get a USE_ADIPLS build option


rjfarmer:
Given that building adipls breaks the moment you look at it (well change compiler flags or set things like FFLAGS in your shell), can we get a way to disable building adipls (like we can do with gyre)?  

That way i can completely ignore adipls.

PR #445, opened 2022-08-15
Add build option to disable building adipls

rjfarmer:
Works on #163

### DIFF
--- a/adipls/build_and_test
+++ b/adipls/build_and_test
@@ -2,7 +2,7 @@
 
 function error {
 	echo
-	more diff.txt
+	cat diff.txt
 	echo
 	pwd
 	echo "TEST FAILED -- compare obs.txt to obs.test.txt"
@@ -34,7 +34,7 @@ fi
 
 if [ ! -f ../skip_test ]
 then
-	if [ ! -f skip_test ]
+	if [ ! -f skip_test ] && [ ! -f test/build_skip_test ]
 	then
   		cd test
   		./ck >& diff.txt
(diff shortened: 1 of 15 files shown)


warrickball:
First, thanks for picking this up!

The `astero` tests won't build with `USE_ADIPLS=NO` because there's no `libadipls.a`. (The build also fails with `USE_SHARED=YES` for the same reason.) I think we want to be able to build and run these tests even with `USE_ADIPLS=NO`, at least to mimic how we handle GYRE. There, if `USE_GYRE=NO`, the tests (e.g. `astero/test_suite/astero_gyre`) still run as much as possible via the stub code and feign success by printing the relevant output lines in `run_star_extras.f90`, even though GYRE isn't installed.

I can think of two ways of handling this.

1. Catch the `USE_ADIPLS` flag in the relevant `astero` makefiles and remove the `-ladipls` link.
2. Set up the `adipls` makefiles to still produce a `libadipls.a` with the stub interfaces.

I prefer 2. because that's what we do for GYRE. 1. might be easier to implement (via `astero/make/makefile_base`) but I'm not sure it'd actually work because there'd still be lines of code to call ADIPLS procedures, they just wouldn't be executed.

I have four substantial deadlines in the next three weeks, so don't foresee myself contributing anything until at least then.

rjfarmer:
I'd rather do 1.

The problem with doing a stub file in adipls is its not clear what would need to be wrapped. As there is no public/private split in adipls or a adipls_lib.f90 file (like gyre has). So i suggest we just filter out -ladipls in astero build. Then fake the tests results from there. I did add stub files in astero/private for adipls alreadt so we just need to tweak the test cases to check for when ADIPLS_IS_ENABLED is false.

warrickball (approved):

PR #446, opened 2022-08-15
Add ierr to pack_for_export

rjfarmer:
This way we can catch when the solver has gone into the weeds and starts
probing random locations in the EOS parameter space.


adamjermyn (approved):
Looks fine to me. Also looks related to issue #285 (though with less restrictive conditions than we tried to impose there).

PR #450, opened 2022-08-26
Push net/ state into net_info

rjfarmer:
Simplify much of the net memory management by moving net_info components into allocatables instead of pointers, as well
as making net_info not  a pointer either.

I plan to push more of the net state into net_info so then it becomes easier to pass around the net (For instance making functions
take both the temp and n, but n contains the temp anyway so one of these is redundant).

This has also fixed a hard to find bug in burn_const_P which was using (probably) random values for T,Rho as we where
accessing the wrong part of an rpar array (which never errored as we had a large work array stuck to the end of rpar which
masks the wrong access).

Hopefully  by the end of this we can just call net_get(..., rates_only=.true.) and get back the rates for that zone, without having to re-implement half of eval_net

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -96,6 +96,9 @@ net
 The derived type net_info (conventional given the symbol ``n``) is no longer a pointer. If you declare a local copy of the variable, you should also ensure to do ``n% g => g`` to make sure that net_info knows
 about the ``net_general_info`` derived type. ``g`` can be had from a call to ``get_net_ptr(handle, g, ierr)``.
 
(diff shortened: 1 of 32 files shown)


fxt44:
whoa! freaking awesome!

rjfarmer:
Test suite https://testhub.mesastar.org/rf%2Fnet_rewrite/commits/246ac13 has only 2 failures for now, which isn't too bad. Though there are several test cases (like https://testhub.mesastar.org/rf%2Fnet_rewrite/test_cases/star/ppisn) that finish in fewer steps and less retires. Which is surprising, I was kinda hoping this would go for no change so we could be more confident i got things right.

So we should be wary of just accepting a passing test suite as it seems some models are changing.

fxt44:
running 246ac13b now. happy to dig into some of the test suite cases that may be suspect.

Issue #55, opened 2020-05-31
Add checks for uninitialized integers?


rjfarmer:
Something I found while fixing thinags in 13837 is that we don't currently have anyway to set integer arrays (and I presume scaler ints) to an invalid value to check for uninitialized values. For floats we can set things to a nan. Maybe we should expand the set_nan code to set int arrays to -huge(int) as an invalid value? We won't get the nice crash on use we get with a nan but at least it will be easier to see when we haven't set an array properly.

PR #457, opened 2022-09-30
fix to after_step_loop so incoming result value is considered

matthiasfabry:
When using r15140 with linux SDK 20.12.1 to run binary models, I unfortunately noticed no history or profiles of the stars were saved. I tracked the issue down to `run_star_support: after_step_loop`, where `star_finish_step` was not called since `result = -470533012` (or any other number that happened to be in that memory location) at that point. 
Newer compilers (such as the macOS SDK 22.6.2) seem to implicitly put `result = 0 (== keep_going)`, which avoids this issue, but still, if `result == terminate` upon invocation of `after_step_loop`, this value is not considered because of the `intent(out)`
simply changing the intent to `inout` avoids problems on any compiler, and lets `after_step_loop` take in the `result` value you give it.

Weird that this issue did not turn up earlier, SDK 20.12.1 was supposed to be a compatible SDK for r15140. Are there tests that logs/history and/or profiles are actually created/written?

### DIFF
--- a/star/job/run_star_support.f90
+++ b/star/job/run_star_support.f90
@@ -657,13 +657,13 @@
       end subroutine before_step_loop
 
 
-      subroutine after_step_loop(id, inlist_fname, &
-             dbg, result, ierr)
+      subroutine after_step_loop(id, inlist_fname, dbg, result, ierr)
          integer, intent(in) :: id
          type (star_info), pointer :: s         
          character (len=*) :: inlist_fname
          logical, intent(in) :: dbg
-         integer, intent(out) :: result, ierr
+         integer, intent(out) :: ierr
+         integer, intent(inout) :: result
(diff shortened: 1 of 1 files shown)


rjfarmer:
Ouch, nice catch

Test cases do write out history and profile data but nothing checks it actually gotten written, we depend on gfortran throwing an error if it can't write out a file. 

We also do extensive testing looking for uninitialized variables, but binary has many issues that need fixing for this (and they get caught in the order they appear during a run so something in the shutdown code is likely the last bug(s) we will find). Also its mostly geared at finding issues with float's (which we can set to NaN and catch) while integers are harder to set to a "poison" value (See #55 ). But this is something we will need to have another look at.

rjfarmer (approved):

Issue #456, opened 2022-09-29
metallicity dependence of stellar winds


mathren:
Hi all! 
I was looking at the wind mass-loss implementation in MESA, and I think there is a long-standing issue at this line:
https://github.com/MESAHub/mesa/blob/db6331e58f493bd910a032815bc554d12a8d1220/star/private/winds.f90#L341
This explicitly allows the surface metallicity to change as the star evolves (and possibly interacts with companions!). This is fine everywhere in the code because physically the surface metallicity can evolve in time, of course.

However, I think it is an issue in the wind mass loss routine (at least for line-driven winds of hot massive stars). For line-driven winds, the bulk of the opacity below the sonic point is provided by iron group elements, and the other metals may contribute to the radiative acceleration post-sonic-point (thus only influencing the wind v_infinity but not Mdot). Most stellar wind recipes are designed using Z as a proxy for surface iron abundance. See for instance the discussion in sec. 3 here https://arxiv.org/abs/2205.09125 (for O and WNh stars) or here: https://ui.adsabs.harvard.edu/abs/2016ApJ...833..133T/abstract (for WR of various sub-classes).

Except for pollution from a nearby SN explosion, which is usually a small effect, the iron-group element abundances are not modified throughout the star's lifetime. So even if H and He drop (maybe because of mass loss), and N increases (say because of rotational mixing or accretion), affecting Z, the driving of the wind should *not* change. Presently MESA is likely doing something incorrect. This is likely not a major problem (most winds scale with Z^x with x~0.5-0.8 and are uncertain between 0.1-1dex anyways), but it is a theoretical inconsistency that can be easily fixed, e.g., by changing the line above to use `Z=Z_base` in the wind routines. `Z_base` is a constant throughout the evolution (like the surface Fe-group abundances), related to opacity (i.e. what matters for the wind driving). Even if this means that X+Y+Z_base is not one in the wind routines, that should not be a problem. Using Z_base instead of Z would be a simple fix that makes things nuclear-reaction-network independent.

What do the developers think about this? Would such a "fix" be welcome?

rjfarmer:
Yes, a fix would be helpful. 

Though we need a fallback for when zbase isn't set (when not using type2 opacties). Initial_z would work but that then needs a fallback if that's not set (when loading a model you don't need to set initial_z). Maybe that then just falls back to the surface Z.

mathren:
Is there any variable that stores (or could be misused to store) the initial surface Z?
[...]

PR #458, opened 2022-09-30
use Z = Zbase in wind mass loss routines

mathren:
fix for issue #456

Wind mass loss rate is determined by iron opacity, unlikely to change during the evolution. The surface Z can change (because of changes in H, He, C, N, O) but this usually does not change the iron group elements, and should not affect the line-driven wind mass loss rate (see e.g., [Tramper et al.
2016](https://ui.adsabs.harvard.edu/abs/2016ApJ...833..133T/abstract), [Sabhahit et al. 22](https://arxiv.org/abs/2205.09125) and references therein).

If Zbase is set (for type 2 opacities), use that fixed constant value in the wind routine. If Zbase is not set, fall back on the previous behavior.

### DIFF
--- a/star/private/winds.f90
+++ b/star/private/winds.f90
@@ -54,9 +54,9 @@
                if (s% report_ierr) write(*, *) 'other_adjust_mdot'
                return
             end if
-         end if         
+         end if
       end subroutine set_mdot
-      
+
 
       subroutine do_set_mdot(s, L_phot, M_phot, T_phot, ierr)
          use chem_def
@@ -197,7 +197,7 @@
          end if
 
          if (dbg) write(*,1) 'wind_mdot 1', wind_mdot
-         
+
          if (using_wind_scheme_mdot) then
             if (s% no_wind_if_no_rotation .and. .not. s% rotation_flag) then
                s% mstar_dot = 0
@@ -338,7 +338,14 @@
                 end if
                 X = surface_h1
                 Y = surface_he4
-                Z = 1 - (X + Y)
+                ! use the Zbase as a proxy for Fe-group abundances
+                ! unlikely to change during the stellar evolution
+                if (s% kap_rq% Zbase /= -1) then
+                   Z = s% kap_rq% Zbase
+                else
+                   if (dbg) write(*,*) 'Zbase not set, using present-day Z for wind'
+                   Z = 1 - (X + Y)
+                end if
 
                 if (use_other) then
                    if (dbg) write(*,*) 'call other_wind'
@@ -652,7 +659,7 @@
          scale_height = s% rlo_wind_scale_height
          if (scale_height <= 0) then
             scale_height = s% Peos(1) / (s% cgrav(1)*s% m(1)*s% rho(1) / (s% r(1)**2)) / Rsun
-         end if 
+         end if
          roche_lobe_radius = s% rlo_wind_roche_lobe_radius
          ratio = R/roche_lobe_radius
          !write(*,2) 'R/roche_lobe_radius', s% model_number, ratio
@@ -671,7 +678,7 @@
          mdot = s% rlo_wind_base_mdot* &
             exp(min(6*ln10,(R - roche_lobe_radius)/scale_height))
          eval_rlo_wind = s% rlo_scaling_factor*mdot ! Msun/year
-         
+
          !write(*,1) 's% rlo_wind_base_mdot', s% rlo_wind_base_mdot
          !write(*,1) 'R - roche_lobe_radius', R - roche_lobe_radius, R, roche_lobe_radius
          !write(*,1) 'scale_height', scale_height


Debraheem:
@mathren, should we add a flag to check if 'use_Zbase_for_Type1 = .true.'?

 I'm thinking of a situation where you have a high-Z accretion which enhances the surface metallicity such that it doesn't correspond to the initial metallicity(i.e. Zsurf >> Zbase = Zinitial) . In this scenario the surface Fe abundance would be enhanced.
 
You'd be using type2 inside the star with zbase, type1 at the surface with Z>Zbase and the zbase inside the star would not necessarily correspond to Z value of the surface? 

Without this check, I could imagine a high-Z accretion model would under-predict the wind? thoughts?

mathren:
I'm struggling to see a physical situation when a star would accrete Fe-rich material (which is what matters for the wind, even 100÷ metals but only being CNO should not affect the wind mass loss rates directly, at least not based on the formulae we typically use).

The one case that could occur is accretion of some ejecta of the companion's SN explosion (and there are claims of detecting alpha enhancement because of that in the literature), or making binaries with different metallicity of the two stars somehow (which is hard enough to imagine that if someone wants to model this they should probably take care of this with customized routines).

Can you elaborate on what scenario you have in mind for accretion of Fe rich material?

Debraheem:
> Can you elaborate on what scenario you have in mind for accretion of Fe rich material?

In hopes of finding where our Zbase assumption breaks down, I was just imagining an arbitrary scenario where you accrete material that is higher in metal content than the reference Zbase of your model. I'm happy with your answer.

If we are primarily concerned with Fe and other alpha elements, wouldn't heavy element diffusion on the main sequence also modify the surface metallicity giving a Z value lower than Zbase? Is this a valid concern?

mathren:
Maybe, for sufficiently long lifetimes. But then there is also radiative levitation to partially compensate. And the timescales for these phenomena are (very) long compared to the lifetime of stars for which mass loss using those formulae apply...Those empirical and/or theoretical formulae are based on observations/calculations with a certain baseline L, one should not use Vink+01 for the Sun for example.

It's of course not impossible to invent an academic exercise where this would matter, but I can't think from the top of my head of something that would not be sufficiently contrived that any interested user should not expect the code to work out-of-the-box for their purpose.

One could in principle not use Zbase but the actual mass fraction of Fe-group elements (that would probably be the most correct thing to do, and let it evolve as the model prescribes), but then those are not included in small nuclear reaction networks...So I think for 99.9% of applications using Zbase is an adequate fix, and the remaining cases seem to me special enough that interested users should work them out for themselves (e.g., using `other_wind_hook`)

fxt44:
> arbitrary scenario where accrete material that is higher in metal content than the reference Zbase of your model.

examples in the literature include (a) iron enrichment of very metal poor stars due to accretion from the ism during galaxy assembly (b) accretion of iron rich planetesimals at zams (e.g., exoplanet hosts) and on white dwarfs.

mathren:
@fxt44 very good points! But wouldn't those be for stars with low mass, low L,  and thus no **line-driven** winds? The metallicity scaling in these mass loss recipes are really based on line driving (when they are not empirical rates, then the argument is the range of L observed, and it's a bit debated right now), so not really applicable for accretion of planetesimals nor stars long-lived enough to accrete significant amount of ism? (Or there is no reason to think so from a theoretical perspective)?

Debraheem:
Thank you both, I appreciate this discussion. For those including Fe in their network, I'm considering adding an Fe abundance scaling as opposed to Zbase as an optional flag for more self consistent wind rates. This would cover approx21 and most general networks and would alleviate the worries over diffusion/levitation etc.

fxt44:
i'm just giving examples where accretion iron is different that zbase iron, not adjudicating them with further constraints ;)

mathren:
@Debraheem this would good to me, with the caveat that from the line-driving perspective it's a list of Fe and Fe-group isotopes that matter. Should it have a sum over all the isotopes with 45<~A<~62  available (or whatever we define as "iron group", possibly to be fetched in the references above to match what can be seen/what is included in the wind mass loss calculations)? 

I'm imagining the situation where someone uses a network including fe56 (e.g., approx21) and then switches to a large network that has more isotopes of fe, but not all are counted in the wind routine and inexplicably their wind mass loss rate goes down because of this. It would be pretty hard to understand without being on top of the radiatively driven mass loss literature...
I don't think we want to add a hidden dependence on the mass loss rate on the nuclear reaction network.

This was my original reason to stick with Zbase and not specify a list of isotopes.

But having a cumbersome `if network is this then ... etc.` is in my opinion over complicated: those wind mass loss recipes also have some estimate of errors on the coefficients which we blissfully ignore in evolutionary calculations and they may have a much larger impact than what we are discussing here. No way to get it completely "right"...and the complication added by trying to be precise on the Fe abundance for the corner cases seems not justified.

PR #459, opened 2022-10-10
pgbinary

matthiasfabry:
This adds the minimal infrastructure to `mesa` for `pgbinary`, which mimics functionality to `star/pgstar`.
Main feature is that `pgstar` plots (including `pgstar_grid`s) can be plotted inside `pgbinary`.

To achieve this, changes to the code are:
- changed workings of `binary_history` to more closely mimic `star/history`
- changed `binary_photos` so restarts work with `pgbinary`
- made `pgstar_grid` subplottable
- All `star/pgstar*` modules are copied to `include/` with `./export`, so `pgbinary` can import them

Other changes are mostly large diffs that are just spacing related, introduced by the automatic formatter in my IDE.

Below I showcase the envisioned use of `pgbinary`:
I plot a `pgbinary_grid` window with 4 panels. 
Two of them contain `pgstar_grid`s, which show the two stars' information as if `pgstar` was used. Note that `pgstar_flag == .false.`! This is intended. If `pgstar_flag` were true, in this example nothing special would happen as all `pgstar`'s  `_win_flag`s or `_file_flag`s are `.false.`. One can of course create any `pgstar` window or file on top of `pgbinary` ones by using the appropriate flags.
The two other panels show some `binary_history` info, and a new plot, `pgbinary_orbit', giving a visual impression of the orbit of the two objects (be they point masses or not). This plot may not be super interesting from an astrophysics point of view, but it might from an educational/didactical point of view, it also provides a natural sense of the mass ratio of the system.

I would love to hear your feedback!
Matthias

example png
pgbinary.zip


star/make/makefile_base
```diff
@@ -254,6 +254,7 @@ clean:
 	-@rm -f *.f90 *.o *.mod *.smod *genmod.f90 *.so *.a .depend .order
 
 install:
+	@$(CP_IF_NEWER) pgstar*.mod $(MESA_DIR)/include
```
rjfarmer, line 257: I'm going to try and come up with a better solution for giving binary access to the pgplot code instead of use just copying everything out of star.

binary/test_suite/double_bh/inlist
```diff
@@ -17,9 +17,9 @@
 
 
 
-&binary_pgstar
+&pgbinary
 
-      read_extra_binary_pgstar_inlist1 = .true.
-      extra_binary_pgstar_inlist1_name = 'inlist_project'
+      read_extra_pgbinary_inlist1 = .true.
```
rjfarmer, line 22: Would you mind adding some examples in the a test case (with the default pgbinary_flag=.false.) That way people are more likely to notice there are now plot options.

matthiasfabry: absolutely!

star/make/makefile_base
rjfarmer: Okay i couldn't come up with a better solution without rewriting alot of the pgplot code

warrickball:
As in #460, it looks like there's a huge amount of code change for purely whitespace. e.g. it looks like most of `star/private/pgstar_full.f90` is being rewritten purely because of whitespace changes. Some of this is well justified, e.g. there are some lines with inconsistent indentation being fixed. But almost all of the changes are leading indentation and, even if you disagree with the codebase, I _think_ that, at least for now, the consensus is that the code churn isn't worth it.

Personally I have no problem with the _new_ files being indented differently.

matthiasfabry:
Of course I respect your viewpoint as code-owners, but I believe that reformatting _is_ worth it in keeping the codebase modern. Also, i believe reformatting is not such a big deal as it seems. Take a look at #463 (which I left as draft purposefully). 

That aside, the easiest thing to do for now is you giving a list of files which I should revert, keeping in mind that I would have to basically redo that file's changes from the state pre- this PR.

rjfarmer:
Firstly, thanks for the hard work on this @matthiasfabry .

I've tested and ran the code and things run fine, including restarts. So code wise I'm happy for this to be merged, the only minor things left are to add a note to the changelog and you should add your name to the AUTHORS file ( in the main version there is a section for contributors to put there name). We want to make sure you get appropriate credit for your hard work.

That said I would prefer we don't merge this until after the release, just because its a large number of changes and the RC1 should just be about bugs fixes and documentation updates. Hopefully we will have a full release in a few weeks.

Regarding whitespace: Our current plan is we will merge this and #460 (once @orlox signs off on it) with the whitespace as is (so you don't need to make any changes to your files). Once these have been merged we will finish writing a document on what our preferred whitespace and code format is and then do a mass change ourselves.

The only remaining thing (which is our fault for not having decided this before). Would you mind setting the copyright statement as: 

The MESA team & Matthias Fabry  (& and any other names that where present in the files)

on your new files plus those with you made major changes in (i'll leave it yourself to deciede what a major change is). We are planning to write a more formal documentation on this but you are the first to make a major enough contribution that we have needed to think about this. We will make a pass later to clean up the statements in the other files, so you only need to worry about the ones you've changed.

orlox:
Will go through this next week, an extra nice thing of this functionality is that we can get rid of having as a default `append_to_star_history = .true.`. We have been using that option as a workaround that duplicates all binary_history.data into the star's history.data to be capable to include that information in pgstar plots.

warrickball:
I just noticed that the changes to the indentation of `binary/defaults/binary_job.defaults` [broke the documentation rendering](https://docs.mesastar.org/en/latest/reference/binary_job.html). The ReST interpreter expects the four leading spaces so that the block is interpreted as code. `binary_job.defaults` seems to be the only file that was affected so I'll just re-indent it on `main`.

evbauer:
Nice catch. Since that's just a documentation change, let's go ahead and push that to the `release/r22.11.1` branch as well.

evbauer:
@matthiasfabry just for future reference, there's a fairly detailed guide to documentation formatting at [star/defaults/FORMAT](https://github.com/MESAHub/mesa/blob/main/star/defaults/FORMAT).

These files follow a rather specific set of formatting practices because they serve the dual purpose of setting the default values of our fortran namelists and also getting compiled as ReST documentation of those defaults.

PR #460, opened 2022-10-10
Fix detachment in implicit

matthiasfabry:
allows `b% other_implicit_function_to_solve` to signal that the binary is detached, so `b% check_implicit_rlo` can take action accordingly


warrickball:
I don't want to pick on this PR as "wrong" (it's not!) but there's a general cosmetic point that I don't think we've discussed and possibly should. 

The only actual change in [`binary/other/mod_other_implicit_rlo.f90`](https://github.com/MESAHub/mesa/pull/460/files#diff-b7f2fe908eb92528901636e653b6c55369311c6eb76576b13f3765db7d5477d9) seems to be the addition of the `detached` logical argument. Pretty much the whole file is marked as changed, though, I think only because of the change to indentation. I don't think we have strict style here but it's hard to see what's actually changed when so much is just indents.

rjfarmer:
@matthiasfabry can you see if your editor can be configured to use https://editorconfig.org/ then you can point it at https://github.com/MESAHub/mesa/blob/main/.editorconfig which sets our indentation style. Or just make sure its set to 3 spaces per indentation.

warrickball:
> @matthiasfabry can you see if your editor can be configured to use https://editorconfig.org/ then you can point it at https://github.com/MESAHub/mesa/blob/main/.editorconfig which sets our indentation style. Or just make sure its set to 3 spaces per indentation.

I suspect a lot of *existing* code doesn't adhere to this style, either...

matthiasfabry:
the diff view allows to hide whitespace changes, which clears up a lot! But yes, spacing should be addressed I believe. Eg. why do all module declarations start with 6 spaces? With the fortran standard of 80 characters per line, that's nearly 10% of your space gone. Nearly all of the spacing changes in files is my formatter removing those leading 6 spaces.

@rjfarmer I do use the .editorconfig, however it says nothing on indentation

warrickball:
> the diff view allows to hide whitespace changes, which clears up a lot!

Aha! I wasn't aware of that option. Thanks!

rjfarmer:
>the diff view allows to hide whitespace changes, which clears up a lot! But yes, spacing should be addressed I believe. Eg. why do all module declarations start with 6 spaces? With the fortran standard of 80 characters per line, that's nearly 10% of your space gone. Nearly all of the spacing changes in files is my formatter removing those leading 6 spaces.

Fixed form fortran requires things to start in column 6, though we do use free-form fortran we never really changed that style point.

matthiasfabry:
Of course. For this small a PR, I've reverted some commits.
Although I would like to see the dev team pushing to a more modern free-form Fortran indentation style. I'm sure there must be tools out there to reformat the whole `.f90` codebase this way.

warrickball:
Is it just me or did this [break `jdot_gr_check` and `star_plus_point_mass`](https://testhub.mesastar.org/main/commits/8f2c2d2)?

orlox:
I can reproduce locally, checking now

warrickball:
Our cluster is back up and I first need to work through some compilation failures with `ifort`.

orlox:
Just a clarification, I meant that I locally reproduced the problem so it is real. Looking for a fix now.

binary/private/binary_mdot.f90
```diff
@@ -59,7 +59,6 @@ integer function check_implicit_rlo(binary_id, new_mdot)
          end if
          s => b% s_donor
          use_sum = .false.
-         detachment = .false.
```
warrickball, line 62: @orlox @matthiasfabry (wasn't sure if tagging was necessary)

I'm hunting for a bit-for-bit issue that was probably introduced in this PR. Should this line have been replaced with `detached = .false.`, rather than simply deleted? I'm working my way through the procedure calls to see if `detached` is always set before it's referred to.

warrickball: Great! I'll let you make whatever changes you have lined up then rerun on `ifort` here.

orlox:
fix commited!

Issue #465, opened 2022-11-11
Wrong X_0 in OSC and FGONG files when initial_y < 0


AntoNoll:
Hi everyone,

I noticed that, both in FGONG and OSC files, when `initial_y` < 0, the initial hydrogen abundance is wrong. This is due to the fact that it is defined as `global_data(5) = 1d0 - (s%initial_y + s%initial_z)` in both `star/private/pulse_fgong.f90` and `star/private/pulse_osc.f90`. An easy fix should be to implement the enrichment law in both files (in which case I can do a PR). But, maybe there is a safer way to directly obtain the _actual_ initial helium abundance?

warrickball:
Confirmed in the latest commit (c295eeb). You're right that it'd be safer to consistently set the actual initial Y. At a glance, the same enrichment law Y=0.24+2Z is implemented in at least three places, which might warrant refactoring anyway.
```
$ grep initial_y star/private/*.f90 | grep 24
star/private/create_initial_model.f90:         if (initial_y < 0) initial_y = max(0d0, min(1d0, 0.24d0 + 2d0*initial_z))
star/private/pgstar_production.f90:            if (initial_y < 0) initial_y = max(0d0, min(1d0, 0.24d0 + 2*initial_z))
star/private/pre_ms_model.f90:         if (initial_y < 0) initial_y = max(0d0, min(1d0, 0.24d0 + 2*initial_z))
```
I don't think we have any initial step that checks or resets controls before the evolution begins. Somewhere near the start of `model_builder` in `star/private/init.f90` might be a place to set `s% initial_y` to the value from the enrichment law. We might then have to catch that PGstar call.

warrickball:
I've just created a new branch `fix_pulse_X0` (3b2ab53a) to try fixing this by resetting `s% initial_y` in `star/private/init.f90`.

AntoNoll:
Thanks! It should indeed allow a more reliable access to the initial helium abundance.

PR #466, opened 2022-11-17
Refactor basic enrichment law (Y=0.24+2Z) for initial models with `initial_y = -1`

warrickball:
When an initial model is constructed with `initial_y < 0`, a basic enrichment law `Y=0.24 + 2Z` is invoked:
https://github.com/MESAHub/mesa/blob/68644ccd15db719d33c62ad03754135cbbdd2920/star/defaults/controls.defaults#L59-L73
The enrichment law doesn't carry through to the FGONG or OSC pulsation formats, which @AntoNoll flagged in #465. On investigating, this enrichment law was being invoked in at least three places. This PR refactors all of this by changing the value of the control `s% initial_y` itself to use the enrichment law, so that subsequent routines only see that (positive) value.

`s% initial_y` is only re-assigned just before an initial model is constructed. I don't often start from existing models or photos. I'm not sure if this might cause unexpected behaviour but I assume that if you aren't creating an initial model, you never go down the modified code path. The test suite is [all passing](https://testhub.mesastar.org/fix_pulse_X0/commits/1af7515) with the `[ci optional]` flag.

### DIFF
--- a/atm/private/atm_t_tau_relations.f90
+++ b/atm/private/atm_t_tau_relations.f90
@@ -398,7 +398,7 @@ contains
 
     real(dp) :: x
 
-    ! Evaluate q'(τ) for Ball (2021, RNAAS ...)
+    ! Evaluate q'(τ) for Ball (2021, RNAAS 5, 7)
     ! using fit to solar simulation by
     ! Trampedach et al. (2014, MNRAS 442, 805–820)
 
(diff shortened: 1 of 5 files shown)


rhdtownsend:
Looks good to me.

rhdtownsend (approved):

Issue #368, opened 2022-02-07
Refactor read_extra_..._inlist into arrays


warrickball:
Since time immemorial (before r4366), various MESA inlists have had the option to read other inlists recursively, with options like
```f90
    read_extra_star_job_inlist1 = .false.
    extra_star_job_inlist1_name = 'undefined'
    read_extra_star_job_inlist2 = .false.
    extra_star_job_inlist2_name = 'undefined'
    read_extra_star_job_inlist3 = .false.
    extra_star_job_inlist3_name = 'undefined'
    read_extra_star_job_inlist4 = .false.
    extra_star_job_inlist4_name = 'undefined'
    read_extra_star_job_inlist5 = .false.
    extra_star_job_inlist5_name = 'undefined'
```
The canonical default `inlist` itself uses this to load other inlists.

Should we change these to arrays? I say *yes*. We can also scrap the logical flag and just use the empty string `''` to indicate that an inlist should not be read (unless I'm missing a use case where this helpful). e.g. I imagine a user would write
```f90
    extra_star_job_inlist_name(1) = 'inlist_project'
```
and the defaults file containing
```f90
    extra_star_job_inlist_name(:) = ''
```
to cover all values that aren't used. E.g., in a loop like this should work:
```f90
do i=1, max_extra_inlists
   if (extra_star_job_inlist_name(i) == '') continue
   ... ! otherwise read it
```
If the logical flag is useful, we can instead use
```f90
do i=1, max_extra_inlists
   if (.not. read_extra_star_job_inlist(i)) continue
   ... ! otherwise read it
```
The only downside I can think of is that this will (trivially) break some backwards compatibility but I don't think that's a strong argument.

warrickball:
> let me ask, what is gained by moving to arrays?

We can refactor processing the inlists into loops, which at a rough count would elimiate ~300 lines of code. It also means that future changes only have to be applied once, not 5 times. Plus (though I don't think anyone has ever asked for more) it becomes trivial to increase the limit past 5.

> i suspect most of the test suite will need adjustments to make this work ...

I think a few tactical `sed`s will take care of almost all of these.

fxt44:
fine by me. sharing said tactical sed scripts might be useful to some users.

PR #470, opened 2022-12-20
Inlist arrays

matthiasfabry:
addresses #368 as envisioned.

This cleans up several hundred lines into neat loops and makes this feature more easily extendable to a larger amount of extra inlists.


warrickball:
Nice! As I already said in the issue, I think this is an improvement and, at a glance, it looks like you've caught all the occurrences (including in the docs!). I think three things remain to be done.

1. The `astero` module should be refactored in the same way. This should be straightforward.
2. Can you merge into a target branch `MESAHub:inlist_arrays`? We learned from the last PR that we should do things that way so we can run the full test suite before merging with `MESAHub:main`. We might need to create the `MESAHub:inlist_arrays` branch for you to merge into.
3. This will be disruptive for users, so it'd be nice to distribute a script that will make the appropriate changes. I haven't cooked one up but I imagine there's a single `sed` command that will do it, though we need something that works on both Macs and Linux.

Also, though I don't mind too much, I would've gone a step further and removed the `read_extra_star_job_inlist` logical completely in favour of just ignoring empty filenames but perhaps there's a use case I'm not foreseeing.

evbauer:
I've updated the base branch to `MESAHub:inlist_arrays`. I'll go ahead and merge onto that branch so we can get some testing started, but will also leave the PR open so we can get Warrick's suggestions addressed and you can potentially push more commits as necessary on your end. Nice work Matthias!

evbauer:
You can see testing results here: https://testhub.mesastar.org/inlist_arrays/commits

evbauer:
Oops, merging into `MESAHub:inlist_arrays` closed the PR, obviously. I've opened #471 (and referenced this PR) for when we're ready to merge all the way into main.

I guess this multi-stage PR merging procedure is still something we need to get used to. Is there a more elegant way than what I've just done?

warrickball:
Am I right that you did some sort of command-line merge that unexpectedly closed the PR? I wouldn't have expected the PR to close either but we're learning. In that case, we might have to let the contributor get it to a point where the devs take over then do the merge so that testing can begin. Still not ideal: we somehow want users to be able to respond to TestHub failures.

evbauer:
I actually didn't do anything from the command line. Everything was through the github online interface.

Overall, this seems to have gone pretty well. It's just a bit more convoluted than I had previously appreciated, but Matthias opened a new PR with fixes onto the right testing branch, so everything looks fine to me.

PR #471, opened 2022-12-20
Inlist arrays

evbauer:
This PR just exists to migrate #470 (@matthiasfabry) from a testing branch into main once it is ready. I'll leave it as a draft PR until testing is complete and @warrickball's comments are addressed.

https://github.com/MESAHub/mesa/pull/470#issuecomment-1359639171
> Nice! As I already said in the issue, I think this is an improvement and, at a glance, it looks like you've caught all the occurrences (including in the docs!). I think three things remain to be done.
> 
> 1. The `astero` module should be refactored in the same way. This should be straightforward.
> 2. Can you merge into a target branch `MESAHub:inlist_arrays`? We learned from the last PR that we should do things that way so we can run the full test suite before merging with `MESAHub:main`. We might need to create the `MESAHub:inlist_arrays` branch for you to merge into.
> 3. This will be disruptive for users, so it'd be nice to distribute a script that will make the appropriate changes. I haven't cooked one up but I imagine there's a single `sed` command that will do it, though we need something that works on both Macs and Linux.
> 
> Also, though I don't mind too much, I would've gone a step further and removed the `read_extra_star_job_inlist` logical completely in favour of just ignoring empty filenames but perhaps there's a use case I'm not foreseeing.

### DIFF
--- a/astero/defaults/astero_pgstar.defaults
+++ b/astero/defaults/astero_pgstar.defaults
@@ -83,40 +83,12 @@
    ! Include other inlists
    ! ---------------------
 
-   ! read_extra_astero_pgstar_inlist1
+   ! read_extra_astero_pgstar_inlist
    ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-   ! extra_astero_pgstar_inlist1_name
+   ! extra_astero_pgstar_inlist_name
    ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-   ! read_extra_astero_pgstar_inlist2
-   ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-   ! extra_astero_pgstar_inlist2_name
-   ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-   ! read_extra_astero_pgstar_inlist3
-   ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-   ! extra_astero_pgstar_inlist3_name
-   ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-   ! read_extra_astero_pgstar_inlist4
-   ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-   ! extra_astero_pgstar_inlist4_name
-   ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-   ! read_extra_astero_pgstar_inlist5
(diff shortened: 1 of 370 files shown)


orlox:
One thing I'm curious about this "merge to internal branch->merge to main" workflow, if at the end this is added to main through a "squash and merge", does git keep good track of authorship?

evbauer:
With a squash and merge of this PR, I think the result would show up as something similar to 842538ab01f7db09496b34cc5c166d063bf24623, with it showing both me and @matthiasfabry as authors. I'm not sure if that's entirely fair to Matthias, since he did almost all of the work.

We could also consider doing regular merges, at the expense of no longer having a linear commit history on main. Perhaps it's time for that if we want to emphasize making sure that people get credit for the code they've contributed.

warrickball:
I think @matthiasfabry might not be able to edit this any more because it's in the `MESAHub` space. If so, I'll make the relevant changes to `astero` and try to devise a simple find-replace script.

matthiasfabry:
It seems so. Unless I can push directly onto the MESAHub branch, I'm locked out so to say. I've got `astero` refactored on my fork though.

warrickball:
What happens if you open a PR to merge your changes into `MESAHub:inlist_arrays`? If they show up here after merging your PR into this one we can then test them (though I expect they're working just fine).

warrickball:
Okay, at least that (PR onto a PR) worked the way I expected it to. Once the tests pass, this should be fine to merge after we update the CHANGELOG. I don't mind whether or not we wait until we've devised some scripts for users to adjust to the new set up. They can be documented later.

warrickball:
My first go would be
```
$ sed -E 's/([1-5])([_a-z]*) =/\2\(\1\) =/' 
```
built up using [this online `sed` tool](https://sed.js.org/). I scraped text to test on with
```
$ grep -h "inlist[1-5]" $MESA_DIR/*/defaults/*.defaults | sed '/ *\!/d'
```

evbauer:
This is passing the test suite. It looks like the only outstanding issues are
* uploading a script that will convert inlists to the new format
* adding a changelog entry that links to the script

warrickball:
This is safer because it first selects lines that have `inlist[1-5]` before applying the change. 
```
$ sed -E '/inlist[1-5]/s/([1-5])([_a-z]*) =/\2\(\1\) =/'
```
My first suggestion will incorrectly replace other matching controls (e.g. `gamma1_limit` to `gamma_limit(1)`). I'd appreciate another `sed` guru signing off (@rhdtownsend? @rjfarmer?), after which we just need the changelog entry. We can also tell people to apply the `-i` flag for in-place replacement.

PR #472, opened 2022-12-21
pull into test branch

matthiasfabry:


matthiasfabry:
regarding your comments Warrick:

1. addressed `astero` here
2. I'm not really wizard with shell scripting. I did all of these refactorings with 'find and replace' of my IDE. It shouldn't be too hard though if someone knows the syntax of `sed`. One needs to loop over `i=1..5` and str='star_job', 'controls', ... and move the index `i` aft and enclose in brackets.
3. I personally like sticking with the `logical`s, we do this type of controlling elsewhere too: for example procedure hooks also have the `use_{}` `logical` and then the pointer itself (instead of just checking if the pointer is null or not, although now that I write this, might be impossible to do in fortran). Also it's generally easier to conceptualize if something is `false` that it doesn't do anything, rather than a string being empty.

warrickball (approved):

warrickball:
No problem. I'll merge this and think about some scripts to do the find+replace.

PR #474, opened 2022-12-21
allow access to `do_star_job_controls_before`

matthiasfabry:
allow access to `do_star_job_controls_before` in `run_star_support`, which is needed by `data/star_data/zams_models/create_zams_models`.

I'm not entirely sure why this module turned private since r15140, but `create_zams_models` is pretty useful imo and someone forgot to leave this function public. In general, do we risk anything by leaving stuff public? Maybe name clashes or the like?

### DIFF
--- a/star/job/run_star_support.f90
+++ b/star/job/run_star_support.f90
@@ -54,7 +54,7 @@
       public :: failed
       public :: id_from_read_star_job
       public :: MESA_INLIST_RESOLVED
-      public :: do_star_job_controls_after
+      public :: do_star_job_controls_before, do_star_job_controls_after
       
       ! deprecated, but kept around for use by binary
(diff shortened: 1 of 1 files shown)


warrickball:
Any objections to merging this?

evbauer:
I don't object. FWIW, it looks like this module went private (with lots of exceptions for individual routines) in #186, but all I can see as far as documentation of the change is a log of one of the squashed commit messages:

> improve run_star_support interface for use with twin_studies [ci split]

Anyway, I think Matthias is right that this just needed somebody to remember to make it public again. It looks like rob already did something similar with `do_star_job_controls_after` in 911ac22279e85d618756b3d6357ee4f99a2e7a03.

Issue #433, opened 2022-08-02
orbital velocity in "/binary/private/binary_wind.f90"


xiaoe475:
At line 118 of "/binary/private/binary_wind.f90", only `b%m(s_i)` is taken into account. However, in Hurley et al 2002 eq. 8, the orbital velocity is $v_{\rm orb}^2=\frac{GM_{\rm b}}{a}$, where $M_{\rm b}$ is the **total mass of the binary** (eq. 5 in Hurley et al. 2002), and $a$ is the orbital separation. 

Is there a small mistake?

warrickball:
Just had a quick look and this does look like a bug. Here's the offending line:
https://github.com/MESAHub/mesa/blob/c3ccc90c4361e9cac96ade42b324a1b7c285e745/binary/private/binary_wind.f90#L117-L118
I presume the solution is to change line 118 to something like
```f90
v_orb = sqrt(standard_cgrav * (b% m(1) + b% m(2)) / b% separation)
```

PR #476, opened 2023-01-10
Fix+enhance wind accretion

matthiasfabry:
Made the trivial fix to #433 by changing the orbital velocity.
I also added the functionality to change the wind_beta value since the default 1/8 is only appropriate for cool supergiants.


evbauer:
Nice! I'll once again edit this to merge over onto a branch for testing, then open a new PR for final approval from Pablo before it's merged into main.

PR #477, opened 2023-01-10
Fix wind accretion

evbauer:
From testing branch onto `main` after merging #476 from @matthiasfabry.

Original message:

> Made the trivial fix to https://github.com/MESAHub/mesa/issues/433 by changing the orbital velocity.
I also added the functionality to change the wind_beta value since the default 1/8 is only appropriate for cool supergiants.

### DIFF
--- a/binary/defaults/binary_controls.defaults
+++ b/binary/defaults/binary_controls.defaults
@@ -813,15 +813,19 @@
 ! wind_BH_alpha_*
 ! ~~~~~~~~~~~~~~~
 
-! Bondi-Hoyle accretion parameter for each star. The default is 3/2 taken
-! from Hurley et al. 2002, MNRAS, 329, 897, in agreement with
-! Boffin & Jorissen 1988, A&A, 205, 155
+! Bondi-Hoyle accretion parameter for each star. The default for alpha is 3/2
+! taken from Hurley et al. 2002, MNRAS, 329, 897, in agreement with
+! Boffin & Jorissen 1988, A&A, 205, 155.
+! The default for beta is 1/8=0.125 in accordance for results of cool
+! supergiants from Kucinskas A., 1999, Ap&SS, 262, 127
 ! "\_1" refers to first star, "\_2" to the second one.
 
 ! ::
 
     wind_BH_alpha_1 = 1.5d0
     wind_BH_alpha_2 = 1.5d0
+    wind_BH_beta_1 = 1.25d-1
+    wind_BH_beta_2 = 1.25d-1
 
 
 ! max_wind_transfer_fraction_*
--- a/binary/private/binary_ctrls_io.f90
+++ b/binary/private/binary_ctrls_io.f90
@@ -489,6 +489,8 @@
          b% do_wind_mass_transfer_2 = do_wind_mass_transfer_2
          b% wind_BH_alpha_1 = wind_BH_alpha_1
          b% wind_BH_alpha_2 = wind_BH_alpha_2
+         b% wind_BH_beta_1 = wind_BH_beta_1
+         b% wind_BH_beta_2 = wind_BH_beta_2
          b% max_wind_transfer_fraction_1 = max_wind_transfer_fraction_1
          b% max_wind_transfer_fraction_2 = max_wind_transfer_fraction_2
 
@@ -679,6 +681,8 @@
          do_wind_mass_transfer_2 = b% do_wind_mass_transfer_2
          wind_BH_alpha_1 = b% wind_BH_alpha_1
          wind_BH_alpha_2 = b% wind_BH_alpha_2
+         wind_BH_beta_1 = b% wind_BH_beta_1
+         wind_BH_beta_2 = b% wind_BH_beta_2
          max_wind_transfer_fraction_1 = b% max_wind_transfer_fraction_1
          max_wind_transfer_fraction_2 = b% max_wind_transfer_fraction_2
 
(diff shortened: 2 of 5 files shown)


evbauer:
Assuming tests come back clean [here](https://testhub.mesastar.org/fix_wind_accretion/commits), I'd suggest a rebase and merge once we get approval from @orlox.

warrickball:
7b2a166 supposedly broke `wind_fed_bhhmxb` but I can't reproduce the failure locally with SDK 22.6.1 on Fedora 37. I've submitted the test suite on `bluebear_ifort` so hopefully that'll return results in the next hour. The issue appears almost immediately in the [logs](https://logs.mesastar.org/ebb4dbdcb435cb668e52e7e7101c89d4b3bc206a/cannon_conroy_intel/wind_fed_bhhmxb/out.txt):
```
...
 All this and more can be saved in binary_history.data during the run.
__________________________________________________________________________________________________________________________________________________

       step    lg_Tmax     Teff     lg_LH      lg_Lnuc     Mass       H_rich     H_cntr     N_cntr     Y_surf   eta_cntr   zones  retry
  lg_dt_yrs    lg_Tcntr    lg_R     lg_L3a     lg_Lneu     lg_Mdot    He_core    He_cntr    O_cntr     Z_surf   gam_cntr   iters  
    age_yrs    lg_Dcntr    lg_L     lg_LZ      lg_Lphoto   lg_Dsurf   CO_core    C_cntr     Ne_cntr    Z_cntr   v_div_cs       dt_limit
__________________________________________________________________________________________________________________________________________________

 rl_relative_gap for each component                       NaN                       NaN
 rl for each component                       NaN                       NaN
 r for each component   408924441591.08075        0.0000000000000000     
 m for each component   3.9768117416420936E+034                       NaN
 separation, angular momentum                       NaN                       NaN
 jdot, jdot_mb, jdot_gr, jdot_ml:                       NaN   0.0000000000000000                            NaN                       NaN
 jdot_ls, jdot_missing_wind, extra_jdot:   0.0000000000000000        0.0000000000000000        0.0000000000000000     
 error solving rl_rel_gap
 retry       1
...
```
So somehow `m(2)` is garbage by that point, even though it's fine earlier in the log:
```
...
 initial_bh_spin is larger than one. It has been set to one.

                                                     m2    1.0000000000000000D+01
                                                     m1    2.0000000000000000D+01
                                 initial_period_in_days    1.5000000000000000D+00
                             initial_separation_in_Rsun    1.7134537316182040D+01
                                        jdot_multiplier    1.0000000000000000D+00
                                                     fr    5.0000000000000003D-02
...
```
I can't see how that commit would've affected this. If this is still an issue when I next have a chance to look, I'll try SDK 20.3.2, since that what `cannon_conroy_intel` apparently uses.

evbauer:
Ah, it looks like `wind_BH_beta` 1 and 2 are declared but not assigned in `private/binary_ctrls_io.f90`. I'll go ahead and push what I think should be a simple fix.

evbauer:
Well that fixed the immediate problem, but the test is now failing after running for a while. @matthiasfabry can you run `wind_fed_bhhmxb` including the fix from fd908b5 and take a look at why it's failing now?

matthiasfabry:
The run doesn't fail for me. The checksum is different, but that's to be expected as we changed the accretion efficiency.

warrickball:
What are your system's details? I've just run locally with SDK 22.6.1 on Fedora 37 and got the same output as the two machines on TestHub. Note that the run is completing normally without crashing but fails to produce the required string 
```
Properly tested sub and super Eddington phases
```
The run instead ends with
```
System did not had both sub and super Eddington phases F T
```

matthiasfabry:
Oh, that's cause i've changed the `wind_BH_beta` value to `7` in the inlist. If you change it to the default you get ` Properly tested sub and super Eddington phases` again, although the checksum for the final model still doesn't match...

evbauer:
What are you comparing the checksum for the final model against? That may or may not be an actual problem. E.g. we don't worry about comparing checksums across different commits (the mod files contain a version_number string with the commit hash, so the checksum will change even if the rest of the model contents remain identical). We just make sure multiple machines can agree on the final checksum for a given commit.

It sounds like we might just need one of the owners of this test case (@orlox or @rjfarmer) to decide whether the passing condition should be modified for the new behavior with the updated value of `wind_BH_beta`, or whether we should revert that change so that it keeps passing according to the old standard.

matthiasfabry:
> What are you comparing the checksum for the final model against?

I'm simply doing `./ck` and getting `md5sum: WARNING: 1 computed checksum did NOT match
final.mod: FAILED`

As of yet I'm not familiar with the whole testing infrastructure. I can only confirm this test compiles and runs fine on macos 13.1 with SDK 22.10.1. If you want me to send files, let me know.

evbauer:
Ah, I think that's probably giving you something misleading, depending on where your previous version of the `checks.md5` file came from... We used to use the `./ck` script for something like what you're doing, but it functions somewhat differently now in the era of testhub. There's a decent amount of testing infrastructure documentation [here](https://docs.mesastar.org/en/release-r22.11.1/developing/test_suite.html), but I can't guarantee it's comprehensive. It might be easier just to schedule a quick call sometime if you're interested in having me walk you through what to look at.

Anyway, I think now we just ought to wait to hear back from Pablo and/or Rob for final approval of this test's behavior. No need to send files or anything else for now.

evbauer:
Thanks for the update Pablo! Testing looks good. Should we go ahead and merge?

orlox:
All tests are green after adjusting the inlist for the test case to reproduce the old behavior. This is what I prefer as it verifies that wind mass accretion properly respects the Eddington limit (I did add a small comment to the inlist regarding how the choice might not be the best physics-wise). So I'm ok with merging. Squash & merge?

PR #480, opened 2023-01-19
WD CO Phase Separation Updates

evbauer:
The most significant update here is a small modification to the Skye interface to make the ionization energy offset term optional with an inlist option. It turns out that turning off the ionization term for some of the phase separation energy calls makes the models behave better. The total energetics is unchanged, but the location of heating deposited in the crystallizing WD interior changes.

This also updates the EOS calls in phase separation to be at constant (P,T) rather than constant (rho,T), following the way things are done for CPM. In my testing, this doesn't seem to have a noticeable impact, but I believe it's the more formally correct thing to do at least.

Last note: this PR shows a long commit history of 42 commits, but most of these commits were already merged in the previous PR that introduced phase separation. I guess because that merge was squashed into one commit, git no longer recognizes that those commits are already part of the code. But it's really only the last 7 commits on this branch that introduce new behavior. I'll squash and merge again assuming there are no issues with testing.


fxt44 (commented):
looks fine to my eyes. i have some more test suites running on this branch,

PR #483, opened 2023-01-20
#!/bin/bash shebang in executable

gautam-404:
I am unable to execute `mesa/star/work/clean` using a subprocess in Python. While doing `subprocess.Popen('./clean')`, I get `OSError: [Errno 8] Exec format error: './clean'`. Adding a shebang solved it for me.

Since `clean` is also an executable like the `mk` or `rn` executables, why is there no /bin/bash shebang? (Kindly, pardon my naivety)

I can technically bypass this issue if I run it with `subprocess.Popen('./clean', shell=True)` or if I use `os.system('./clean')`. Both of these run as a subshell and hence are able to execute with no issues. However, subshells come with an overhead so I am trying to avoid these alternatives.

TL;DR I believe that adding a shebang here would allow non-shell processes to execute this file.


gautam-404:
Now that I think of it, adding a shebang would essentially make it run as a subshell. So, I guess it doesn't really matter in my case if the shebang exists or not...

warrickball:
Just for consistency, I think this is a sensible change. As you say, the other scripts (`rn`, `re`, `mk`, etc) all have the shebang, so why doesn't `clean`?

Two minor things I suggest while you're making this change anyway. First, the other scripts leave a blank line after the shebang. e.g `star/work/rn` is
```
#!/bin/bash

rm -f restart_photo

date "+DATE: %Y-%m-%d%nTIME: %H:%M:%S"
./star
date "+DATE: %Y-%m-%d%nTIME: %H:%M:%S"
```
so I'd suggest doing the same. (I think it makes sense to squash and merge this PR anyway.)

Second, none of the `clean` scripts in any of the test suites have the shebang, either, so you can change those too if you feel like it.

gautam-404:
Hi there, I have made the suggested changes. All the clean files should now be consistent with other executables.

gautam-404:
Please do not merge this yet, I'll soon update this same change to the `clean` files in `mesa/binary` and its test suite

warrickball:
> Please do not merge this yet, I'll soon update this same change to the `clean` files in `mesa/binary` and its test suite

If you like, you can[ change the PR to a draft](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft) until you're ready for it to be merged.

gautam-404:
I have made the necessary changes and I think the PR is ready for merge now

warrickball:
This looks great. Thanks! This is conceptually one change in a lot of places so I'll squash and merge. I'll leave it open a little later in case other developers want to comment.

evbauer:
Good idea. We should probably add this change for `mesa/astero` as well, right?

warrickball:
D'oh! You're absolutely right. @gautam-404, do you want to make the same change in the corresponding places in `astero` (e.g. `astero/work/clean`, `astero/test_suite/*/clean`)?

gautam-404:
Hi @warrickball , I have made those changes. Please see if everything is as required.

PR #487, opened 2023-02-05
Add other_close_gaps hook

simonguichandut:
Adding a hook for the "close_gaps" routine in star/private/mix_info.f90.  Following @rjfarmer's[ suggestions](https://lists.mesastar.org/pipermail/mesa-users/2022-November/014151.html) (thanks!)

### DIFF
--- a/star/other/other_close_gaps.f90
+++ b/star/other/other_close_gaps.f90
@@ -1,0 +1,51 @@
+! ***********************************************************************
+!
+!   Copyright (C) 2010  The MESA Team
+!
+!   MESA is free software; you can use it and/or modify
+!   it under the combined terms and restrictions of the MESA MANIFESTO
+!   and the GNU General Library Public License as published
+!   by the Free Software Foundation; either version 2 of the License,
+!   or (at your option) any later version.
+!
+!   You should have received a copy of the MESA MANIFESTO along with
+!   this software; if not, it is available at the mesa website:
+!   http://mesa.sourceforge.net/
+!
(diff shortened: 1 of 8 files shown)


evbauer:
Nice! I've edited the target of this PR to be a branch for our automated testing before a final merge into `main`, which is just our standard practice for external PR's right now. We can merge onto that testing branch once we get approval from @rjfarmer.

star/other/other_close_gaps.f90
```diff
@@ -0,0 +1,51 @@
+! ***********************************************************************
+!
+!   Copyright (C) 2010  The MESA Team
```
rjfarmer, line 3: Should be set to 2023  The MESA Team & Simon Guichandut

star/other/other_close_gaps.f90
```diff
@@ -0,0 +1,51 @@
...
+
+
+      use star_def
+
+      implicit none
+      
+      contains
+      
+      
+      subroutine null_other_close_gaps(id, ierr)
```
rjfarmer, line 40: Should have the same signature as the actual hook.

rjfarmer (commented):
Looks good, will run locally later to check its wokring

Other things to add:

- Should be mentioned in the changelog
- We also need to add a basic version in the star/test_suite/other_physics_hook (that does nothing), so that it gets tested.

rjfarmer:
Issues resolved by committing onto the mesa branch other_close_gaps_hook

PR #494, opened 2023-02-13
Add back some missing rates

rjfarmer:
It's been reported we lost some rates between 10398 and 22.11.1. This PR works on adding back those ones (or fixing whatever bug is causing them to get lost).

Current list:

- [x] r_he4_he4_to_h1_li7    
- [ ] r_be10_wk-minus_b10
- [ ] r_ni66_wk-minus_cu66
- [ ]  r_h3_wk-minus_he3


rjfarmer:
I dont have  a full solution but i have tracked the problem to here:

https://github.com/MESAHub/mesa/blob/c263191db595196fd9b9a295f60b5f908dcac6fa/rates/private/reaclib_input.f90#L335

The weak rates are getting caught in the check for defining the reverse rates based on binding energies and getting skipped. So I think we just want to add a check to skip weak rates at this point.

Issue #442, opened 2022-08-12
control Convective_bdy_weight


annethoul:
Evan asked me to report this as a bug.

The control  convective_bdy_weight was available in version 12115. It is not available in 12778 and later.

Doc:
       !### convective_bdy_weight
        !### convective_bdy_dq_limit
        !### convective_bdy_min_dt_yrs

        ! Mesh function to enhance resolution near convective boundaries
        ! including regions that are newly nonconvective because of moving boundary.
        ! EXPERIMENTAL

       convective_bdy_weight = 0
       convective_bdy_dq_limit = 1d-4
       convective_bdy_min_dt_yrs = 1d-3

When it was available I produced the following plot, which shows that it worked at least in some cases to smooth the convective boundary. I would like to test this control more extensively to understand what it does and where it works. 
So it would be nice to restore those controls.
![cpm_conv_bdy_weight](https://user-images.githubusercontent.com/56915840/184422354-f25b986f-2020-415f-940d-674f9698bc3f.png)

fxt44:
it is available in r12778 as a run_star_extras

% ls /Users/fxt/mesa_work/r12778/star/test_suite/other_physics_hooks/src
% more /Users/fxt/mesa_work/r12778/star/test_suite/other_physics_hooks/src/convective_bdy/convective_bdy.inc

perhaps a similar run_star_extras can be implemented in more recent releases.

fxt44:
and i remembered this exchange:

```
On May 26, 2020, at 10:28 AM, Bill Paxton <paxton@kitp.ucsb.edu> wrote:

please take a look at star/test_suite/other_physics_hooks where this is an implementation of convective_bdy_weight done in run_star_extras.  use that as a guide.

-b


```


```
On May 26, 2020, at 10:24 AM, Joel Ong via Mesa-users <mesa-users@lists.mesastar.org> wrote:

Greetings all,

The controls namelist entries convective_bdy_weight and associated parameters appear to have been removed in r12778 without any documentation in the release notes. Without these options, I'm obtaining discontinuous evolution of some quantities associated with g-mode oscillation frequencies on the red giant branch, where the boundary of the surface convection zone moves inwards over time (see attached plots). Is there any way to restore these options in r12778?

Best regards,
Joel Ong
```

evbauer:
[...]

PR #500, opened 2023-02-27
Restore a simplified version of the convective_bdy_weight control

evbauer:
Working on #442. @efarrell1 and I spent some time digging through the history to figure out what was needed. It seems the reason that this functionality got deleted was because it was closely associated with some of the mixing time-smoothing options and the `smooth_convective_bdy` control, which were also deleted.

The previous version of this control added resolution at both convective boundaries and in newly non-convective regions (to help with receding convective cores). This PR should restore a simplified version that only adds resolution for convective boundaries, but doesn't make any attempt to track the newly non-convective piece.

We need to do some testing to make sure this provides the intended fix for convective core evolution that Anne raised in #442.


fxt44:
nice. full test suite rolling.

efarrell1:
Made some tests with the changes. With convective_bdy_weight = 2, we seem to get a smooth evolution of the convective core boundary. Not much difference between convective_bdy_weight = 2 and = 10. Without convective_bdy_weight, we get a similar problem to the one in Anne's original plot.

conv_boundaries

Issue #288, opened 2021-06-25
Add a global time_delta_coeff control for binary timestep controls


orlox:
For general convergence testing it would be good to implement this functionality of star also in binary. It should also catch up any environment variables defined to increase resolution.

matthiasfabry:
I've implemented the required flag on the linked branch, but I'm unsure on how to systematically test this.

warrickball:
`cannon_conroy_intel` automatically tests all branches, so this is [already being picked up TestHub](https://testhub.mesastar.org/binary_time_delta_coeff/commits/ab6abe9). You've used `[ci skip]` in the commit message, though, which means `cannon_conroy_intel` (and any other system that's set up to recognise the commit message flags) only builds MESA and doesn't run the tests. These flags are listed [here](https://docs.mesastar.org/en/release-r22.11.1/developing/test_suite.html#continuous-integration-testing).

Another good practice is to run locally any relatively quick tests that you think may have been affected. I do this in the `astero` module, to at least make sure that the tests will start. (I loop over the tests and run them with `timeout 30s ./rn` or similar.)

I've just submitted a run to `bluebear_ifort` now, and results for that branch should start coming back in about 30 minutes. `bluebear_ifort` isn't set up to run automatically or detect the commit message flags.

matthiasfabry:
I wasn't referring to how to actually running the test suite, but rather to how we'd design new tests to see if this works as intended. No test in the binary suite is directly impacted (_ie_ there's no new physics or something and it's a new control), but of course all tests are indirectly impacted since they have to pass through `binary_pick_new_timestep`. 

So I guess you're right in that we should test whether all current testcases still give consistent results, but then again we'd need to design something that verifies our new control works as intended.

orlox:
I presume the behavior we'd want to see from the testhub now is that there was no change at all from the old behavior. If we want to have a semi-automated validation of this, we would need a new test case that does a run twice. That could be done in conjunction with `time_delta_coeff` and `mesh_delta_coeff`. By doing a run with defaults, and a run with all these set to 0.5, one could check that to a certain tolerance the number of steps and the resolution by the end of the simulation doubled.

warrickball:
Aha, I apologise for misunderstanding. 

[...]

PR #501, opened 2023-03-01
Binary time delta coeff

matthiasfabry:
Addresses #288. 
Apart from having a maybe a unit test this should be ready to be tested on the suite


matthiasfabry:
Why did my push trigger the `Build MESA` action? I don't think I did anything abnormal

warrickball:
This has just failed to compile on `bluebear_ifort` because the unit test has unformatted write statements. Different compilers have different default output styles, so the diffs fail if the format is unspecified. The solution, which I'll try to fix over the weekend since I have the ifort build, is just to use explicit formats. Some common formats are defined in `include/formats` (which is what some other unit tests use) or, for floats, we tend to use `(1pd26.16)`.

warrickball:
[Fixed](https://testhub.mesastar.org/binary_time_delta_coeff/commits/dd45c1a) in dd45c1a. I don't know why that buggy build action is triggering either...

warrickball:
Are we waiting on anything here? @orlox has approved and it looks like I fixed any ifort-related errors, so it looks like it should be okay to me. The unexpected CI action doesn't look like an issue with the changes in the PR.

evbauer:
Seems to be passing here and on [testhub](https://testhub.mesastar.org/binary_time_delta_coeff/commits/9de0c6d). Should we go ahead with a rebase and merge?

PR #502, opened 2023-03-01
Enable RTI source terms

rjfarmer:
RTI has been broken for ~2 years since 181c9fac as we had a check for rho < 1d99 and if true we would turn off the RTI source term (and thats of course always true).


fxt44:
rolling two full test suites and will look carefully at ccsn_IIp when its done.

rjfarmer:
After chatting with Jared, it seems we are now back to where we were with 15140.

PR #504, opened 2023-03-04
Update GYRE to v7.0

warrickball:
A new version of GYRE ([v7.0](https://github.com/rhdtownsend/gyre/releases/tag/v7.0)) was released Jan 18. This updates MESA by replacing the bundled archive and updating one API change (`md%n_k` → `md%n`) in a few places. I manually cross-checked the results of all the `gyre_in_mesa` tests in `star` as well as `astero_gyre` in `astero` and the only changes were around 10⁻⁷ level changes in some growth rates.

I won't merge this without approval from @rhdtownsend. There's no particular rush but we should merge before the next release.

### DIFF
--- a/gyre/build_and_test
+++ b/gyre/build_and_test
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 if [[ ! -d gyre ]]; then
-    tar xf gyre-6.0.tar.gz
-    mv gyre-6.0 gyre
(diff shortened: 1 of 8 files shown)


evbauer:
@rhdtownsend just flagging that we're targeting doing a release about a month from now. Think we can get a quick approval of this before then?

rhdtownsend (commented):
Nice work.

PR #505, opened 2023-03-06
change order of extras_binary_finish_step and binary_finish_step

matthiasfabry:
Since normal `finish step` saves all vars into `_old`, this way changes made in `extras` are reflected in the `binary_info` correctly. Particularly important if the next step would require a retry.


evbauer:
Testing looks good. @orlox please go ahead and merge this one when you're ready.

Issue #490, opened 2023-02-09
Turn off gold_tolerances2 in defaults?


rjfarmer:
So we keep having a discussion about this and I want to try and see if we can make a decision on this.

We/I have had many reports from users about MESA getting harder to evolve in versions post ~15140. I think some of this can be attributed to the tighter tolerances implied by use_gold_tolerances2=.true. in the defaults. While we want tighter controls in test cases to make sure MESA gets better, its not clear that forcing users to have this on in the defaults is beneficial, especially with people doing stellar engineering to models.

I propose we turn off gold_tolerances2 in the controls defaults and instead just add it to the test suite cases. That way it's clearer to users that this is an option they may want to enable/disable.

fxt44:
i have no objection to this suggestion. is there evidence that turning off use_gold_tolerances2 actually alleviates "harder to evolve"?

rjfarmer:
I generally turn it off when starting a project as it seems to make things easier, but I don't have a formal test of this.

fxt44:
would a (full) test suite run with use_gold_tolerances2=.false. be useful to try?

rjfarmer:
https://testhub.mesastar.org/main/commits/c13e2a4

evbauer:
I would be in favor of this.

I've had intermittent encounters with gold2 tolerances making models harder to evolve. Turning them off usually doesn't solve my problem, but it has helped often enough that it is part of my workflow of things to check now, especially for binary evolution involving degenerate objects.

evbauer:
I also think the gold2 tolerances are largely related to the push for better energy conservation circa MESA V, but in light of the nuances pointed out in MESA VI section 5, I'm not sure it's sensible to push for that level of conservation by default.

Related to this, I would suggest significantly relaxing the default values for `max_abs_rel_run_E_err` and `warn_when_large_rel_run_E_err`.

fxt44:
full test suites rolling on c13e2a4. this will be interesting on several levels!

fxt44:
_Related to this, I would suggest significantly relaxing the default values for max_abs_rel_run_E_err and warn_when_large_rel_run_E_err._

yes. i do this habitually.

evbauer:
https://github.com/MESAHub/mesa/blob/c13e2a49c69be60ed39858fdaa30f35c5637f1b9/star/defaults/controls.defaults#L11430

I would change this default value to `0.1d0`.

[...]

PR #506, opened 2023-03-09
Turn off gold2_tolerances in defaults but leave on in test suite

rjfarmer:
Works on #490 

This should leave test cases that explicitly set gold2 on/off alone and only sets cases which don't specify gold2 to now be on.

https://testhub.mesastar.org/rf%2Fgold2_off/commits/dc8a8c4

### DIFF
--- a/astero/test_suite/astero_adipls/inlist_astero_adipls
+++ b/astero/test_suite/astero_adipls/inlist_astero_adipls
@@ -30,6 +30,7 @@
       x_ctrl(1) = 375d0 ! target frequency
 
       energy_eqn_option = 'dedt'
+      use_gold2_tolerances = .true.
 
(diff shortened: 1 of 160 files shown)


evbauer:
Should I go ahead and push my suggested changes to the default values of `warn_when_large_rel_run_E_err` and `max_abs_rel_run_E_err` to be part of this PR as well?

rjfarmer:
Sure go ahead and push

evbauer:
Ok, pushed! I think I've got cannon back online running these tests too. It seems to be having some filesystem issues (lots of failures on the git lfs checks earlier today), but I was finally able to get past that.

fxt44:
looks good to my eyes. go for the merge.

evbauer (approved):

PR #508, opened 2023-03-13
Add conv_core_cpm Test Case

### LINKED ISSUE #425: CPM is broken since 15140

evbauer:
Add a test case that should catch if a convective core on the MS hasn't grown to the appropriate mass coordinate with CPM on. This will close #425, which is already mostly completed other than this last piece.

Just need to write the README, and I think this should be ready to merge.

Thanks to Lynn Buchele, Morgan Chidester, and Anne Thoul for contributing a lot of this test case setup based on MESA Summer School 2022.


evbauer:
This PR now also contains a couple miscellaneous documentation fixes.
* alphabetizes where test cases show up here: https://docs.mesastar.org/en/latest/test_suite.html#star-tests
* fixes a broken link to one of the `ppisn` inlists in its README file.

PR #509, opened 2023-03-14
Cleanup ccsn_IIp

rjfarmer:
Alot of the code in the run_star_extras isn't even used as it was part of inlist_shock_part6 (which did post shock breakout evolution) which hasn't existed for 2 years. So lets remove it.

Also lets give the many extra controls useful names (which also revealed several controls that overlapped in index).

Before merging remake the standard mods to add back the missing RTI

Photo handling has been updated to the newer testhub way of reading and writing photos.

Tagging Jared  @aurimontem

Still working on adding a check for RTI mixing

### DIFF
--- a/star/test_suite/ccsn_IIp/src/run_star_extras.f90
+++ b/star/test_suite/ccsn_IIp/src/run_star_extras.f90
@@ -30,52 +30,73 @@
       use utils_lib, only: mesa_error, is_bad
             
       implicit none
-      
-!         x_ctrl(2) stop_m
-!         x_ctrl(3-6) magnetar - part 6 only
-!         x_ctrl(7) for setting stop_m in part2, fraction of He layer.
-!         x_ctrl(11-12) Ni mass - part 6 only
-!         x_ctrl(14) turn off RTI - part 6 only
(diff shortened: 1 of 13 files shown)


star/test_suite/ccsn_IIp/src/run_star_extras.f90
fxt44, line 1: very cool. much better than before!

fxt

rjfarmer:
Jared could you have a look at the changes as well as the new rti test? @aurimontem

aurimontem (commented):
This looks a lot cleaner! Great that it preserves the spirit of some of the relic code and much better explains the x_ctrls.  One not-directly-actionable note is that the other_alpha_MLT routine no longer matches the 20M_to_core_collapse test suite, which I think did away with changing alpha in the core and envelope (which was a useful control for RSG stellar engineering). The RTI h1 fraction check looks exactly like what I had in mind; for redundancy's sake it might be worth also having a check for density profile mixing as well, but that's harder to think up in a robust way -- perhaps even checking if alpha_rti is nonzero anywhere in the model. I may have more comments after the weekend when I get a chance to do a bit of stress-testing.

Issue #510, opened 2023-03-16
”log_mtransfer_rate“ in history.data sometimes contain zero values at the onset of the mass transfer process


xiaoe475:
The mass transfer rates ("log_mtransfer_rate") in history.data sometimes contain zero values at the onset of the mass transfer process, which is unphysical for a  low-mass star with stable mass transfer in a binary system. 

It seems that there is a bug when converting mass transfer rate values to its logarithmic value.

![image](https://user-images.githubusercontent.com/70212620/225503327-aea29e33-1582-429f-bf2c-1c965981ebe4.png)

matthiasfabry:
Could you provide the configuration (inlists, mesa version) that reproduces this bug?

xiaoe475:
I used MESA-r22.11.1,  and just made a copy of $MESA_DIR/binary/work to simulate a binary with m1=1Msun, M2=1.4Msun, and Porb=1 day. The inlist files and a plotting python script are in the file "work.zip".

[work.zip](https://github.com/MESAHub/mesa/files/10989787/work.zip)

<img width="1682" alt="截屏2023-03-16 19 01 34" src="https://user-images.githubusercontent.com/70212620/225597635-a821582b-2b1b-4436-8ae2-66538bde203d.png">

This figure shows the "lg_mtransfer_rate" as a function of "model_number". When model_number = 131,  132, 1489 and 1518, "lg_mtransfer_rate" shows zero values (actually, 1e-99).

matthiasfabry:
This happens because of the implementation of `safe_log10` in the `math` module.
in most cases when the mass transfer rate is _very_ low (eg ` b% step_mtransfer_rate == -7e-285), the calculation:
`abs(b% step_mtransfer_rate) / Msun * secyer = 0d0`
end up being 0, and `safe_log10` returns `-99` as expected
However sometimes (eg in the steps mentioned above) `abs(b% step_mtransfer_rate) / Msun * secyer = 1.4e-290`, which is a `IEEE_NON_NORMAL` number and so `safe_log10` returns `1d-99`.

I think we could relax this test of `NORMAL`ness to also allow `DENORMAL`s. 
I think comparing to `1e-99` still works with `DENORMALS`. We would then need to test for `IEEE_IS_FINITE`.

warrickball:
https://github.com/MESAHub/mesa/blob/52fea84ea763f61adc232d2589fbae0b990e1831/math/public/math_lib_crmath.f90#L170-L185

[...]

PR #511, opened 2023-03-21
change safe_logs test from IEEE_NORMAL to IEEE_FINITE

matthiasfabry:
addresses #510.

All tests on [Cannon](https://testhub.mesastar.org/change_safe_logs/commits/3529123) are passing, so I'd propose to implement this change into main. Although we'd probably want to test also on other systems too, since these low level routines could be machine/architecture/compiler dependent.

### DIFF
--- a/math/public/math_lib_crmath.f90
+++ b/math/public/math_lib_crmath.f90
@@ -153,7 +153,7 @@ contains
     real(dp), intent(in) :: x
     real(dp)             :: log_x
 
-    if (.NOT. IEEE_IS_NORMAL(x)) then
+    if (.NOT. IEEE_IS_FINITE(x)) then
 
        log_x = 1.E-99_dp
 
@@ -172,7 +172,7 @@ contains
     real(dp), intent(in) :: x
     real(dp)             :: log10_x
 
-    if (.NOT. IEEE_IS_NORMAL(x)) then
+    if (.NOT. IEEE_IS_FINITE(x)) then
 
        log10_x = 1.E-99_dp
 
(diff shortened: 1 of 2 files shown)


fxt44:
i've started full test suite runs on intel and m1 macs.

math/public/math_lib_crmath.f90
```diff
@@ -172,7 +172,7 @@ elemental function safe_log10_ (x) result (log10_x)
     real(dp), intent(in) :: x
     real(dp)             :: log10_x
 
-    if (.NOT. IEEE_IS_NORMAL(x)) then
+    if (.NOT. IEEE_IS_FINITE(x)) then
 
        log10_x = 1.E-99_dp
```
evbauer, line 177: Do we want to return `1.E-99` here or `-99`?

math/public/math_lib_intrinsic.f90
```diff
@@ -171,7 +171,7 @@ elemental function safe_log_ (x) result (log_x)
     real(dp), intent(in) :: x
     real(dp)             :: log_x
 
-    if (.NOT. IEEE_IS_NORMAL(x)) then
+    if (.NOT. IEEE_IS_FINITE(x)) then
 
        log_x = 1.E-99_dp
```
evbauer, line 176: Do we want to return `1.E-99` here or `-99`?

evbauer:
Or maybe we would want these cases to return some other large negative value (like `-101`) that specifically signals that they come from hitting this condition?

rhdtownsend (approved):

PR #516, opened 2023-03-27
Speed up OP Mono Mombarg Routines

evbauer:
This cleans up some of the behavior at the surface of models when `op_mono_method = 'mombarg'`.

This will fall back to the regular opacity routines for the atmosphere rather than OP Mono in this case, since the approximations from the `'mombarg'` routines seem to cause convergence issues in the atmosphere.

The original version of the `'mombarg'` routines actually did something similar to this by accident due to a bug that yielded `NaN` values in the atmosphere and forced falling back from OP Mono to regular opacities anyway. A fix to conductive opacities in #438 caused this behavior to go away and significantly slowed down the example MESA VI work directory as of release `r22.11.1`.

This PR yields a significant speedup for the MESA VI example work directory relative to `r22.11.1`, and is even somewhat faster than the original version of these routines before the conductive opacity bugfix.

### DIFF
--- a/kap/private/condint.f90
+++ b/kap/private/condint.f90
@@ -25,32 +25,32 @@
 
 
       module condint
-
+      
       use const_def, only: dp
       use math_lib
       use utils_lib, only: mesa_error
-
+      
       implicit none
 
 
       integer, parameter :: num_logTs=29, num_logRhos=71, num_logzs=15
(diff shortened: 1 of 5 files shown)


fxt44:
these changes look fine to my eyes. will run test suite when i return from travel.

Issue #520, opened 2023-04-02
Missing documentation in binary_controls.defaults : sync_mode = "tdyn_div_tkh"


Debraheem:
Following a recent MESA-user mailing list post, I looked into the binary_controls.defaults to find that there no documentation to toggle the 'sync_mode' &control 'tdyn_div_tkh', which exists in $MESA_DIR/binary/private/binary_tides.f90: subroutine sync_spin_to_orbit. The documentation for this control might have been accidentally overlooked?

 Should something like the following be added to binary_controls.defaults?

"! + "tdyn_div_tkh" : Each layer is synced independently using tdyn_div_tkh to iteratively calculate the efficiency, given the timescale."

matthiasfabry:
Good catch!

addressing in #521.

PR #521, opened 2023-04-03
add documentation on the `tdyn_div_tkh` option

matthiasfabry:
addresses #520.

I'm not exactly sure about the implementation, that's why I'm asking for a quick review

### DIFF
--- a/binary/defaults/binary_controls.defaults
+++ b/binary/defaults/binary_controls.defaults
@@ -1045,7 +1045,12 @@
 ! "\_1" refers to first star, "\_2" to the second one.
 ! Options are:
 
-! + "Uniform" : Each layer is synced independently given the sync timescale.
+! + "Uniform" : Each layer is synced independently but uniformly,
+                given the sync timescale.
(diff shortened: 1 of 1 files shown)


fxt44:
better than before! 

presumably this gets automagically reflected in the html docs, https://docs.mesastar.org/en/release-r22.11.1/reference/binary_controls.html .

evbauer:
Yep! Quick reminder that the documentation lines in these files need a comment character (`!`) at the beginning of each line, and you can preview what the rendered docs pages will look like by navigating to `$MESA_DIR/docs` and compiling with `make html`.

orlox:
FYI, there is actually an undocumented reason why I chose not to document this. The code that does this initially originated in the Bonn stellar evolution code, there is some description in a paper that I need to dig up (dont remember the exact reference). But the actual implementation was missing details from that description, and overall I did not feel ok directly promoting the existence of this option as a control. So it ended up with the option existing in code but not in docs. Minimally if we document this, I need to dig up that reference and put a caveat that the actual implementation includes additional details to be understood directly from the code.

PR #524, opened 2023-04-18
Add public interface for k_div_T

rjfarmer:
I've had a request to make the binary_tides.f90 function k_div_T made publically available. Comments on the API?

### DIFF
--- a/binary/public/binary_lib.f90
+++ b/binary/public/binary_lib.f90
@@ -229,5 +229,30 @@
 
       end subroutine binary_set_star_job_namelist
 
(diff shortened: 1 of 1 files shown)


fxt44:
`real(dp) function binary_compute_k_div_T(binary_id, is_donor, has_convective_envelope, ierr)`

would returning "k" or T" be useful auxiliary info?

rjfarmer:
> `real(dp) function binary_compute_k_div_T(binary_id, is_donor, has_convective_envelope, ierr)`
> 
> would returning "k" or T" be useful auxiliary info?

I don't think so. It's the kind of quantity that's  just used as  (k/T) and not k or T

Issue #526, opened 2023-04-28
Central Carbon burning almost finished before He depletion in massive stars


Debraheem:
Hello,

I've been testing some models in r22.11.1 and I've noticed an "issue", which I can't identify as a runtime issue or bug?

The problem:

When running massive stars of any mass in r22.11.1, I've noticed that central carbon burning (c12ag) begins before He is depleted and depletes the core of Carbon before He burning is complete. This is a departure in burning behavior from what I've encountered in previous mesa versions as well as much of what I've read in the literature. Typically I would expect Helium burning to finish with some c12ag burning adjusting the C/O ratio before Carbon ignition (c12+c12) occurs later at higher rho and T. Instead I'm encountering models where the central core region is almost entirely depleted of Carbon before the convective core vanishes. 

I've noticed this behavior across a variety models regardless of the inlist controls I choose, in stripped and non-stripped stars, when comparing back and forth from r22.11.1 to r15140.

To illustrate this I've taken the "20M_z2m2_high_rotation" test_suite case in r15140 and r22.11.1 and modified them both to be relatively identical. They share the following:
Machine: Mac, OS: Ventura 13.2, sdk version mesasdk-x86_64-macos-22.10.1.
- 20 Msun
- No mass loss
- Henyey convection (though turning on TDC doesn't change the result)
- no semi convection or thermohaline
- no rotation
- similar mesh and temporal resolution
- identical overshooting only in core.
- same abundance, kap, and eos
- identical rate preferences ( jina prefs =2  in r15140, which is essentially the same as r22.11.1 new default)
( I've tested changing any one of these, and  the result in r22.11.1 is identical, so I'm pretty sure none of these inlist controls will effect the result anyway. )

I used the rates module to manually compare the r12+12 and r-3alpha rates to ensure they are identical, and I'm using the same c12ag rate in both models, which I also compared directly and from the rates cache. 

For simplicity, I ran both models to center_he4 = 1d-15, but you could in principle restart from a photo and run them further.

In r15140, At the stopping condition center_He = 1d-15 a significant fraction of c12 remains in the core: center_c12 ~ 0.286
In r22.11.1, At the stopping condition center_He = 1d-15 little c12 remains in the core: center_c12 ~ 3e-6

Here are the models, if anyone would like to try running them, they run in 10-15 minutes on a 10 core machine. I've included pgstar movie in each model directory.
[...]

PR #529, opened 2023-05-02
fixed reverse c12_to_he4_he4_he4 rate

Debraheem:
changed 'do1' to 'do1_reverse' for c12_to_he4_he4_he4 rate, as suggested by Evan Bauer. Fixes Helium burning issue, see https://github.com/MESAHub/mesa/issues/526#issue-1688392000


rjfarmer (changes requested):
Dont add this file

When the build fails you can do:
cp rates/test/tmp.txt rates/test/test_output

and then when net fails do:
cp net/test/tmp.txt net/test/test_output

rjfarmer:
You can add this to the changelog and known bugs:

````
Rates
~~~~~

There has been a bug present in the rate `r_c12_to_he4_he4_he4` since r21.12.1 up to r22.11.1. 
This causes an excessive amount of C12 to be burnt during core helium burning. 
Users are highly recommend to update to the latest MESA.

See `gh-526 <https://github.com/MESAHub/mesa/issues/526>`_`
````

rjfarmer (approved):

evbauer:
I'll note that it looks like we had a refactor of a lot of this code in 15e034ba70aab432ff98b314541ce3fd8d274eac, which came between r21.12.1 and r22.05.1. So the bug didn't exist in this particular form in r21.12.1. For the changelog and known bugs documentation, I think we just want to say that this applies to r22.05.1 and r22.11.1.

Debraheem:
Thanks Evan,

>"I think we just want to say that this applies to r22.05.1 and r22.11.1",

We confirmed that the issue did impact 21.12.1 right? Wouldn't we want to
warn users from using 21.12.1 as well, even if the bug existed in a
different form?

rjfarmer:
Eb can you double check it affected 21.12.1? 15e034b was where the bug was added and that was committed on 2022-02-09. Maybe you had a stale local cache of the rates?

PR #533, opened 2023-05-08
Reintroducing velocity drag for v flag

Debraheem:
In a previous version of mesa (r15140), there existed controls to apply a velocity drag to the outer envelope of a model. This drag functions by converting kinetic energy into heat. This control was useful for preventing instabilities in the envelope from accelerating material at glorious speeds, and is also useful in the preventing radial pulsations and other instabilities that developed near the surface of a model. I've found this control particularly useful for getting 1-7 Msun stars through the thermal-pulsing AGB, and for generating smooth hr tracks in Massive stars from Carbon burning on. I've had success taking massive stars from 15-150 Msun to cc (infall ~ 300 km/s) with this this control.

It should be noted that if you run the drag factor 'drag_coefficient = ' at 1d0 too long, you risk heating the surface up too much and generating a 90 degree kink in the rho-T plane, which can cause some issues. Therefore I reccomend, if  a user is going to turn on this control to be weary. I have found that 'drag_coefficient = ' 0.5d0 or 0.8d0 to be enough for most cases.

Another caveat is that by preventing large accelerations from developing in the envelope, you are preventing surface from being accelerated as much, and therefore your corresponding stellar radius will be affected to some extent.

This control was removed between r15140 and r21.12.1, which I imagine was part of the large amount of internal clean up of the code. I'm hoping it is alright to reintroduce, but let me know what your thoughts are? 

Eventually, it would be nice if we could modify the test_suite "20M_pre_ms_to_core_collapse" to remove
-----
      relax_to_this_tau_factor = 1.5d6
      dlogtau_factor = 0.1d0
      relax_initial_tau_factor = .true.
-----
and replace with something like (maybe)
-----
    drag_coefficient = 1d0 ! or 0.8d0
    min_q_for_drag = 0.8d0
-----

- For now this control only works for v_flag, but eventually a u_flag drag could be reimplemented, which was useful for running ppisn models in an older version (r11701).

**A note, I did not toggle these controls at all, but the build appears to be failing because the test model in star fails. See $MESA_DIR/star/test/log when the build fails.**

### DIFF
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -8136,10 +8136,24 @@
       ! ::
 
     use_dPrad_dm_form_of_T_gradient_eqn = .false.
-    use_gradT_actual_vs_gradT_MLT_for_T_gradient_eqn = .false.
+    use_gradT_actual_vs_gradT_MLT_for_T_gradient_eqn = .false.    
     
     
-    
+      ! drag_coefficient
+      ! ~~~~~~~~~~~~~~~~
+      ! min_q_for_drag
+      ! ~~~~~~~~~~~~~~
+
+      ! only when v_flag.  adjusts both v and energy transfer from kinetic to thermal.
+      ! only for v(k) when q(k) > min_q_for_drag.
+      ! kill off fraction of v = drag_coefficient (i.e. set to 1 to keep v near 0)
+      ! useful for preventing the development radial pulsations during advanced
+      ! burning in massive stars and AGB stars.
+
+      ! ::
+
+    drag_coefficient = 0d0
+    min_q_for_drag = 0d0
     
 
       ! for hydro comparison tests (e.g., Sedov)
--- a/star/private/alloc.f90
+++ b/star/private/alloc.f90
@@ -985,6 +985,9 @@
                   action == do_copy_pointers_and_resize) &
                s% alpha_mlt(1:nz) = s% mixing_length_alpha
 
+            call do1(s% dvdt_drag, c% dvdt_drag)
+            if (failed('dvdt_drag')) exit
+
             call do1(s% vc, c% vc)
             if (failed('vc')) exit
 
--- a/star/private/ctrls_io.f90
+++ b/star/private/ctrls_io.f90
@@ -350,6 +350,7 @@
     P_theta_for_velocity_time_centering, L_theta_for_velocity_time_centering, &
     steps_before_use_TDC, use_P_d_1_div_rho_form_of_work_when_time_centering_velocity, compare_TDC_to_MLT, &
     velocity_logT_lower_bound, max_dt_yrs_for_velocity_logT_lower_bound, velocity_q_upper_bound, &
+    drag_coefficient, min_q_for_drag, &
     retry_for_v_above_clight, &
 
     ! hydro solver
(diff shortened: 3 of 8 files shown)


fxt44:
is this on a branch? i'm not finding it on the mesa test hub commits.

evbauer:
The branch is on Eb's fork rather than on the MESAhub repo, so testhub won't pick it up. We'd need to make a MESAhub branch and merge onto that if we want to go ahead and start testing.

fxt44:
ok. thanks.

evbauer:
You're quite right about the history here. Drag got removed during the major cleanup of January 2021 right after the release of 15140, very early days of the github era. I went digging back through both slack and github and found pretty much no discussion of this. But we do have a nice brief comment in the commit message of PR #156 that got rid of drag:

> remove drag (only used in dev_ppisn - can be added later if needed)
v_drag_factor
v_drag
q_for_v_drag_full_off
q_for_v_drag_full_on
drag_coefficient
min_q_for_drag
turn_on_drag_in_H_envelope

It sounds like you're saying it is needed after all, so there shouldn't be a problem adding this back in. We just got rid of it at the time because it wasn't clear if it was actively being used.

evbauer:
I'll go ahead and merge this over onto a testing branch, and then we can see where things stand.

Debraheem:
Okay, thanks Evan. I'd be interested to know what is breaking, or if it is a personal mistake I made.

I'll try to reintroduce v_drag for u_flag again in the near future. As you mentioned, I've only seen it used in the ppisn case, in particular Rob's models (which I also used in my paper) included it.

evbauer:
Looks like we were just missing an initialization of the `drag` and `drag_energy` variables. That was causing the star unit test to fail because `drag` was injecting a random nonsense term in the case where the `v_flag` wasn't on. Fixing that should at least allow MESA to compile now. We'll see what the rest of the test suite thinks.

evbauer:
> Do we have a test case that flexes this? Or could we resurrect or create a new one?
> 
> I have used it in the past to get through annoying shocks in post-novae, so I'd welcome its return. But, I don't have a canned example showing it working "correctly".

We definitely should! But evidently we got rid of the option in the first place because the test suite wasn't using it, other than what was at the time an experimental dev test case.

evbauer:
I'll also caution that the energy part of this was actually turned off in 15140:
https://github.com/MESAHub/mesa/blob/13fe30d52d41e46845d6e7f1142a7c3d5e84a4f4/star/private/hydro_energy.f90#L275

So we might have some work to do on energy conservation before we're fully ready to resurrect this. Or we could decide we just want it as a stellar engineering option that is allowed to violate energy conservation, but we would need to document that better than it was in previous versions.

Debraheem:
> > Do we have a test case that flexes this? Or could we resurrect or create a new one?
> > I have used it in the past to get through annoying shocks in post-novae, so I'd welcome its return. But, I don't have a canned example showing it working "correctly".
> 
> We definitely should! But evidently we got rid of the option in the first place because the test suite wasn't using it, other than what was at the time an experimental dev test case.

It was previously found in the 'test_ms_to_cc' test case from r15140. This is similar to the '20M_pre_ms_to_core_collapse' test_suite. Maybe we can remove 'relax_to_this_tau_factor' and explore whether we can get a smooth hr to cc with/without drag?

evbauer:
Sounds good. Do you want to go ahead and open a new PR onto the `reintroduce_velocity_drag` branch to add this into one or two of these test cases? It looks like the test suite is [perfectly happy so far](https://testhub.mesastar.org/reintroduce_velocity_drag/commits) now that we got past the initialization issue, but of course we haven't actually turned on velocity drag anywhere yet.

rjfarmer:
I'm think we also need to save dvdt_drag in the photo. But perhaps that can wait untill the implementation is known to work.

Debraheem:
see https://github.com/MESAHub/mesa/pull/535#issue-1708338805

mjoyceGR:
hey, very late to the party but is this in now?

PR #535, opened 2023-05-12
New test suite make_pre_ccsn_13BVN for testing velocity drag, for v_flag

Debraheem:
Inspired by a recent mailing list post, I've brought the make_pre_ccsn test_suite back from the dead. In particular, make_pre_ccsn_13BVN as seen in MESA IV. This test suite is very similar, short of a lot of clean up, and a couple additions (e.g. ledoux criterion, TDC, Chugunov screening mode, OP_split_burn, tighter timesteps to cc, and most importantly v_flag velocity drag which the model cannot evolved without). I've created a special pgstar for the test case that tracks all sorts of things related to energy conservation and the drag force/energy (I've attached a small movie of the model runnning). The model is evolved implicitly from pms to core-collapse (300 km/s infall), no u_flag, and takes ~ 60 minutes to run on my machine (~ 4280 timesteps). This test suite is a bit sensitive, but I've tested in a variety of ways to try to ensure that small changes to things like the mass will still lead to the model completing. As far as energy conservation goes, I did not see any issues in energy conservation or energy error arising from the velocity drag, and observed similar energy errors with v_drag off. If the model is run with no drag, it can still hit cc, but the surface velocities grow very large (> and order of magnitude larger than the sound speed) which has crashed different renditions of this model (just not the one I shared.). The largest challenge with this test case comes from burning in the core. Burning in the degenerate core often introduced quite large errors in energy conservation and I had to find a sweet spot in which to introduce OP_split_burn so as to let the model evolve fully coupled through Si burning-to-depletion, before turning on the operator split (above ~ 4d9 K) to ease Fe burning and eventually core-collapse.

Please test it and make any necessary changes you see fit, or let me know if this is the wrong test case for this. I would implore others to test it as well. (See readme.rst for a small explanation of the test case, but feel free to modify it.)

One thing to note is that I accidentally touched the .mod files in 20M_pre_ms_to_core_collapse, so if you could make sure I didn't break them, as I don't know how to return them to their initial state.

the pgstar movie can be found [here](https://drive.google.com/file/d/16VdSXoSz-ON2EH1rHY68Nia3LbVoespC/view?usp=sharing)


fxt44:
why the conflict with star/private/hydro_energy.f90 ?

Debraheem:
> why the conflict with star/private/hydro_energy.f90 ?

just some spacing issue, I resolved it.

rjfarmer:
To revert the files find the commit before the one you just made and do:

git checkout [commit ID] -- path/to/file

and do that for each file before making a new commit. When we merge we can squash things and hide these changes.

Debraheem:
Thanks Rob, I reverted the .mod files to an old commit before I touched them, though the files changed tab still shows 3 changes made in each .mod file:

  git checkout 48b33f8 -- star/test_suite/20M_pre_ms_to_core_collapse/standard_late_pre_zams.mod
  git checkout 48b33f8 -- star/test_suite/20M_pre_ms_to_core_collapse/standard_zams.mod
  git commit -m "reverted .mod files to older commit, thanks Rob"

Let me know if this looks correct?

Debraheem:
If I delete the .mod files in the "files changed" tab, will that also work?

Debraheem:
Sorry for the mess, I realized why I was having issues. I had two files named "late_pre_zams.mod" and "zams.mod" in addition to the original "standard_late_pre_zams.mod" and "standard_zams.mod". So removing  "late_pre_zams.mod" and "zams.mod" solves this conflict. I comitted two extra files that shouldn't have existed in "star/test_suite/20M_pre_ms_to_core_collapse/" in the first place. Nonetheless, thank you for the help! I understand how git checkout functions now, and I'm slowly adjusting to using git. Maybe my next pull request will be a first without mistakes :)

rjfarmer:
@Debraheem the runtime for this on my laptop is about ~ hour. Would you mind splitting inlist_to_cc to an inlist up to si burning and one until cc?  I can then tweak the rn scripts to skip parts when we don't need the full-run computing.

Issue #534, opened 2023-05-09
relaxing entropy fails when the model is allready partly evolved


rjfarmer:
https://github.com/MESAHub/mesa/blob/d864bf22dbb4dad4b048e5c854f84545e9393931/star/private/relax.f90#LL845C37-L845C64

The problem is we check for when star_age > time for relaxation. But if star_age starts off being > timescale_for_relax_entropy * num_timescales_for_relax_entropy (because the model is already partly evolved) then we take only 1 step in the relaxation process and immediately return from the relaxation.

We should just keep track of the initial star_age when the relax begins and check (star_age - initial_star_age) > ...

relax_angular_momentum_check_model also seems to have this problem. We should also check the other relax routines for the ones based on timescales.

rjfarmer:
Fixed in 3c9bfba2

PR #536, opened 2023-05-13
add star_pointer initial_star_age to track when relaxation begins

Debraheem:
I gave a https://github.com/MESAHub/mesa/issues/534#issue-1702284534 a crack. Let me know if this looks correct? 
Line 845 might be too long? Didn't see any issues in the other relaxation routines. Will double check.

### DIFF
--- a/star/private/relax.f90
+++ b/star/private/relax.f90
@@ -715,10 +715,11 @@
          x(1:num_pts) => rpar(1:num_pts)
          f1(1:4*num_pts) => rpar(num_pts+1:lrpar-1)
          f(1:4,1:num_pts) => f1(1:4*num_pts)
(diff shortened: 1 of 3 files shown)


rjfarmer:
Thanks. What would be better would be instead of adding a new variable in star is to use the rpar arrays. These are used in the relax routines to pass extra info around. If you look in the entropy routine we already set it big enough to hold the number of points where the entropy is defined at. So we can just add one more variable at the end of that (remembering to also increase lrpar which keeps track of how many elements we have). Then in relax_entropy_check_model we can extract back out the last element as the initial_age

Debraheem:
Thanks, I was a little worried about precision. I like your approach much better, it's quite smart!

PR #539, opened 2023-05-19
Add `-std=f2008` flag when compiling `work` folders

### LINKED ISSUE #537: possible issue with dx_limit_min_x in release candidate 23.05.1
Per controls.defaults:

      ! Limit on magnitude of decrease in any cell hydrogen abundance during a single timestep.
      ! dH here is ``abs(xa(j,k) - xa_old(j,k))`` for any cell k and all species j, eg 'h1', 'he4', etc.
      ! special 'species' X (any hydrogen), Y (any helium) and Z (any metals) are
      ! allowed here. E.g. 'Z' will trigger the timestep control if any metal isotope
      ! abundance individually satisfies the conditions below.
      ! Considers all cells except where have convective mixing.

    dX_limit_species(1) = 'h1'
    dX_limit_species(2) = 'he4'
    dX_limit_species(3:) = ''

      ! dX_limit_min_X
      ! ~~~~~~~~~~~~~~

      ! dX limits only apply where xa(j,k) >= this limit.

      ! ::

    dX_limit_min_X(:) = 1d99

However, when I specify

`dX_limit_species(1) = 'h1' `
` dX_limit_min_X(1) = 1d-2`

in `&controls`, I receive

> At line 666 of file ../private/ctrls_io.f90
> Fortran runtime error: Bad data for namelist object dx_limit_min_x

The error is the same when I use `dx_limit_min_X(:)` and similar variations (e.g. specifying the rest of the `dX_limit_species` array)

This is an attempt to replicate use of 
 `dH_limit_min_H = 1d-2` 
 which works with version 15140

warrickball:
Fixes #537, where @mjoyceGR showed that a bad namelist item after an array leads to an error message that points to the array item, rather than the bad namelist item. Following up on the [associated GCC bug report](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=109897), using the `-std=f2008` flag seems to fix this behaviour. This PR just adds that flag in a place that fixes the error. There are still parts of `star` that are compiled without the flag but that doesn't seem to matter.

This led to small changes in `carbon_kh` and `astero_adipls`, which apparently didn't already conform the the Fortran 2008 standard and [failed to compile](https://testhub.mesastar.org/whb%2Fstar-work-std-f2008/commits/e0e56f4).

The test suite is now [passing](https://testhub.mesastar.org/whb%2Fstar-work-std-f2008/commits/dfd53d7).

### DIFF
--- a/astero/test_suite/astero_adipls/src/run_star_extras.f90
+++ b/astero/test_suite/astero_adipls/src/run_star_extras.f90
@@ -103,8 +103,8 @@
             l, order, freq, inertia, x, y, aa, data, nn, iy, iaa, ispcpr, ierr)
          integer, intent(in) :: l, order
          real(dp), intent(in) :: freq, inertia
-         real(dp), intent(in) :: x(1:nn), y(1:iy,1:nn), aa(1:iaa,1:nn), data(8)
          integer, intent(in) :: nn, iy, iaa, ispcpr
+         real(dp), intent(in) :: x(1:nn), y(1:iy,1:nn), aa(1:iaa,1:nn), data(8)
          integer, intent(out) :: ierr
          ierr = 0
          write(*,*) 'astero called my_other_adipls_mode_info'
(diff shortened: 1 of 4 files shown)


evbauer (approved):

evbauer:
Thanks Warrick!

PR #544, opened 2023-06-01
Fix ideal neutral energy

evbauer:
Ideal EOS is currently including the full ionization energy as part of the internal energy, which is inconsistent with its assumption that the matter is fully neutral. I suspect this is just an artifact of porting ideal over from skye_ideal, which does assume full ionization.


fxt44:
looks good to my eyes. i wonder if the "if(temp > 1d4) then" bit introduces a discontinuity.

evbauer:
It definitely does. I got rid of that in the next commit anyway, because ideal always assumes neutral right now. Maybe we don't want to do that, but it's definitely what the EOS does right now...

evbauer:
Testing looks happy so far. I'll go ahead and merge once that's complete, assuming no failures.

Issue #548, opened 2023-06-02
Bad logT guess value for entropy relaxation in eosDE mode


evbauer:
Spurred by this mesa-users post: https://lists.mesastar.org/pipermail/mesa-users/2023-June/014529.html

It looks like the issue here is that the EOS needs to solve for temperature given density and energy, but is unable to converge to the right answer in a reasonable number of iterations because the initial logT guess is bad.

That's set here:
https://github.com/MESAHub/mesa/blob/295a043c38de7f25720cd2d39221d47bda3375da/star/job/run_star_support.f90#L3060-L3062

For very low input energy (~10^10 erg/g), this selects `T_guess_rad` (~10^6 K) when it should select `T_guess_gas` (~10^2 K) as our initial guess. I believe the `max` in that last line should be a `min` to select the most reasonable temperature guess.

fxt44:
nice catch!

might there be cases where the larger temperature (either one) is a more appropriate guess?

evbauer:
I don't think so. My understanding of the logic here is that we want to select whichever of radiation and ideal gas would have a dominant contribution to the energy in a given temperature ballpark, and I think this leads to an understandable sign error in the implementation.

If at fixed internal energy, radiation energy gives a _larger_ `T_guess` value than ideal gas does, then we're actually saying that radiation needs a higher temperature to achieve the relevant energy scale -> ideal gas is dominant -> we should select ideal gas instead of radiation. So I think we always want `min` instead of `max` here.

I guess I should check with @orlox if that all tracks with the right intent, since I assume he was the one that implemented these relaxation features?

fxt44:
this makes sense. thanks. i double check from pablo would be useful.

PR #549, opened 2023-06-05
Fix logic of initial Temperature guess for eosDE entropy relaxation

evbauer:
See #548.

### DIFF
--- a/star/job/run_star_support.f90
+++ b/star/job/run_star_support.f90
@@ -3059,7 +3059,7 @@
                   else
                      T_guess_gas = 2*var2*s% abar(k)*mp/(3*kerg*(1+s% zbar(k))) ! ideal gas (var2=energy)
                      T_guess_rad = pow(var2/crad,0.25d0)
-                     logT_guess = log10(max(T_guess_gas,T_guess_rad))
+                     logT_guess = log10(min(T_guess_gas,T_guess_rad))
                      call eosDT_get_T( &
                         s% eos_handle, &
                         s% species, s% chem_id, s% net_iso, s% xa(:,k), &


warrickball:
This has been open for a while and it looks pretty clear to me that this is the correct fix. Here's a quick Python script to plot what's happening for pure hydrogen:
```python
import numpy as np
import matplotlib.pyplot as pl

sigma = 5.670374419e-5
c = 2.998e10
a = 4*sigma/c
k = 1.380649e-16
m_p = 1.672622e-24

Z = 1
A = 1

T = np.logspace(6, 9, 1001)
Egas = 3/2*k*T/m_p*(Z+1)/A
Erad = a*T**4

kwargs = {'lw': 3}

pl.clf()
pl.loglog(T, Erad+Egas, 'k-', label='total energy', **kwargs)
pl.loglog(T, Egas, label='gas energy', **kwargs)
pl.loglog(T, Erad, label='rad energy', **kwargs)

E = np.logspace(15, 19, 1001)
Tgas = 2*E*A*m_p/(3*(1+Z)*k)
Trad = (E/a)**0.25
pl.plot(np.minimum(Tgas, Trad), E, '--', label='min(Trad,Tgas)', **kwargs)

pl.legend()
pl.xlabel('temperature (K)')
pl.ylabel('energy (erg)')
```
image
It's not the world's finest plot but the main thing is that the black curve is the total energy of an ideal gas + radiation mix at a given temperature or, because it's monotonic, the temperature at a given energy. The dashed green curve is what's selected by choosing the *lesser* of the two temperature estimates at a given energy, which is what this PR suggests.

It might be clearer if you swap *x* and *y* in all the plots:
image

fxt44:
thanks warrick. makes the case of choosing the minimum very clear for this eos.

evbauer:
Thanks! That makes the case very nicely and definitively. I'll go ahead and merge now.

PR #550, opened 2023-06-06
Make other routines self contained

rjfarmer:
Moves all use statements into the other routines. This should make teaching things easier as now people dont need to know to also copy random lines from the top of the file.


fxt44:
nice. thank you!

PR #551, opened 2023-06-13
Other pressure implicit

### LINKED ISSUE #370: Adding a new hook other_pressure_implicit
Will Schultz (@wcschultz) and I have been working to implement some of the results from his recent paper analyzing 3D RHD simulations (https://arxiv.org/abs/2009.01238) in MESA, and are doing this through the other_pressure hook.

We've noticed that the other_pressure hook doesn't provide a way to pass through partial derivatives, and I think we could improve this pretty easily and provide an `other_pressure_implicit` hook.

The basic idea is to promote `extra_pressure` from `real(dp)` to `auto_diff_real_star_order1`. Then the user can set partials with respect to any of the solver variables, and those will get passed right along into the solver, allowing a fully implicit hook.

I think the way we'd implement this is with a new hook, which gets called in `calc_Ptot_ad_tw` in `star_utils.f90`. That's the only place we assemble the total pressure, so that should automatically capture all the places we'd need to evaluate an implicit other pressure.

Any thoughts/suggestions/objections before we give this a whirl?

rjfarmer:
Closes #370

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -10,6 +10,19 @@ Changes in main
 
 .. _Backwards-incompatible changes main:
 
+Backwards-incompatible changes
+------------------------------
+
+Hooks
+-----
+
+The ``other_pressure`` hook has been converted to use ``auto_diff``
+thus the variable ``s% extra_pressure`` is now an ``auto_diff``
+and allows for the setting of the partial derivatives the
+pressure with respect to other variables.
+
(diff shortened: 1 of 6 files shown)


rjfarmer:
Test suite https://testhub.mesastar.org/other_pressure_implicit/commits/acbfa84

evbauer:
Looks good to me!

PR #552, opened 2023-06-13
Pgstar inlist arrays

matthiasfabry:
Large refactor of the pgstar infrastructure (put as draft for now).

The main change is to have the windows with multiple entries (`Text_Summary(1-9), History_track(1-9), History_panels(1-9), Color_magnitude(1-9), Grid(1-9)).` be (mostly) handled in arrays, in the spirit of #368. This reduces the required defaults by several thousands of lines, and cleans up the `*.f90` files of the affected windows, aiding code maintenance.

Only thing that isn't yet possible is having procedure pointers in an array, which means that the decorator pointers are unaffected. As a matter of fact, I would like to push for dropping support for these functions; if users want custom plotting/annotations, I feel like they are much better off using their own library. This would further simplify our `pgstar` implementation.

I need help again however from the `sed` wizards for devising clever substitution scripts, where not only `Grid2_*` is changed to `Grid_*(2)`, but also `Grid2_*(1)` to `Grid_*(2, 1)`, etc.

### DIFF
--- a/star/test_suite/diffusion_smoothness/inlist_diffusion_smoothness
+++ b/star/test_suite/diffusion_smoothness/inlist_diffusion_smoothness
@@ -128,27 +128,27 @@
    pgstar_interval = 1
 
    ! Brunt B profile
-   Profile_Panels1_win_flag = .true.
-   Profile_Panels1_win_width = 14
-   Profile_Panels1_show_grid = .false
-   Profile_Panels1_title = 'diffusion_smoothness'
+   Profile_Panels_win_flag(1) = .true.
+   Profile_Panels_win_width(1) = 14
+   Profile_Panels_show_grid(1) = .false
+   Profile_Panels_title(1) = 'diffusion_smoothness'
 
-   Profile_Panels1_xaxis_name = 'zone'
-   Profile_Panels1_xmin = -101d0
-   Profile_Panels1_xmax = -101d0
-   Profile_Panels1_xaxis_reversed = .true.
-   Profile_Panels1_show_mix_regions_on_xaxis = .true.
+   Profile_Panels_xaxis_name(1) = 'zone'
+   Profile_Panels_xmin(1) = -101d0
+   Profile_Panels_xmax(1) = -101d0
+   Profile_Panels_xaxis_reversed(1) = .true.
+   Profile_Panels_show_mix_regions_on_xaxis(1) = .true.
 
-   Profile_Panels1_num_panels = 1
-   Profile_Panels1_yaxis_name(1) = 'brunt_B'
-   Profile_Panels1_ymin(:) = -101d0
-   Profile_Panels1_ymax(:) = -101d0
-   Profile_Panels1_other_yaxis_name(1) = ''
+   Profile_Panels_num_panels(1) = 1
+   Profile_Panels_yaxis_name(1,1) = 'brunt_B'
+   Profile_Panels_ymin(1,:) = -101d0
+   Profile_Panels_ymax(1,:) = -101d0
+   Profile_Panels_other_yaxis_name(1,1) = ''
 
-   Profile_Panels1_file_flag = .true.
-   Profile_Panels1_file_dir = 'pgstar_out'
-   Profile_Panels1_file_prefix = 'brunt_'
-   Profile_Panels1_file_interval = 10000
-   Profile_Panels1_file_width = 14
+   Profile_Panels_file_flag(1) = .true.
+   Profile_Panels_file_dir(1) = 'pgstar_out'
+   Profile_Panels_file_prefix(1) = 'brunt_'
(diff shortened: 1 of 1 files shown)


fxt44:
is 

setenv MESA_FORCE_PGSTAR_FLAG true

sufficient for these changes get tested within the test suite?

binary/defaults/pgbinary.defaults
```diff
@@ -357,2474 +357,517 @@
...
+    History_Track_ysigma(:) = 0
 
       ! ::
 
-   show_History_Track4_annotation1 = .false.
-   show_History_Track4_annotation2 = .false.
-   show_History_Track4_annotation3 = .false.
+    show_History_Track_annotation1(:) = .false.
+    show_History_Track_annotation2(:) = .false.
+    show_History_Track_annotation3(:) = .false.
```
warrickball, line 421: Did we want this to become a 2d array (I.e. `show_History_Track_annotation(:,:) = .false.`)?

warrickball:
Do you have a list of the regular expressions/find+replace commands you used to make the changes? I've set up a little sandbox to start experimenting but figured I'd just be reinventing what you've already done if you used regexes in the first place.

binary/defaults/pgbinary.defaults
matthiasfabry: I don't know exactly? Philosophically, these are different, as there aren't `pgstar_array_length` of them, only 3. I'd leave this as is.

matthiasfabry:
I have not, I used the find and replace feature of my IDE to do most of the renaming. @mjoyceGR sent me a good starting point to also get into `sed`ing:
```
s/Grid\(.\)_*(/Grid_*(\1, /g
s/Grid\([^_]\)_*/Grid_*(\1)/g
```
If this covers everything, then what remains is to repeat this for `Profile_Panels`, `History_Panels`, `History_Track`, `Text_Summary`, `Color_Magnitude`, including entries in `pgbinary`.

warrickball:
That's helpful, thanks! I've just been experimenting with that in [my sandbox](https://github.com/warrickball/sed-pgstar) and currently have
```
sed -E \
    -e 's/(History_Track|Grid|Profile_Panels|History_Panels|Text_Summary)([1-9])([_a-zA-Z]*) = /\1\3\(\2\) = /' \
    -e 's/(History_Track|Grid|Profile_Panels|History_Panels|Text_Summary)([1-9])([_a-zA-Z]*)\(/\1\3\(\2\,/' \
    inlist_original > inlist_modified
```
where `inlist_original` is the inlist that you want to change. I haven't tested this but it seems to catch basically everything. I generated test data by dumping everything in `*/test_suite/*/inlist_pgstar` to a couple of files and, though there are a bunch of changes on `main`, I didn't see anything obvious left.

I'll try to construct my test data a bit more carefully and iterate again in the next few days but I think this covers most cases.

pmocz:
What is the status of this PR? Is it nearly ready to go?

I can help fix conflicts with main (a lot of it is due to removing unused variables in star, or updates and formatting to inlists) and check the `sed`s

warrickball:
I think it's worth fixing the conflicts with `main` because this is a change we should implement at some point.

My [previous comment](https://github.com/MESAHub/mesa/pull/552#issuecomment-2156782641) is as far as I got with `sed` commands to fix existing inlists. If a sure-fire way of updating inlists is the blocker, I can prioritise this (though that doesn't mean "do it soon").

It's probably worth a look now that we've just made a release. We should have at least a few months to bash our own heads against this before it becomes part of another release.

pmocz:
> I think it's worth fixing the conflicts with `main` because this is a change we should implement at some point.
> 
> My [previous comment](https://github.com/MESAHub/mesa/pull/552#issuecomment-2156782641) is as far as I got with `sed` commands to fix existing inlists. If a sure-fire way of updating inlists is the blocker, I can prioritise this (though that doesn't mean "do it soon").
> 
> It's probably worth a look now that we've just made a release. We should have at least a few months to bash our own heads against this before it becomes part of another release.

I fixed conflicts with main. And applied seds to inlists I could find with grepping. Perhaps we could merge into main after we take another look for outstanding inlists and add an sed script to help users change their inlists, + online documentation, so that it can start being tested before next release?

warrickball:
I'm getting a little bit of MESA momentum back but will be under some time pressure until at least October 10. What are the specific actions here? I'm guessing I can test the `sed` completeness with something like:

1. Checkout out this branch.
2. Checkout all the `inlist_pgstar` and `inlist_pgbinary` files from `main`.
3. Run the `sed` script on all of them.
4. Run the test suite.

If there are failures, we can then fix them, rinse and repeat. If not, we're basically most of the way we can get in terms of how well the test suite currently covers all the PGstar options, at which point I think we're okay to merge.

PR #553, opened 2023-06-16
[ci optional] Implemented Sunny Wong rotation brunt change

### LINKED ISSUE #372: brunt in rotation
In $MESA_DIR/star/private/rotation_mixing_info.f90, lines 595-596, the total Brunt and Brunt composition term are calculated again. I was wondering if it would be equivalent to simply replace those lines with s% brunt_N2 and s% brunt_N2_composition_term ?

The attached image shows a MESA profile comparing these, where I let a WD accrete at constant Mdot, turn Tayler-Spruit on only, and run with small timesteps (0.002 yr) until the model settles into a quasi-steady state. 
I extracted the necessary lines from rotation_mixing_info and put them as extra_profile_columns_data in my run_star_extras. In the middle panel, the N2_mu calculated (orange solid) is much noisier than the usual s% brunt_N2_composition_term (green solid).
For this WD I  averaged the inner 0.85 Msun abundances, so there should be no N2_mu in the inner 0.85 Msun. 

This may have an impact on the Tayler-Spruit viscosity because it is calculated as the harmonic sum between the structure term and composition term. In the final panel, I take the shear profile from MESA (q in top panel), and then calculate the Tayler-Spruit viscosity, one using lines 595-596 (orange line) and one using brunt_N2_composition term from profile_columns.list (green). The former is agrees with the viscosity that the MESA profile gives (blue line), but the latter is what the viscosity should be as the inner 0.85 Msun has no composition gradient .

nu

rjfarmer:
People seemed happy to merge this in #372 so let's check to see if the test suite passes and then merge it.

Closes #372

### DIFF
--- a/star/private/rotation_mix_info.f90
+++ b/star/private/rotation_mix_info.f90
@@ -590,10 +590,8 @@
             Hj(1) = Hj(2); Hj(nz) = Hj(nz-1)
 
             do i = 1, nz
-               dlnRho_dlnP = s% grad_density(i)
-               dlnT_dlnP = s% grad_temperature(i)
-               N2(i) = -grav(i)*(1/gamma1(i) - dlnRho_dlnP)/scale_height(i)
-               N2_mu(i) = -grav(i)/scale_height(i)*(1/chiRho(i) - delta(i)*dlnT_dlnP - dlnRho_dlnP)
+               N2(i) = s%brunt_N2(i)
+               N2_mu(i) = s%brunt_N2_composition_term(i)
             end do
 
             do k=1,nz


rjfarmer:
https://testhub.mesastar.org/brunt_in_rotation/commits/3f511c0

evbauer:
Hmmm, looks like we have another instance of a photo checksum failure without an fpe: https://testhub.mesastar.org/brunt_in_rotation/commits/69b4927/test_cases/star/ppisn

I suspect there's something non-trivial going on related to the state of brunt-related quantities.

evbauer:
Actually, it looks like this change is definitely not equivalent to the previous way of doing things. We have a line specifically skipping updates to `s% brunt_N2` during solver iterations:
https://github.com/MESAHub/mesa/blob/3fb71741486e8397b8292b23d4562a2ac311ccd7/star/private/hydro_vars.f90#L596-L598
https://github.com/MESAHub/mesa/blob/3fb71741486e8397b8292b23d4562a2ac311ccd7/star/private/solver_support.f90#L844

So the quantities we're trying to take advantage of here are actually going to become stale during solver iterations. I'll push an experimental commit getting rid of that check to see what testing looks like, but I'd definitely like some input from @orlox before proceeding with a merge.

evbauer:
Actually, I'm realizing that was probably a red herring, as we don't update mixing info during solver iterations anyway, so it shouldn't matter if the brunt is getting updated during the iterations here. I'll revert that.

The photo checksum failure does seem to be telling us something important, and it's happening quite consistently, so we'll have to track that down.

rjfarmer:
If you diff the output of two runs it looks like there are differences immediately after the restart.  So I'm guessing there is an issue with how the Brunt is computed at the start of a run that differs from how the old form worked. 

Also, https://testhub.mesastar.org/brunt_in_rotation/commits/e03088c/test_cases/star/ppisn shows it's not a parallelisation issue.

evbauer:
@orlox and I discussed working on this some more today. Not clear on what our timeline will be yet, but I think we at least see a path forward that involves mostly trying to work through all the fpe's we can squash related to rotation and mixing.

evbauer:
No worries! My cluster was backed up with a bunch of other stuff I'm running anyway, so I happened to have paused automated testing anyway. So nothing even needs to be canceled in this case.

Debraheem:
cleaned it up :) , now to figure out where the fpe error is coming from...

PR #557, opened 2023-07-06
Resurect the ionization module

rjfarmer:
We deleted the ionization code in  5485cec175f04bcb017007d6d81dcb087ca20635. However, it turns out there is
interest in the community to have this information available. This is coming from people doing common 
envelope simulations who would like to know where the ionization zones are to get to be able to compute the energies in the envelopes.

We deleted this originally as there was no expertise in the dev group to keep this working, answer questions about it,
and it had not been documented anywhere as to what it does and how it does it.

So this commit resurrects the code and gets things compiling again.

Things to address before merging back:

- Who maintains this going forward?
- We need to write up the method and document it somewhere. I'm told @LarsBildsten has/had a paper draft from many years ago which could be used as a basis for this?
- We need to document that it might return values inconsistent with the eos as things are computed slightly differently.
- Related, I just noticed that the preprocessor has defined its own set of physics constants that should be ported to use mesa/const.
- I have only tested that MESA compiles and will generate output in a profile file, more testing is needed.
- The ionization_ output looks by-eye to be reasonable but the typical_charge_ values just seem to report 0 for me so that needs debugging further (code comments say it should never return 0).


fxt44:
thanks rob. this is one piece i've never really delved into as it appeared out of the blue. now that you've resurrected it, i'll take a deeper look at what its doing. ... and all good questions ...

star/private/diffusion_procs.f90
```diff
@@ -1184,7 +1184,7 @@ subroutine setup_struct_info( &
 
 
          subroutine do1(k)
-            use mod_typical_charge, only: eval_typical_charge
+            use ionization_lib, only: eval_typical_charge
```
evbauer, line 1187: `eval_typical_charge` got ingested into the `star` module when we deleted ionization, so we'll need to make a decision about whether to get ride of `mod_typical_charge` from `star` and go back to the old way of doing things, or keep this new way.

The ionization module is mostly about something else, so I think my preference would be to keep `eval_typical_charge` as its own independent thing still living in `star`, and delete the things associated with it out of the `ionization` module.

evbauer:
I believe this is ready to merge back in now. Do we also want to reintroduce the option to include the ionization as profile column output? That would involve reverting parts of #146.

Right now, this is still mostly undocumented, so I'd prefer not to give easy access to ionization profile columns that might confuse users, at least until we can offer better support to explain where the physics comes from. Hopefully the common envelope use case mentioned above can get what is needed by writing code to call the `eval_ionization` routines directly?

Should this get a changelog entry?

evbauer:
Let's merge this back in to clear the release-blocker aspect of the issue. This at least makes the code available to those who wish to experiment with it. No changelog entry or profile column outputs until this is better documented.

PR #558, opened 2023-07-08
shmesa

earlbellinger:
Command line utilities for MESA. See the [README](https://github.com/MESAHub/mesa/blob/4ed45b4121dbf997880e25da8d0d0281a134a1f2/scripts/shmesa/README.md) for usage. 

As per the suggestion in the [Discussion](https://github.com/MESAHub/mesa/discussions/517), it is now to be run as a standalone program rather than to be sourced. Therefore, recommended usage will be to add `PATH=$PATH:$MESA_DIR/scripts/shmesa` to your `~/.bashrc`, and then you can call `shmesa` from anywhere.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -108,6 +108,9 @@ uses export to set variables):
     export MESASDK_ROOT=/Applications/mesasdk
     source $MESASDK_ROOT/bin/mesasdk_init.sh
 
+    # add shmesa (the MESA command line tool) to your PATH 
+    PATH=$PATH:$MESA_DIR/scripts/shmesa
+
 
 If your machine uses csh as its shell, use ``setenv`` instead of ``export``.
     
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -109,7 +109,7 @@ uses export to set variables):
     source $MESASDK_ROOT/bin/mesasdk_init.sh
 
     # add shmesa (the MESA command line tool) to your PATH 
-    PATH=$PATH:$MESA_DIR/scripts/shmesa
+    export PATH=$PATH:$MESA_DIR/scripts/shmesa
 
 
 If your machine uses csh as its shell, use ``setenv`` instead of ``export``.
--- a/scripts/shmesa/shmesa
+++ b/scripts/shmesa/shmesa
@@ -3,24 +3,28 @@
 #### MESA SHMESA 
 #### Command line utilies for MESA 
 # provides commands such as `shmesa change` and `shmesa grep` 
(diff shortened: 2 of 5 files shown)


docs/source/installation.rst
```diff
@@ -108,6 +108,9 @@ uses export to set variables):
     export MESASDK_ROOT=/Applications/mesasdk
     source $MESASDK_ROOT/bin/mesasdk_init.sh
 
+    # add shmesa (the MESA command line tool) to your PATH 
+    PATH=$PATH:$MESA_DIR/scripts/shmesa
```
evbauer, line 112: I think we want an `export` here to get the path update into the user's environment.

earlbellinger: I thought $PATH is exported by default (so doesn't need to be re-exported). Nevertheless I've added it as there's no harm.

scripts/shmesa/shmesa
```diff
@@ -0,0 +1,455 @@
...
+        dir_to_zip=$1
+        zip_name="${dir_to_zip}_mesa.zip"
+    fi
+
+    if [[ ! $(find "$dir_to_zip" -type f -iname "*inlist*" -o -iname "*.mod") ]]; then
+        echo "Warning: No files with 'inlist' in their names or files ending in '.mod' found in the directory '$dir_to_zip'."
+    fi
+
+    # Create a zip file containing only inlists and models
+    zip -r "$zip_name" "$dir_to_zip" -i '*inlist*' '*.mod'
```
evbauer, line 256: Can we include `src/` directories here too so that `run_star_extras` get bundled up too? And maybe some other files like `*.list` and `*.net` would be helpful as well?

In playing with this just now, I also noticed that `make/run_star.mod` and `make/run_star_extras.mod` got picked up by this, but the mod files are fortran objects that we don't want to share. Is there an easy way to exclude those?

earlbellinger:
some extra planned features, including some suggestions from the summer school participants: 

- `max`/`min`/`first`/`last`: point at `history.data` or a profile file, supply a column name and get the relevant output (e.g., `shmesa max LOGS/history.data star_age`) 
- `closest`: point at `history.data` or a profile file and supply a column and a value as well as additional column names. `closest` will find the row whose value is closest to the supplied one and print the values of the other columns belonging to that row (e.g., `shmesa closest center_h1 0.1 model_number star_age`)
- `extras`: replace the standard `run_star_extras.f90` file with the full template 
- `add`: the `shmesa change` functionality only changes existing inlist parameter names; `add` could insert names into the inlist (e.g., `shmesa add inlist_project controls do_element_diffusion .true.`)

evbauer:
I love the `extras` idea!

PR #559, opened 2023-07-14
Bjorklund wind

matthiasfabry:
The massive star community would like to see the Bjorklund+21 wind mass loss rates made available in MESA as an alternative for the Vink+01 rates


evbauer:
I'd suggest adding a quick changelog entry, and then I think we can merge.

docs/source/changelog.rst
```diff
@@ -28,7 +28,11 @@ Previously we had logic to determine if a extra history value should be saved
 as an int or a float (users can only provide data as a float). This was error
 prone, so now we save extra history values as floats.
 
+new features
```
warrickball, line 31:
```suggestion
New features
```

Issue #493, opened 2023-02-13
Sould eps_nuc contain eps_nuc_neu_total?


rjfarmer:
So dealing with #488 and #492 made me realise I think we are inconsistent with eps_nuc and the nuclear neutrinos (luckily i think this is only an output issue).

I think #492 is still the right fix to bring split burn consistent with the non-split burn case. But we have problems in the output.

In net/ we already subtract the nuclear neutrino losses from eps_nuc:

approx nets:
https://github.com/MESAHub/mesa/blob/6cc0f72fbc2e082968e82ed89b8fae0ae899b01e/net/private/net_eval.f90#L386

soft nets:
https://github.com/MESAHub/mesa/blob/6cc0f72fbc2e082968e82ed89b8fae0ae899b01e/net/private/net_derivs_support.f90#L180

As net/ is already subtracting the losses, nothing outside of net/ should ever do  eps_nuc- eps_nuc_neu

Except in the profile output we have:
https://github.com/MESAHub/mesa/blob/6cc0f72fbc2e082968e82ed89b8fae0ae899b01e/star/private/profile_getval.f90#L470
https://github.com/MESAHub/mesa/blob/6cc0f72fbc2e082968e82ed89b8fae0ae899b01e/star/private/profile_getval.f90#L473

and history.f90
https://github.com/MESAHub/mesa/blob/6cc0f72fbc2e082968e82ed89b8fae0ae899b01e/star/private/history.f90#L835

Luckily when packing eps_nuc into the matrix we just use eps_nuc:
https://github.com/MESAHub/mesa/blob/6cc0f72fbc2e082968e82ed89b8fae0ae899b01e/star/private/hydro_energy.f90#L238

So i think we just need to clean up the profile/history output and add a note somewhere so we don't do this again? @fxt44

PR #566, opened 2023-07-21
Update ZAMS models and how they are generated.

wmwolf:
Reduce stopping condition for ZAMS models to L_nuc/L < 0.999. This seems to prevent models from depleting too much hydrogen before stopping.


wmwolf:
Using @evbauer's nifty plot scripts, here's what we get now. The high mass models are a little suspicious now; perhaps we are truncating these before they are fully on the main sequence?

Another option would be to use the more local constraint suggested by @orlox: |dL/dM-eps_nuc|/L < delta everywhere in the model (for some reasonable choice of delta).
CenterXfrac
HR

evbauer:
This looks much better! I suspect the problems on the very high mass end have to do with the fact that for `initial_mass > 90`, the models get created as 90 Msun models and then relaxed upward in mass. So they start off somewhat too luminous and too hot at the center, but not terribly far from the MS, so maybe not a huge problem.

Would there be any downside to changing the Lnuc/Lzams factor to 0.99 instead of 0.999? That way we would catch them even earlier, but I think they would still always have to be very close to the MS.

evbauer:
I tried with 0.99 instead of 0.999, and here's what I got. I guess it isn't all that much better, but the dip at 1.26 is somewhat smaller.

iTerm2 zgXej6 CenterXfrac

data/star_data/zams_models/create_z2m2_y28/src/run_star_extras.f90
```diff
@@ -379,7 +379,7 @@ integer function evolve_to_zams_check_model(s, id, lipar, ipar, lrpar, rpar)
          real(dp), intent(inout), pointer :: rpar(:) ! (lrpar)
          evolve_to_zams_check_model = bare_bones_check_model(id) 
          if (evolve_to_zams_check_model /= keep_going) return
-         if (s% L_nuc_burn_total >= s% L(1)/Lsun) then
+         if (s% L_nuc_burn_total >= 0.999d0 * s% L(1)/Lsun) then
```
warrickball, line 382: Is this basically the same as the [control used](https://docs.mesastar.org/en/release-r23.05.1/reference/controls.html#lnuc-div-l-zams-limit) in the work folder?

https://github.com/MESAHub/mesa/blob/ee3de3fd1f0c26f3abe9344becf4cde0e5a39cc6/star/work/inlist_project#L48-L50

If so, could we use that control, in case people want to generate their own MS tables without having to modify this `run_star_extras.f90`?

wmwolf: I believe it ends up being identical to how stopping near ZAMS works in inlist controls. Since it's the same for all masses, we can probably move it into the inlist, but given the non-standard setup in this directory, that might have ramifications I haven't thought through.

I'll give it a try.

Debraheem:
Hello, I gave it a try using:
----
         if (s% L_nuc_burn_total+s% power_nuc_neutrinos >= s% L(1)/Lsun) then
            evolve_to_zams_check_model = terminate
            s% termination_code = t_extras_check_model
         end if
----
I think this is could be a more robust stopping condition? What people commonly think of as ZAMS is when the photon-luminosity from nuclear reactions in the core is equvalent to the surface surface photon luminosity. This expression removes the neutrino component, so the photon-luminosity in the core from nuclear reactions does not exceed the surface luminosity of the star due to neutrino losses. This ties back to the definition of L_nuc_burn_total which comes from power_nuc_burn, which is calculated from (n% eps_nuc = n% eps_total - n% eps_neu_total).  see https://github.com/MESAHub/mesa/issues/493. It's not significantly better, but visually it looks like a very-slight improvement over  using (s% L_nuc_burn_total >= 0.99d0 * s% L(1)/Lsun)? 

It could also be interesting to try (s% L_nuc_burn_total+s% power_nuc_neutrinos >= 0.999d0 * s% L(1)/Lsun)?

CenterXfrac
HR

evbauer:
This does looks pretty good, but isn't this expression you've implemented **adding back in** the energy from neutrinos that should just be free streaming out of the star? I'm not sure this actually makes physical sense, unless I'm losing track of a sign somewhere.

data/star_data/zams_models/create_z2m2_y28/src/run_star_extras.f90
wmwolf: I tried this, but the control is not respected. I think it's because this part of the evolution is triggered by the `evolve_to_check_point`, which can only be stopped by `extras_check_model` (or so the comment in `star_lib.f90` seems to indicate). So for now, this needs to just live in `run_star_extras.f90`.

Debraheem:
So I'm in agreement with both of you, but I think my take is slightly different:
In my opinion, by using Lnuc, you are factoring in neutrinos.

The definition of L_nuc_burn_total by default subtracts off the neutrinos from nuclear reactions
 (L_nuc_burn_total = Lnuclear_reactions - Lnuclear_neutrinos)
, so adding Lnuclear_neutrinos back in "+ s% power_nuc_neutrinos"  lets you know what the actual photon luminosity diffusing through the stellar envelope is instead of the original 'photon_luminosity - neutrino_luminosity'. The envelope of the star doesn't feel the effects of neutrinos since the neutrinos are radiated out of the core, so canonically the thermal timescale of the envelope should be determined by the photon luminosity from nuclear reactions alone while the thermal timescale of the core is determined by the interplay between the photon and neutrino luminosity, e.g. Lnuc ? 

It just depends what you mean by ZAMS. If you are asking for when the photon luminosity from the surface is equivalent to the 'photon luminosity of the core', then you'd adopt my suggestion, but if you are asking for when the photon luminosity of the surface is equivalent to the 'net energy produced in the core (regardless of form: neutrino vs photon)' then you would adopt the original condition just using L_nuc_burn_total.


Nonetheless, I'm super curious how Pablo's condition will perform!

evbauer:
Ok, I think I follow your reasoning here @Debraheem. However, I think my preference would be to try to do the absolute simplest thing that should robustly produce what we would identify as ZAMS upon inspection. So I'd advocate sticking to Lnuc/L = 0.99 as a stopping condition, unless we can identify a specific potential pathology of that condition that we want to avoid.

wmwolf:
An update to my quest to do a more "robust" solution: In short, I don't think this is a good idea because it is too complicated.

### More Information
In general, the value
$$\mathrm{max}\left(\left\vert\frac{dL}{dm}-\epsilon_{\mathrm{nuc}}\right\vert\right)$$
does drop sharply as a star enters the ZAMS. However, the minimum values varies mostly directly with the mass. I tried computing a bunch of pre main sequence models, determined the minimum such value, multiplied it by three to give the models some breathing room, and then fit a polynomial to get the stopping condition as a function of mass. It is still not robust over the entire mass range, with some stopping too early and other stopping too late.

I think for one given pre-main sequence evolution track, identifying the local minimum in this quantity  while $X_1 > 0.69$ or so is actually an excellent way to **define** the ZAMS a bit more rigorously, but predicting where to stop with this technique *a priori* is not practical over the large range of masses in question.

### One last option
When this *does* work, it produces very smooth HR diagrams, central hydrogen mass fractions that are consistently around 0.695, and $L_{\mathrm{nuc}}/L$ that are within one or two percent of unity. So I'm going to try using the central hydrogen abundance as a proxy instead to see if it is qualitatively "better". If not, I think we just go with $L/L_{\mathrm{nuc}} \geq 0.99$ and call it a day.

fxt44:
hi bill - do you have a plot of the polynomial and how did over the range of masses examined?

fxt44:
you should write this up!  i'm unaware of any literature discussing various zams definitions in drop-from-a-cloud pre-ms models. research note for a lower bar or a manuscript for a higher bar.

warrickball:
How well are you resolving the stopping condition in time?

wmwolf:
I'm not doing anything special in my models; it's just using the default tolerances. I'll try tightening that up to get a little more uniformity in the final central hydrogen concentrations.

orlox:
One extra comment regarding the dL/dm condition. It is best to use a normalizing value in there. So the value I've checked in the past is:

$$\left|\frac{dL}{dm} - \epsilon_\mathrm{nuc}\right|\frac{M}{L} < \delta$$

This prevents having to use a delta value with units, although one still needs to explore what is a reasonable value.

But in any case I think any choice that gives a sensible ZAMS line and similar Xc should be good enough.

wmwolf:
I've set the stopping condition to be when the central hydrogen mass fraction drops below 0.697 (but no more than $10^{-5}$ below that) and regenerated the models. The plots look fine (see below). I also merged new commits from `main` back in to this PR. I did a squash and merge, which I think was the correct thing to do. Unless there are no objections, I think we can merge this into `main`.

### Central Hydrogen
Pardon matplotlib's tick labeling choice, but all vertical values are very close to 0.697
CenterXfrac

### HR
I do like that we get a smoother HR diagram with this stopping condition.
HR

wmwolf:
I've moved the criterion for the ZAMS to using `x_ctrl(1)` in the inlist for easier maintenance. I didn't push the "new" model data (which should be identical) to prevent LFS waste, so this should affect any test results. I think we can merge this into `main`.

PR #569, opened 2023-07-26
Increase maximum nuclear network size

rjfarmer:
Conventionally we have been limited in the size of a mesa problem
(zones * isos) by the size of the solver_work arrays, which was
~ (zones * isos* isos* A). Where A is a constant based on the fact that the
solver work arrays contain multiple sub-arrays that get split apart
into the different arrays needed. A was about ~10.

By splitting the work array up into its consistent parts each individual
array can now reach SIGNED_INT_MAX (2^31). So we should scale closer to
the maximum problem size being max_isos = sqrt(2^31 / zones). For
1000 zones that is potentially ~1400 isos. Whether we can actually hit
that depends on whether there are other arrays that scale worse than
isos * isos * zones.

Most of the changes are just the code of do_solver being indented. I did this
as I moved the code to a new contained procedure so we can guarantee calling
cleanup() after it finishes. If we miss calling cleanup() we will leak a lot
of memory very quickly.

I've tested (with the help of @mathren) mesa_495.net which runs and needs about ~80GB
of RAM. If we ever try the ~1400 that would need about ~1400^2/495^2 * 80GB = ~650GB of RAM.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -24,10 +24,23 @@ pressure with respect to other variables.
 run_star_extras
 ---------------
 
-Previously we had logic to determine if a extra history value should be saved 
+Previously we had logic to determine if a extra history value should be saved
 as an int or a float (users can only provide data as a float). This was error
 prone, so now we save extra history values as floats.
 
+
+
+
+Maximum net size
+----------------
+
+Previous MESA versions where limited to ~300 isotopes in the nuclear network,
+this has now been (slightly) relaxed.
+In principle the maximum network size is sqrt(2^31/nz). Thus for nz=1000 zones
+this is about ~1400 isotopes, at 5000
+zones this is ~650 isotopes. Note this will require a huge amount of RAM,
+``mesa_495.net`` requires at least ~80GB of RAM, and it is estimated
+that for 1400 isotopes you will need ~650GB of RAM.
 
 
 Changes in r23.05.1
--- a/star/private/star_solver.f90
+++ b/star/private/star_solver.f90
@@ -36,7 +36,7 @@
       implicit none
 
       private
-      public :: solver, get_solver_work_sizes
+      public :: solver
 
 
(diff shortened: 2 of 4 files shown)


evbauer:
Awesome Rob!

My cluster has some `bigmem` nodes with 512 GB of RAM. Do we want to push this a bit further with something like `mesa_833.net`? I guess it's not obviously necessary to do something like that before merging, but I'm game to help out if we want to try throwing this at something requiring up to ~500 GB.

mathren:
@evbauer I've actually tested up to `mesa_833.net` and got successful runs (but at default spatial resolution, `mesh_delta_coeff=1.0`). I guess you need to go to `mesa_1508.net` or decrease `mesh_delta_coef` for heavier tests...

rjfarmer:
If anyone wants to take a stab at a 1000+ I say go for it, 1500 might too big though so we'd need a custom net that's a few hundred isotopes smaller.

Just note that the ZAMS models have ~3000 zones during model 1, so I'm not sure if we can push much larger without making a new starting model, that starts with fewer zones, as sqrt(2**31/3000) ~ 846.

Debraheem:
Is there a specific test case you're using for this? I have access to a few 2TB memory nodes on our ASU cluster and could give it a stab.

rjfarmer:
Just try star/work with a net about 1000 isotopes.

Debraheem:
I tested the work directory on my macbook which dynamically allocates swap, using some custom networks, just taking a couple timesteps to test that it works:
# zones = 1078

mesa_204~ (sucessful)
mesa_495 ~  (sucessful)
mesa_833 ~  (sucessful)
mesa_1121 ~  (sucessful)
mesa_1281 ~  (sucessful)
mesa_1384 ~ (sucessful)
mesa_1432 ~ segfault (failure) 

So the failure point is roughly where Rob guessed it was.

PR #570, opened 2023-07-27
New default c12ag rates

Debraheem:
Hello,

I was thinking it could be useful to update the c12ag rates in MESA away from JINA, and instead to the most recent median rate recommended by the meta-analysis/review  done by [Deboer et al. 2017](https://link.aps.org/doi/10.1103/RevModPhys.89.035007), with updated relolution from [Mehta et al. 2022](https://iopscience.iop.org/article/10.3847/1538-4357/ac3130/meta). I have included rate text files for the +-3 3 sigma uncertainties as well. 

I have not added any documentation yet, but that can be done before a merge. I more or less want to test the rate across the test_suite and see if it causes any "issues" in other models. I expect the median 0p0 sigma rate to weaker than the current default JINA rate, producing models with more higher C/O ratios in their cores. 

 If we decide that it is not ideal to make this the default rate, maybe we could still offer them for use in the $MESA_DIR/data/rates_data/rate_tables folder? This is a zero-low maintenance inclusion in the code. 

Let me know if you have any thoughts or opinions on this?


rjfarmer:
So I'm not opposed to putting the rates in the rate tables folder as an easy way for people to access them. But making the default? I am hesitant, though maybe this should be a decision for the people who will be left supporting this.

My concerns with replacing rates is a more general issue with rate options, rather than anything wrong with this PR:

1. The rate selection code is complex and it is already hard to determine where a particular rate comes from. If we start manually adding new rates this will get harder.  
2. Would adding one new rate open the gates to updating all the other rates? Are we prepared to say no to nuclear physicists who want their updated rate added to MESA?
3. Are we the right people to be making the choices on individual rates? The deBoer rate has already been updated in [Shen et al 2023](https://ui.adsabs.harvard.edu/abs/2023ApJ...945...41S/abstract). So is the deBoer version even the right one to use if we did limit ourselves to just updating the c12(a,g)o16 rate? What happens in a few years when I assume another update will come out?

Ideally, we would just update to a newer REACLIB which includes new rates but REACLIb's probably not going to have an update any time soon. So we might need to make a decision, like with the wind routines, what the defaults should be and how much we want to add ourselves/letting users customise things.

fxt44:
my opinion is not to make a new default. too many real and potential support issues. the current mechanism for using any tabulated rate seems sufficient for users (including experimental nuclear folks)  wishing to try new rates - e.g., https://smnalsu.home.blog

Debraheem:
Thanks, you both raised great points. I'm now in agreement that it probably shouldn't become the default. 

To Rob's point "Would adding one new rate open the gates to updating all the other rates? Are we prepared to say no to nuclear physicists who want their updated rate added to MESA?"

Should I include the rates in the first place or save this for mesa-contrib? And yes, [Shen et al 2023](https://ui.adsabs.harvard.edu/abs/2023ApJ...945...41S/abstract) nominally alter's the +sigma rates making them stronger, but shouldn't affect the median or weaker rates. I don't have access to these updated rates, which begs your question "What happens in a few years when I assume another update will come out?".

rjfarmer:
I think there is enough interest in the deboer rate to warrant shipping it (and its sigma's) inside the $MESA_DIR/data/rates_data/rate_tables folder to make it easier to use. Especially as you have the high-res version which I don't think is available anywhere else?

fxt44:
https://ui.adsabs.harvard.edu/abs/2022ApJ...935...21C/abstract

and its embedded  zenodo url constituted the public release of the high resolution probability distribution function.

rjfarmer:
I think at this point we just need a short changelog to say these rates are now available but otherwise its is good to merge.

evbauer:
I support having these particular rates easily available in MESA. I would definitely make use of this, and I suspect many others will as well.

And while it probably is wise to avoid making it the default for simplicity of navigating the rate logic, I'd at least suggest we switch some test cases to use the Deboer rate. E.g. `make_co_wd` and at least some massive stars. We can always open a new PR to work on incorporating this into test cases if we want to go ahead and merge this one.

rjfarmer (approved):

PR #571, opened 2023-07-30
Overhaul docs for scaling relation parameters and update solar reference values

### LINKED ISSUE #298: Clarify/document solar values in asteroseismic scaling relations
Adam pointed out at that we have the following option in controls.defaults

```
     ! Teff_sun
      ! ~~~~~~~~
      ! solar value of Teff
      ! ::
    Teff_sun = 5777d0
```

and asks why Teff_sun isn't just defined using the value in const/public/const_def.

Warrick confirms this is the value that enters asteroseismic scaling relations, along with delta_nu_sun and nu_max_sun. The default should be from the same source as those two values. 

It is also a bit funny that other solar values enter into the scaling relations, but only Teff gets this special parameter treatment. 

This and related things could use a little tidying.

warrickball:
To close #298.

Expands description of scaling-relation-related parameters
```
get_delta_nu_from_scaled_solar
nu_max_sun
delta_nu_sun
Teff_sun
```
and revises the last three reference values to 3078 μHz and 134.91 μHz (from [Lund et al. 2017](https://ui.adsabs.harvard.edu/abs/2017ApJ...835..172L/abstract)) and 5772 K (from IAU B3).

Not sure if this deserves a changelog entry.


rjfarmer:
If we are fixing the docs, does it make sense to rename ``Teff_sun`` to something like ``astero_teff_sun``? Just to make clear its only being used for asteroseismology and not elsewhere (where the constant ``Teffsun`` is used instead)

warrickball:
> If we are fixing the docs, does it make sense to rename `Teff_sun` to something like `astero_teff_sun`? Just to make clear its only being used for asteroseismology and not elsewhere (where the constant `Teffsun` is used instead)

I think so. That was part of the original confusion. I'll rename it to `astero_Teff_sun` unless someone proposes a better name / set of names for these controls.

warrickball:
Renaming `Teff_sun` to `astero_Teff_sun` touched `history_columns.list` across all the test suites, which is why all the reviews have been requested.

I also deleted a line that implies the values in the scaling relation are fixed, which they are not.

I'll leave this up over the weekend before merging.

evbauer (approved):

warrickball:
I decided this deserved a changelog entry because we're changing a control (`Teff_sun` → `astero_Teff_sun`), so might break inlists.

PR #572, opened 2023-07-30
Delete unused libraries from GitHub runner

warrickball:
The CI has started failing because it runs out of space before the build can complete. This appears to be a common enough issue that a common workaround, compiled into [this action](https://github.com/marketplace/actions/free-disk-space-ubuntu) (for example), is to delete/uninstall a bunch of unused software that's included on the GitHub runner.

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -1,6 +1,6 @@
 name: Build on Linux with SDK
 
-on: [pull_request, release]
+on: [push, pull_request, release]
 
 jobs:
   build:
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - name: Delete unused packages
         run: |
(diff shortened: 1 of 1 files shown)


rjfarmer (approved):

evbauer:
As long as somebody can confirm that the sed commands I just pushed make sense, let's go ahead and merge.

earlbellinger:
I just took a look and we see 

```
...
      logical :: use_cache_for_eos = .true.
      logical :: eos_root_is_initialized = .false.

      character(len=1000) :: eosDT_cache_dir
      character(len=1000) :: eosDT_temp_cache_dir

      logical :: eos_test_partials
      real(dp) :: eos_test_partials_val, eos_test_partials_dval_dx ! for dfridr from star
      
      
      contains
      
      
      subroutine eos_def_init
         integer :: i
         type (DT_XZ_Info), pointer :: eosDT_XZ_ptr, FreeEOS_XZ_ptr
         include 'formats'
         
         use_cache_for_eos = .true.
...
```

so you may wish to add `/g`.

PR #573, opened 2023-08-01
Deploy the Deboer C12agO16 rate in test suite

evbauer:
Right now I've just replaced all instances that were using the Kunz rate with Deboer instead. Is that the right approach?


Debraheem:
The rate should have almost no impact on the simplex_solar_calibration test case, so maybe it's unnecessary there. Can we get this running on the test hub to see how it impacts these specific test cases?

rjfarmer:
https://testhub.mesastar.org/new_c12ag_tests/commits/f0791c4

Looks like 12M_pre_ms_to_core_collapse failed, it seems to have hit max model number.

Debraheem:
I'm taking a closer look.

Debraheem:
> I definitely agree in principle. Why were are we using the special Kunz rate there right now?

I'm unsure, but the simplex_solar_calibration inlists could use a touch up. We can correct this in a separate pr.

rjfarmer:
We use it there as before I simplified the rate selection code we switchee to just using nacre. Why it was set to nacre in the first place I don't know.

evbauer:
Ok, we can try just getting rid of any special rate in simplex_solar_calibration and see what happens.

Debraheem:
In reference to '12M_pre_ms_to_core_collapse', I think the core in the model is slightly more degenerate with the weaker deboer c12ag rate than with the previous default and so it just takes longer to run. In both cases, there are large temperature swings in central zones due to the use of op_split_burn, but it's slightly more difficult in the deboer c12ag model and therefore takes longer.  Truthfully split burning doesn't shine in these degenerate cc models, as you can see the temperature swings in both 'main' and 'new_c12ag_tests' outputs on the testhub, followed by hard retry limits

new_c12ag_tests: https://logs.mesastar.org/f0791c4974f589d2b35185f78fd433ba59072528/cannon_conroy_intel/12M_pre_ms_to_core_collapse/out.txt

main:
https://logs.mesastar.org/e3c44576ca531e482d9cad6517fa276145ce3ecc/cannon_conroy_intel/12M_pre_ms_to_core_collapse/out.txt


-One option is to increase the max_model_number so the model can hit cc

-Another is to  increase  'op_split_burn_min_T = 1.0d9' ---> 4.0d9, in which case I get the model to hit cc in 6440 timesteps with 778 retries (> 600 occuring after log_center_T > 9.6).

evbauer:
Thanks! That at least provides some good physical intuition for why the model behavior would change with this new rate. @rjfarmer do you have a preference for which approach we take to fix this one up?

rjfarmer:
Given it's a very slow test case I suggest we pick whichever method makes it finish fastest. So that seems like increasing the temperature in op_split_burn.

Debraheem:
I will add the new rate to 20M_pre_ms_to_core_collapse as well shortly.

Debraheem:
@evbauer Don't we need to update all the standard*.mod files in these test_suites as well?

evbauer:
Yes, good point. I tend to forget to do that unless the saved models are causing test suite breakages, but in this case it would make sense to regenerate standard mods for several of these.

Also, quick reminder that you can include `[ci optional]` in your commit message to make sure that Cannon knows not to skip the optional parts when doing automated testing. In this case we want that for `12M_pre_ms_to_core_collapse` because it's only failing for complete runs that include the optional earlier stages.

PR #577, opened 2023-08-10
GYRE update to 7.1, and updates GYRE/GSM output files

rhdtownsend:
It's been a while since the GYRE distro bundled with MESA was updated. This PR brings in GYRE 7.1, which was [released earlier today](https://github.com/rhdtownsend/gyre/releases/tag/v7.1), together with changes to the output of GYRE/GSM-format files. The version of the file output can now be controlled by a new gyre_data_schema control, which is set at a default value 101 (version 1.01). Setting this to 120 will output v1.2 GYRE/GSM files (new with the GYRE 7.1 release), which include eps_grav data to allow simulations of pulsations in non-thermal-equilibrium stars.

The big possible breakage with this PR is that the GYRE 7.1 build requires MESA SDK >= 23.7.2. (I think it will probably build OK with older versions of the SDK, but it will fail the self-tests if the user opts to run them). If reviewers don't like this requirement, I can cook up a script that will modify the SDK version check after the GYRE 7.1 archive is unpacked.


rhdtownsend:
Well, that's funny -- it looks like @warrickball updated GYRE to 7.0 a few months ago! So, this PR represents a less momentous update, but still is important for the eps_grav stuff.

evbauer:
Nice! I don't have a strong opinion about the SDK requirement if you think it would be good to nudge people to upgrade.

I suppose I should update my SDK on Cannon to get the automated testing working on this branch?

evbauer:
Oh, actually it looks like builds are failing for a different reason. I'm realizing now that you said MESA should build okay, but might fail the optional self-tests. A note in the changelog is probably sufficient for that.

Right now I think the compilation is failing because we're just missing an update to `astero/private/gyre_support.f90` to get rid of `GYRE_MODEL_VERSION` there.

```
../private/gyre_support.f90:207:67:

  207 |     call gyre_set_model(global_data, point_data, GYRE_MODEL_VERSION)
      |                                                                   1
Error: Symbol 'gyre_model_version' at (1) has no IMPLICIT type
make: *** [makefile_base:116: gyre_support.o] Error 1
make: *** Waiting for unfinished jobs....

/home/runner/work/mesa/mesa/astero/make
FAILED
```

fxt44:
same, under mesasdk-*-macos-23.7.2 .

rhdtownsend:
That should all be fixed now.

fxt44:
so good so far on 4 machines ...

evbauer:
FYI, the Cannon failures at the moment are spurious. Just problems with the scratch performance due to a "group of improperly architected jobs" by some other user apparently, so you can ignore those for now.

evbauer:
Uh oh, it looks like @wmwolf's accidental force push of his local `eps-grav-gyre` branch had completely overwritten the branch and therefore the contents of this PR... I just pushed what I had on my local machine back onto the branch, so I think it's restored now? @rhdtownsend can you check that the contents of this PR look right and that the branch on GitHub is up to date with your local version?

rhdtownsend:
Looks OK to me. I’ve made one more commit to remove the old (gyre-7.0) release tarball. I think we’re ready to pull the trigger, if others agree.

cheers,

R

evbauer:
Great! Can we just get a quick changelog entry? Then I think it's ready to merge.

PR #588, opened 2023-09-22
Reformatting of pgstar documentation

TrystonRaecke:
These changes:
- Fix some typos in the pgstar documentation
- Fix some formatting for proper display in browser
- Clarify language throughout, removing many references to "this", replacing them with the appropriate explicit variables 
- Reinforce a common documentation format for each window type in pgstar with consistent labelling and ordering

I did not alter any variable names, but did move some around in their respective sections for clarity. 
I mostly opted to use bolded text as opposed to level headings, as the heading levels were not quite clear when output to html and Sphinx seems to not like repeated heading text.

### DIFF
--- a/star/defaults/pgstar.defaults
+++ b/star/defaults/pgstar.defaults
@@ -205,13 +205,6 @@
     pgstar_age_coord = -0.04
     pgstar_age_fjust = 0.0
 
-      ! ### axes line weight
-
-      ! ::
-
-    pgstar_box_lw = 3
-
-
       ! titles
       ! ~~~~~~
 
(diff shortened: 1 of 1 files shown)


matthiasfabry:
Thanks for this Tryston! Is it easy to share a built `html` of the `pgstar` page only so we can quickly see how this looks?

TrystonRaecke:
> Thanks for this Tryston! Is it easy to share a built `html` of the `pgstar` page only so we can quickly see how this looks?

Yep, here is the zipped pgstar html:
[pgstar.html.zip](https://github.com/MESAHub/mesa/files/12707997/pgstar.html.zip)

Let me know if anything else is helpful!

Issue #589, opened 2023-09-27
Ideal EOS Returns Negative Heat Capacity at high T / low Rho


evbauer:
See top left corner of attached plot from EOS plotter for just the ideal gas+radiation EOS at solar composition, where red represents a bad value (either NaN or negative value returned by ideal EOS).

Ran into this while working with Shelley Cheng on some models with high mass loss rates. Looks likely to be something like a floating point subtraction issue, so maybe @fxt44 can offer some advice on how to patch this sort of thing up.

The red in the lower-right corner of the plot is a separate issue described in issue #565.

![iTerm2 EXsVld eos_plotter](https://github.com/MESAHub/mesa/assets/18405113/a884ca9c-ac0c-4164-8cf9-b295824fd41b)

evbauer:
Edit: had a dyslexic moment when I first wrote the title. It's *high* temperature and *low* density that's relevant for this issue. The opposite corner is the topic of #565.

evbauer:
The underlying issue here is that $c_P = c_V \Gamma_1 / \chi_\rho$ and $\chi_\rho \rightarrow 0$ for high temperature and low density because the radiation pressure dominates over the gas pressure. Where we're having problems is the $\chi_\rho \lesssim 10^{-16}$ region. Floating point errors occasionally return a negative value there.

Maybe the simplest hack to avoid this is to set a floor of `1d-99` somewhere around here:
https://github.com/MESAHub/mesa/blob/acd763a682009d91e6ff6f1023b9cf72d21bf8dc/eos/private/skye_thermodynamics.f90#L65

Would that be good enough or should we consider trying something more sophisticated to avoid the floating point issues in the first place? I'll want to check the other EOS quantities in this region as well to see if anything else is returning bad values.

My main motivation here is just to help models avoid crashing in early Newton iterations that might touch this regime. Obviously stellar profiles shouldn't actually live here, but I always want to squash bad numerics in the input physics wherever I can find them so that we can give MESA a better chance at iterating toward a correct solution.

![iTerm2 Jvdx2V eos_plotter](https://github.com/MESAHub/mesa/assets/18405113/1f6859d9-3ff4-4ab1-84b4-956114f8e9f1)

fxt44:
[...]

PR #591, opened 2023-09-30
Avoid some numerically messy subtractions in the EOS

evbauer:
This overwrites the radiation pressure that `auto_diff` would calculate to be $(\rho^2/\rho^2)aT^4/3$. Getting rid of the $(\rho^2/\rho^2)$ term avoids numerical problems for derivatives of radiation pressure with respect to $\rho$.

Should solve #589.


fxt44:
are you sure an only explicit  prad is needed, not erad and srad?

evbauer:
> are you sure an only explicit prad is needed, not erad and srad?

Nope! I'm not sure at all. I'll take a look at that soon. I've converted this PR to a draft in the mean time. Thanks for the guidance!

fxt44:
i suspect you'll find 'em if you scour other derivative based quantities! broadly, in the "chemical picture" its best (as in it avoids swamping issues) to do radiation, ions, and e+e- separately and add.

evbauer:
@rhdtownsend I was a bit surprised to see that `gyre_in_mesa_rsg` is fragile enough to be sensitive to tiny changes in the numerics of the EOS. I ended up changing the target frequency by almost 4%. Does that fix sound okay to you?

PR #592, opened 2023-10-06
Workaround for adipls issues with BLOCK DATA

rhdtownsend:
Test cases now pass on gfortran and ifort systems. Should be good to go.


warrickball:
I trust that the CI builds will be fixed after merging, so the only thing is if someone has tested this latest workaround on a Mac. If so, I think we should merge.

fxt44:
f3548e0 test suite on an intel mac rolling ...

PR #596, opened 2023-11-09
replace sin/cos with multiangle formula in neutrino cooling

zingale:
this reduces the number of expensive intrinsics needed

### DIFF
--- a/neu/private/mod_neu.f90
+++ b/neu/private/mod_neu.f90
@@ -1033,19 +1033,19 @@
          sin1 = sin(u)
 
          ! double, triple, etc. angle formulas
-         ! sin/cos (2 fac1 tau)
+         ! sin/cos (2 u)
          sin2 = 2.0d0 * sin1 * cos1
          cos2 = 2.0d0 * cos1 * cos1 - 1.0d0
 
-         ! sin/cos (3 fac1 tau)
+         ! sin/cos (3 u)
          sin3 = sin1 * (3.0d0 - 4.0d0 * sin1 * sin1)
          cos3 = cos1 * (4.0d0 * cos1 * cos1 - 3.0d0)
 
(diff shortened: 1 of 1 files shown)


zingale:
this is a backport of a similar change made in AMReX-Astro Microphysics:

https://github.com/AMReX-Astro/Microphysics/blob/39a0f40188bc424a7e52d855378031d27e20cde6/neutrinos/sneut5.H#L830

for us, it sped up the neutrino calculation by a few %

warrickball:
I'm not too surprised that changing the specific operations has led to bit-level differences in the output. If I get time I'll do it but someone should check by hand that the changes are small and commit the new reference results.

Incidentally, since it might help readability rather than performance, I guess one could also use the formulae

sin([n+1]θ) = sin(nθ)cos(θ) + cos(nθ)sin(θ)
cos([n+1]θ) = cos(nθ)cos(θ) - sin(nθ)sin(θ)

to write something like

```f90
sin1 = sin(fac1*tau)
cos1 = cos(fac1*tau)

sin2 = 2*sin1*cos1
cos2 = cos1*cos1 - sin1*sin1

sin3 = sin2*cos1 + cos2*sin1
cos3 = cos2*cos1 - sin2*sin1

! etc
```
(Feels a lot like we should be looping over an array here...)

zingale:
fair enough.  Our main motivation here was for GPUs where these intrinsics are very expensive.  Also, some math libraries have a combined sincos() that can save some ops.

evbauer:
This looks reasonable to me. I'll go ahead and change the target and merge over into a testing branch where we can fix up the test output and see how the whole test suite goes. `ndiff` should immediately let us know if any of the test output changes are significant.

warrickball:
Just to be clear, I think these are good changes, they just prompted me to consider other ways of reusing the once-off sines and cosines.

PR #597, opened 2023-11-09
replace sin/cos with multiangle formula in neutrino cooling (part II)

evbauer:
Continuation of #596

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -10,6 +10,7 @@
 *.mp4 filter=lfs diff=lfs merge=lfs -text
 data/star_data/zams_models/zams_z2m2_y28.data filter=lfs diff=lfs merge=lfs -text
 docs/source/assets/zams_z2m2_y28_patched.data filter=lfs diff=lfs merge=lfs -text
+neu/test/test_output filter=lfs diff=lfs merge=lfs -text
 
 # Try to get github to better understand what our filetypes are
 *.inc linguist-language=Fortran
--- a/neu/private/mod_neu.f90
+++ b/neu/private/mod_neu.f90
@@ -694,17 +694,26 @@
 
    !..equation 3.7, compute the expensive trig functions only one time
          cos1 = cos(fac1*tau)
-         cos2 = cos(fac1*2.0d0*tau)
-         cos3 = cos(fac1*3.0d0*tau)
-         cos4 = cos(fac1*4.0d0*tau)
-         cos5 = cos(fac1*5.0d0*tau)
-         last = cos(fac2*tau)
-
          sin1 = sin(fac1*tau)
-         sin2 = sin(fac1*2.0d0*tau)
-         sin3 = sin(fac1*3.0d0*tau)
-         sin4 = sin(fac1*4.0d0*tau)
-         sin5 = sin(fac1*5.0d0*tau)
+
+         ! double, triple, etc. angle formulas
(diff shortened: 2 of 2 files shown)


evbauer:
`ndiff` shows that the largest relative change in the test output is ~7e-12.

warrickball:
Doesn't look like any tests failed but I'm not sure if we need to look more closely at the results. I was trying to have a quick look at a case that might depend relatively sensitively on neutrino rates. I noticed `20M_pre_ms_to_core_collapse`'s  results have changed. e.g. `fe_core_mass` seems to have gone [from 1.5](https://testhub.mesastar.org/main/commits/ece1a5c/test_cases/star/20M_pre_ms_to_core_collapse) before these changes [to 1.58](https://testhub.mesastar.org/update_sneut/commits/e390304/test_cases/star/20M_pre_ms_to_core_collapse) afterwards.

I didn't expect these changes to make a difference.

fxt44:
maybe the neutrino emission routine be tested in standalone  for one composition over the rho-T plane (unit test)? one with the original trig functions and one with the new trig functions, and diff the two.

evbauer:
> maybe the neutrino emission routine be tested in standalone for one composition over the rho-T plane (unit test)? one with the original trig functions and one with the new trig functions, and diff the two.

Isn't this what our built-in installation unit test is already doing? The test output from looping over the following ranges is where `ndiff` returned differences no larger that 7e-12.
https://github.com/MESAHub/mesa/blob/d215b38166687af0351c621f47ddf622415625d7/neu/test/src/neu_support.f90#L12-L16

fxt44:
yea, you are right. i'm mystified then how apparent changes of les than ~7e-12 induce an iron core mass change of nearly 0.1 msun.

warrickball:
Okay, my next thought: do we trust the test case? Is it exceptionally sensitive anyway? I'll have a brief look later to see if there are any other test cases that change suspiciously.

I'll also try to have a look at what the unit test calculates and how it depends on this change. Relative changes of 7e-12 were already bordering on suspiciously large.

Debraheem:
I've checked the high mass test cases and I think we are okay with the Fe core mass change. A majority of the difference between the 'main' and 'update_sneu' branch for the 20Msun_pre_ms_to_cc test case was a result of restarting from a .mod file over photo restarts and the control 'min_dq_for_xa = 1d-4' which limited the convergence of these models. I have also looked at the ppisn test case and the final mass seems to be reasonably converged. The 20Msun and 12Msun test cases are due for a rehaul, but I'd like to make quite a few changes so I will leave it out of this branch.

By adopting the double angle formula, I am finding that bit for bit changes in neu are useful in measuring the sensitivity of massive star model to numerical noise. I'm in favor of adding an inlist control which one can use to toggle between both formulations of these double formula, unless there is a better way to seed bit for bit differences in a model?

fxt44:
_... restarting from a .mod file over photo restarts ..._

why do we prefer .mod over photos on restarts in the test suite? one is complete, the other is not ...

Debraheem:
I think we are fine to merge this pull request!

evbauer:
> why do we prefer .mod over photos on restarts in the test suite? one is complete, the other is not ...

This might be worth discussing on one of the dev calls in terms of overall design choices. Do we have a sense of what would be the biggest drivers of state differences from these different kinds of restarts?

Issue #599, opened 2023-11-10
finite `op_split_burn_min_T_for_variable_T_solver` causes segfault


mathren:
**Describe the bug**
When using `op_split_burn = .true.`,  together with a finite value `op_split_burn_min_T_for_variable_T_solver` (instead of the default `1d99`), one obtains an immediate segfault. 

**To Reproduce**
I uploaded [here](https://users.flatironinstitute.org/~mrenzo/materials/MWE.tar.xz) a minimal work example for r23.05.1 that runs a 3Msun star and turns on `op_split_burn` at logT=7.3. At step 67 the error occurs (the `photos` folder contains a photo calculated with r23.05.1 at step 65 for convenience). 

**Expected behavior**
I realize this is probably due to something not fully implemented -- I would recommend mentioning this is an experimental setting not ready for use in the documentation/controls.default
 
**Screenshots**
I copy here the full backtrace. Line 122 in `$MESA_DIR/net/private/net_initialize.f90` is an empty line (in `r23.05.1`).

```
Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
In file '../private/net_initialize.f90', around line 122: Error allocating 7432401508362160128 bytes: Cannot allocate memory

Error termination. Backtrace:
#0  0x7f9966f7308f in ???
	at /build/glibc-SzIz7B/glibc-2.31/signal/../sysdeps/unix/sysv/linux/x86_64/sigaction.c:0
#1  0x827202 in __net_initialize_MOD_setup_net_info
	at ../private/net_initialize.f90:95
#0  0x8288e9 in __net_initialize_MOD_setup_net_info
	at ../private/net_initialize.f90:121
#2  0x833503 in __net_burn_const_density_MOD_burn_const_density_1_zone
	at ../private/net_burn_const_density.f90:242
#1  0x833503 in __net_burn_const_density_MOD_burn_const_density_1_zone
	at ../private/net_burn_const_density.f90:242
#3  0x818d15 in __net_lib_MOD_net_1_zone_burn_const_density
	at ../public/net_lib.f90:1090
#2  0x818d15 in __net_lib_MOD_net_1_zone_burn_const_density
	at ../public/net_lib.f90:1090
#3  0x75f660 in burn1_zone
	at ../private/struct_burn_mix.f90:1139
#4  0x75f660 in __struct_burn_mix_MOD_do_burn._omp_fn.0
#4  0x75f660 in burn1_zone
[...]

PR #606, opened 2023-11-21
partial-Fix const density burn solver

Debraheem:
See https://github.com/MESAHub/mesa/issues/599. I've moved the control for toggling the const density solver into developer defaults until the const density solver is debugged or removed.


evbauer:
Testing looks good, so I think we should go ahead and merge. Should this close #599, or should we leave that open?

Debraheem:
I think we can close the issue for now, and leave the branch open.

PR #608, opened 2023-11-30
added x_{type}_ctrl inlist controls to binary controls

jaiken17:
This is a small quality of life change that adds ```x_{type}_ctrl``` inlist controls to the ```binary``` module mirroring those already present in the ```star_data``` module. This allows users to define system-specific controls in the ```binary_controls``` namelist instead of using the ```controls``` namelist specific to one of the stars. This is simply because sometimes it it would be preferable to put a special control in ```inlist_project``` rather than in ```inlist1``` when it applies to the whole binary system.

These controls can then be used in ```run_binary_extras.f90``` in a similar way to those in the star_data module:
```
   some_var = b% s1% x_ctrl(1)   ! star_data

   some_var = b% x_ctrl(1)       ! binary
```

The documentation in ```binary_controls.defaults``` has also been mirrored from that in ```controls.defaults```.

### DIFF
--- a/binary/defaults/binary_controls.defaults
+++ b/binary/defaults/binary_controls.defaults
@@ -1494,3 +1494,16 @@
     use_other_CE_binary_evolve_step = .false.
     use_other_CE_binary_finish_step = .false.
     use_other_e2 = .false.
+
+
+! extra params as a convenience for developing new features
+! note: the parameter ``num_x_ctrls`` is defined in ``binary_def.f90``
+
+! ::
+
+    x_ctrl(1:binary_num_x_ctrls) = 0d0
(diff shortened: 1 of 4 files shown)


matthiasfabry:
Thank you very much for this! Although, some of the more senior devs will have to explain to me why we now have `x_ctrl` in `star_controls` and also `xtra` in star_data_step_work`. Do these serve some different intended purpose?

orlox:
Yes, they serve different purposes. x_ctrl are meant to be used through inlists to specify run options. xtra are convenience variables that you can use to store things mid run. In practice you could use the x_ctrl to serve the same purpose as xtra, but I think it's good to keep both distinct.

PR #611, opened 2023-12-18
Add O/Ne Phase Separation Option for Massive WD Crystallization

evbauer:
This is a fairly straightforward extension of the existing C/O phase separation infrastructure to include an option to use the appropriate diagram for O/Ne mixtures in massive WDs.

### DIFF
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -7274,13 +7274,26 @@
       ! do_phase_separation
       ! ~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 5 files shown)


warrickball:
Can we have at least one test case go through each `phase_separation_option`?

evbauer:
Maybe. This would require creating a new test case for ultramassive WD cooling, something like `wd_cool_1.2M`. But I'm not especially eager to try supporting recommendations for that scenario without putting some more thought into it beyond just O/Ne phase separation. What do you think?

Issue #583, opened 2023-08-31
missing atm options


Debraheem:
Hello, the following atm options don't appear to exist in the code any more, but they remain in the documentation. Does anyone know if they should be re-implemented or removed? I looked in $MESA_DIR/star/private/atm_support.f90, but no case exists for these options?

'paczynski_grey'	[Paczynski Grey Atmosphere](https://docs.mesastar.org/en/latest/atm/paczynski.html)
'fixed_Teff'	[Fixed Effective Temperature](https://docs.mesastar.org/en/latest/atm/fixed-teff.html)
'fixed_Tsurf'	[Fixed Surface Temperature](https://docs.mesastar.org/en/latest/atm/fixed-tsurf.html)
'fixed_Psurf'	[Fixed Surface Pressure](https://docs.mesastar.org/en/latest/atm/fixed-psurf.html)
'fixed_Psurf_and_Tsurf'	[Fixed Surface Pressure and Temperature](https://docs.mesastar.org/en/latest/atm/fixed-psurf-tsurf.html)

evbauer:
It looks like `Paczynski_grey` was present in releases up through r15140, but it's not clear that it ever actually worked after the `atm` overhaul circa r12115. E.g. I just tried to run a vanilla 1 Msun model with that option selected in r12778 and I just repeatedly get
```
 do1_create_Paczynski_atm failed in dopri853: idid          -2
 failed in do1_create_Paczynski_atm
 retry       1
```

Bill deleted the `Paczynski_grey` option in #143.

evbauer:
It looks like fixed_Teff, fixed_Tsurf, fixed_Psurf, and fixed_Psurf_and_Tsurf were deleted in #200.

Debraheem:
Thanks Evan, If there is no response to your mailing-list, I'll make a pull to remove these options from the documentation.

fxt44:
yep, circa jan 2021

remove Paczynski atm options
Paczynski_grey
trace_atm_Paczynski_grey
Paczynski_atm_R_surf_errtol
use_T_Paczynski_outer_BC

no real reason for it other than cleanup as it didn't seem to used much (or in the test suite).

on fixed_Teff, fixed_Tsurf, fixed_Psurf, and fixed_Psurf_and_Tsurf, i recall those were used in early versions of the conductive_flame test suite case. now it uses use_other_surface_PT = .true. to do its boundary condition magic.

warrickball:
[...]

PR #612, opened 2023-12-25
reimplement fixed atm options

Debraheem:
A first pass at reimplementing The fixed surface options. See https://github.com/MESAHub/mesa/issues/583

I have tested it to be working locally for 'fixed_Tsurf' in the ppisn test case. 
Feel free to clean up any of it. Do we need a test case for this?
Should this be moved into atm_support.f90, or is it alright to live in hydro_vars.f90?

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -90,6 +90,10 @@ starting models has been recalculated with a more stringent stopping condition,
 and now all pre-computed ZAMS models have a central hydrogen mass fraction very 
 near 0.697.
 
+The ``fixed_Teff``, ``fixed_Tsurf``, ``fixed_Psurf``,  and ``fixed_Psurf_and_Tsurf``
+atmosphere options have been reimplemented, although we caution users that their
+implementation might conflict with mlt_option = ``TDC``.
+
 Changes in r23.05.1
 ===================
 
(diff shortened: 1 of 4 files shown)


fxt44:
just a comment. fixed_Tsurf was implemented to do conductive flames in ~10 gram balls, https://ui.adsabs.harvard.edu/abs/2020ApJ...891....5S/abstract . fixed_Tsurf was then removed in favor of the use_other_surface_PT hook, which one can see in the conductive flame test suite case, https://docs.mesastar.org/en/release-r23.05.1/test_suite/conductive_flame.html#conductive-flame .

warrickball:
I'm happy to see this restored as is. The design question needn't delay users at least having the same implementation they had before.

Not a dealbreaker for me but could we modify `conductive_flame` to use one of these options? AFAICT it's used it's own `constant_surface_PT` routine since at least r15140, which is before the fixed atm options were removed. If we do ever try to redesign this, we'll want tests.

warrickball (approved):

fxt44:
if this test can use one of the options, we can experiment!  the current form has some educational value in showing an example of changing boundary conditions using hooks. but maybe we have enough similar other examples within the test suite.

earlbellinger (approved):

Debraheem:
I have added documentation and tested that the the conductive_burn test_suite produces almost identical results regardless of the choice: "other_surface_pt" hook or 'fixed_Psurf_and_Tsurf' atm option. The results are very slightly different, but I think this is the result of exponentiating the inputs to put them into 'fixed_Psurf_and_Tsurf' and the need to set Pextra_factor = 0d0 when using atm_option = 'fixed_Psurf_and_Tsurf'.

PR #613, opened 2023-12-25
update massive star test suites

Debraheem:
Following https://github.com/MESAHub/mesa/pull/597, I have updated
12Msun_pms_to_cc, 20Msun_pms_to_cc, zams_to_cc_80, and make_pre_ccsn_13bvn.

All should run smoother. The 12 Msun should take roughly half the time. 
The 20 and 80 Msun are now solar metallicity and their surfaces are at tau = 1. 

I have added a run_all script for users who would like to photo restart instead of .mod restart.


evbauer:
It might be helpful to rename `rn_fxt` and `re_fxt` to something slightly more descriptive? Maybe `rn_nomodfiles` and `re_nomodfiles`.

fxt44:
adding to changelog seems best. "small ones", however defined, could be clipped for a wider distribution as part of a release candidate.

evbauer:
Testing looks good! I think we're ready to go ahead and merge.

PR #623, opened 2024-02-15
fix bug in photo derivative

Debraheem:
'denomdt' should be 'denomdd'. If the test_suite passes, this should be fine for a merge.


evbauer:
This is the good stuff! One character, but an important one!

fxt44:
this bug is/was even in my standalone versions. good catch eb!

Debraheem:
Everything looks good for a merge.

Issue #626, opened 2024-02-28
`fe_core_infall_limit` not behaving as expected? 


mathren:
**Describe the bug**

Trying to run models to the onset of core-collapse, I encounter (unsurprisingly) lots of numerical difficulties. One particularly annoying one is that models reach the termination condition `fe_core_infall_velocity` not because of a large infall velocity within the Fe core, but instead because of velocity "spikes" in the outer layers (for a stripped model with no H envelope, but I've seen this in H-rich pre-SN models as well).

For example, using the `trace_history_value_names` to get terminal output close to core-collapse, I can find models that successfully reach the termination condition according to MESA and report:

```
                                 Fe_core        1.8818177594246457
                          fe_core_infall    1.1951597418765050D+04
                      non_fe_core_infall        0.0000000000000000
                               rel_E_err    1.2628869479237786D-05
```
and correspondingly:
```
stop because fe_core_infall > fe_core_infall_limit
    0.1195159742E+10    0.1000000000E+09
```

However, the attached pgplot shows the fe56 mass fraction profile (in cyan, right y-axis, ending at 2Msun, or, according to the terminal output 1.88 -- seems good), and the velocity profile (in yellow, left y-axis units cm/s). Clearly ,all the fe-rich material is still with v=0, but the surface (where numerical artifacts create a non-zero, and at this timestep negative, v) hits the threshold for collapse and the model terminates:

![image](https://github.com/MESAHub/mesa/assets/49067756/c7200aa7-fc97-445f-b744-8dce624521c9)

The relevant `inlist` lines in my setup are:
```
non_fe_core_infall_limit                 = 1d99
non_fe_core_infall_mass               = 1d99

fe_core_infall_limit                         = 1d8
fe_core_infall_mass                       = 1.0d0 !same behavior with default 0.1d0 or 2.0d0 (while m_fe~1.88 ?!)
when_to_stop_rtol                          = 1d-2
when_to_stop_atol                         = 1d-3
```
where I set `non_fe_core_infall_mass` and `non_fe_core_infall_limit` so that the model would never stop because of collapse outside of the fe core, but this does not seem to achieve the intended purpose.

[...]

PR #627, opened 2024-02-29
check that infall velocity is within core

mathren:
addresses issue #626 by adding a check on the mass coordinate of index `k_min` less than `fe_core_mass` or `he_core_mass` (for `non_fe_core_infall_limit`), where `k_min` is the index of the minimum of the infall velocity. This is intended to prevent numerically meeting the infall velocity stopping condition because of a spike at the surface after the formation of a non-zero fe core.

Also fixes the calculation of `mass_sum` (used to check that enough mass has a sufficient velocity compared to the parameter `fe_core_infall_mass` or `non_fe_core_infall_mass`): previously the loop was adding *mass coordinates* of each shells `s% m(k)`, instead of the shell mass `s% dm(k)`.

Tested on 12M_pre_ms_to_core_collapse and 20M_pre_ms_to_core_collapse in the test suite.

This does not attempt to make `drag_energy` optional or implement a `velocity_tau_upper_limit`, for which I will discuss on the side with @Debraheem.

### DIFF
--- a/star/private/report.f90
+++ b/star/private/report.f90
@@ -78,8 +78,8 @@
                   end if
                end do
             end if
-         end do  
-                
+         end do
+
          if (s% eps_nuc_factor == 0d0) then
             s% power_nuc_burn = 0d0
             s% power_nuc_neutrinos = 0d0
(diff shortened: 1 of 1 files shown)


Debraheem:
Awesome, I will merge this into https://github.com/MESAHub/mesa/tree/MR/fix_infall_velocity, and get it running on the test hub.

PR #628, opened 2024-02-29
fix infall velocity

Debraheem:
@mathren's fix for the infall velocity. See https://github.com/MESAHub/mesa/pull/627#issue-2161886494 and https://github.com/MESAHub/mesa/issues/626.

### DIFF
--- a/star/private/ctrls_io.f90
+++ b/star/private/ctrls_io.f90
@@ -39,14 +39,14 @@
  character (len=strlen) :: controls_namelist_name
 
  namelist /controls/ &
-
+ 
     ! where to start
     initial_mass, initial_z, initial_y, initial_he3, &
-
+    
     ! definition of core boundaries
     he_core_boundary_h1_fraction, co_core_boundary_he4_fraction, one_core_boundary_he4_c12_fraction, &
     fe_core_boundary_si28_fraction, neutron_rich_core_boundary_Ye_max, min_boundary_fraction, &
-
+    
     ! when to stop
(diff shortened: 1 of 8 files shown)


Debraheem:
@earlbellinger @evbauer @mathren  looks good to merge.

star/private/ctrls_io.f90
```diff
@@ -350,8 +350,8 @@ module ctrls_io
     include_P_in_velocity_time_centering, include_L_in_velocity_time_centering, &
     P_theta_for_velocity_time_centering, L_theta_for_velocity_time_centering, &
     steps_before_use_TDC, use_P_d_1_div_rho_form_of_work_when_time_centering_velocity, compare_TDC_to_MLT, &
-    velocity_logT_lower_bound, max_dt_yrs_for_velocity_logT_lower_bound, velocity_q_upper_bound, &
-    drag_coefficient, min_q_for_drag, &
+    velocity_logT_lower_bound, max_dt_yrs_for_velocity_logT_lower_bound, velocity_tau_lower_bound, velocity_q_upper_bound, &
+    use_drag_energy, drag_coefficient, min_q_for_drag, &
```
evbauer, line 354: We're missing a couple lines in this file to make these controls functional. We need to assign what gets read in here to the corresponding members of the star info structure. It's a small enough fix that I think I can just push it directly to main, but I wanted to flag that here for future reference.

evbauer: Fixed in a5c60befba1506ab836cc22c624a45af2078a578.

Issue #605, opened 2023-11-16
Port Adrian's Python Implementation of Fraser 2023 Thermohaline into MESA


evbauer:
Working on implementing
https://ui.adsabs.harvard.edu/abs/2023arXiv230211610F/abstract

afraser3:
(Recall that in BGS13, HG19, and this model, once you get w_f it's a simple formula to get Nu_mu)

For the not-so-great model in section 5.1 of that paper (the dotted curve in fig 5), w_f is calculated via the `w_f` function [here](https://github.com/afraser3/problems-with-parasites/blob/main/python/parasite_model.py). This relies upon the following functions in [kolmogorov_EVP.py](https://github.com/afraser3/problems-with-parasites/blob/main/python/kolmogorov_EVP.py):

- `Lmat` sets up the matrix we want the eigenvalues of
- `gamfromL` just passes L to `np.linalg.eig` to get the largest eigenvalue, "gamma" or "gam" in the code here, confusingly called "sigma" in the paper
- `gamma_over_k` and `gammax_kscan` runs the above but for a range of kz's to maximize growth rate over kz
- `gammax_minus_lambda` evaluates Eq 28 in my paper (but in the form LHS - RHS = 0) so that, in `w_f`, scipy.optimize.root_scalar can use it to find the w_f that satisfies Eq 28

Currently working on updating this so that the L matrix is the "with T and C" one that Sam and Pascale worked out (the better one). Otherwise the machinery will all be more or less the same.

afraser3:
I've implemented the "better" model in my repo (the "with T, C" one in FRG23, described in sec 5.2; as opposed to the worse model, "no T, C", described in sec 5.1). Enough changes that in several instances I elected to implement it as separate functions, rather than passing flags to the existing functions. But the structure remains the same. Below I try to list out the changes we need to note for implementation in MESA. It looks tedious, but very little of the structure/logic is changing, you just end up needing to redo a lot of things because you need slightly different parameters. If you want to take it for a spin yourself, [this script](https://github.com/afraser3/problems-with-parasites/blob/main/python/paper_plot_fig8.py) recreates fig 8 from the paper; sadly it takes tens of minutes at least to run.

[...]

PR #633, opened 2024-03-20
Thermohaline hackathon: speeding up FRG24 implementation

afraser3:
The FRG24 model for MHD thermohaline mixing, as currently implemented in the `thermohaline_hackathon` feature branch, is very slow (as pointed out by @evbauer in https://github.com/MESAHub/mesa/commit/d453d85031523d23698559e878bce5ea19de5001 in the context of #605). [I've TeXed up some notes](https://github.com/MESAHub/mesa/files/14658133/FRG24_MESA_speedup_summary.pdf) (TL;DR on page 2) on different ways to speed up the model---some with zero loss in accuracy, some with negligible loss in accuracy, and some with a loss in accuracy that is a factor of a few (but still more accurate than HG19 or the other models currently implemented). This draft PR covers these different speedups. Broadly speaking, they involve a change of basis that block-diagonalizes the matrix we're dealing with, asymptotic reductions of the equations, and reliance on HG19 where appropriate.

Some speedups this doesn't cover, but are worth exploring down the road if needed:

- The matrix whose eigenvalues we seek is very sparse, so ARPACK-NG's associated sparse methods are likely to be much faster than the currently-implemented LAPACK calls, especially in cases where the matrix size needs to be large. An example where I've tested this in the python implementation can be [seen here](https://github.com/afraser3/problems-with-parasites/blob/6684dc7f537938b62c35c25f3cab8311dcc7b627/python/kolmogorov_EVP.py#L814) (note that scipy.sparse is just a wrapper around ARPACK-NG, just as numpy.linalg is to LAPACK as I understand).
- Right now we're just sweeping over a bunch of k_z's when trying to maximize sigma vs k_z, and we're sort of shooting in the dark. Empirically, the peak always seems to fall between k_z/l_f = 10^-5 and 1, but that's a huge range! It should be possible to get at least an order-of-magnitude estimate of where the peak lies by using perturbative matrix methods (as in [this paper](https://doi.org/10.1080/03091929.2023.2268817)).
- In general, reducing the number of Fourier modes N or the density of k_z points scanned will directly speed things up at the cost of accuracy.
- Manipulating the matrix to have real-valued coefficients, which Rich has done in the past (Pascale has too, for a similar problem), would let us use DGEEV instead of ZGEEV. Presumably that's faster, but I don't know by how much.

### DIFF
--- a/turb/public/turb.f90
+++ b/turb/public/turb.f90
@@ -115,6 +115,7 @@ module turb
 
    subroutine calc_frg24_w(pr, tau, r0, hb, db, ks, n, w, withTC, safety, ierr, lamhat, l2hat)
      use parasite_model
+     use thermohaline, only: solve_hg19_eqn32
 
      real(dp), intent(in) :: pr, tau, r0, hb, db 
      real(dp), intent(in) :: ks(:)
@@ -130,13 +131,22 @@ module turb
      if (present(lamhat)) then
         ! lamhat and l2hat already calculated, so can pass as arguments
         if(withTC) then
-           w = wf_withTC(pr, tau, r0, hb, db, ks, n, .false., safety, lamhat, l2hat)
+           if (safety == 0) then
+             ! TODO: is it bad for this minimization to be occurring here AND in thermohaline.f90?
+             call solve_hg19_eqn32(hb, l2hat, lamhat, w, ierr)
+             w = MIN(w, wf_withTC(pr, tau, r0, hb, db, ks, n, .false., safety, lamhat, l2hat))
+           else
+             w = wf_withTC(pr, tau, r0, hb, db, ks, n, .false., safety, lamhat, l2hat)
+           end if
         else
            w = wf(pr, tau, r0, hb, db, ks, n, 0d0, 0, .false., .false., lamhat, l2hat)
         end if
      else
         ! lamhat and l2hat need to be calculated inside the wf routine
         if(withTC) then
+           if (safety == 0) then
+            stop "In turb.f90 calc_frg24_w, have not yet implemented safety=0 for case where lamhat, l2hat not already calculated"
+           end if
            w = wf_withTC(pr, tau, r0, hb, db, ks, n, .false., safety)
         else
            w = wf(pr, tau, r0, hb, db, ks, n, 0d0, 0, .false., .false.)
--- a/turb/public/turb.f90
+++ b/turb/public/turb.f90
@@ -144,7 +144,7 @@ module turb
      else
         ! lamhat and l2hat need to be calculated inside the wf routine
         if(withTC) then
-           if (safety == 0) then
+           if (safety == 0) then  ! TODO: fix this
             stop "In turb.f90 calc_frg24_w, have not yet implemented safety=0 for case where lamhat, l2hat not already calculated"
            end if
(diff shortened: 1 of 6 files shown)


afraser3:
As of this point, I've implemented the speedups that have zero error (change of basis / block diagonalization), and that have asymptotically small error (low Péclet number limit, which introduces errors of order tau, which is usually between 10^-3 to 10^-7). This _should_ be just under 10 times faster than it was before, but I'm not sure how best to test that.

Next step is to implement the low-Rm limit (Sec V of the PDF I linked above), and use HG19 as guardrails to keep the resulting error to acceptable levels.

afraser3:
I think this is working as intended. I've added a `safety` option to choose different combinations of the above speedups to employ:

- safety = 0: Low-tau limit, low-Rm limit, ditch "block a" (the block matrix with the ordinary KH mode), use HG19 away from large R0
- safety = 1: low-tau limit, uses low-Rm limit for "block b" (the block matrix with the weird KH mode), still calculates "block a"
- safety = 2: low-tau limit
- safety = 3 (max): use full model (I don't think this option will ever be needed since tau is always tiny)

[Here](https://github.com/MESAHub/mesa/files/14716336/turb_plotter_safety_test.zip) is a .zip with the resulting `turb/plotter` plots for each case, and you can see they all look very similar!

In the notes I TeXed above I claimed that for safety=0 the appropriate thing to do is to let D_C be the min between HG19's D_C and the one calculated with this method. What I've implemented here is I'm instead letting w_f be the min between HG19's w_f and the one calculated here. I'm not sure which is more appropriate off the top of my head.

afraser3:
For now, `safety=4` reverts back to the model as implemented before I began this PR, for sake of comparison with these other options. Note that `safety=4` and `safety=3` give identical results since no approximations have been made, it's just putting the matrix into a form that lets you get the eigenvalues faster. So `safety=4` should be removed before this PR is merged.

turb/public/turb.f90
```diff
@@ -130,14 +131,23 @@ subroutine calc_frg24_w(pr, tau, r0, hb, db, ks, n, w, withTC, ierr, lamhat, l2h
      if (present(lamhat)) then
         ! lamhat and l2hat already calculated, so can pass as arguments
         if(withTC) then
-           w = wf_withTC(pr, tau, r0, hb, db, ks, n, .false., lamhat, l2hat)
+           if (safety == 0) then
+             ! TODO: is it bad for this minimization to be occurring here AND in thermohaline.f90?
```
evbauer, line 135: I think this is fine. This `calc_frg24_w ` routine just exists to give a public interface for calling the velocities so we can use it in turb/plotter. Any actual stellar evolution goes through the `get_D_thermohaline` routine that is calling the stuff in `thermohaline.f90` independent of this, because right now mesa/star just wants a thermohaline composition diffusion coefficient.

evbauer: But this structure does end up causing a lot of duplicated code at this point. It would be better to have all that code live in a velocity routine in `private/thermohaline.f90`, and leave this public interface as just a wrapper to call that private routine.

afraser3: Rich and I are talking about this now and are putting a pin in this to resolve later.

Issue #634, opened 2024-03-20
segfault in pgbinary_orbit when secondary is a point mass


ryosuke-hirai:
**Describe the bug**
When using the "Orbit" panel in pgbinary, a segmentation fault occurs when the secondary is a point mass.

**To Reproduce**
Use the pgbinary inlist provided in evolve_both_stars in the test_suite.
Set 
evolve_both_stars=.false.

For some machines, the calculation stops with a segmentation fault when the pgbinary module writes into a png file. So far we've reproduced this bug in two linux machines and it doesn't seem to occur in Mac. The Xwindows output seems to be fine. It only crashes when outputting to png.

The problem seems to be that x2s_RL and y2s_RL are both not set when the secondary is a point mass in subroutine orbitpanel in pgbinary_orbit.f90. As a workaround, both arrays can be set to 0 when the secondary is a point mass. This seems to generate the same output without segfaulting.

Debraheem:
@matthiasfabry are you on an intel mac? I was unable to reproduce this on my m1 machine. 

-EbF

matthiasfabry:
Yes I am, on mesa-r23.05.1, sdk version x86_64-macos-23.10.1

Debraheem:
This has been solved, so I will close this issue now.

PR #638, opened 2024-03-28
fix/test pgbinary crashing in single star mode

Debraheem:
See [gh-issue-634](https://github.com/MESAHub/mesa/issues/634), this should fix a bug that crashed pgbinary when running in single-star mode with intel-based compilers. This did not affect my m1 machine.

The issue and fix were both found by @ryosuke-hirai and Sohan Ghodla so we should thank them for this.

To test this i made the star_plus_point_mass test_suite use the same pgstar controls as the evolve_both_stars test_suite to showcase pgbinary running in single star mode. I also changed the pgbinary flag to '.true.' so testing using the intel based compilers can catch if pgbinary produces a backtrace.

Once this is tested, and if the test_suite changes are allowed by @matthiasfabry and/or @orlox, I'll add this bug to the changelog and we can merge this.

### DIFF
--- a/binary/defaults/pgbinary.defaults
+++ b/binary/defaults/pgbinary.defaults
@@ -2971,11 +2971,11 @@
    Orbit_title = 'Orbit'
    Orbit_txt_scale_factor = 1.0
 
-      ! show the roche lobe sizes
+      ! show the star sizes
 
       ! ::
 
-   Orbit_show_RL = .true.
+   Orbit_show_stars = .true.
 
       ! file output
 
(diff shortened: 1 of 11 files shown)


orlox:
@matthiasfabry is the pgbinary expert, so I'll leave this to him

ryosuke-hirai:
I'm not sure if this is the relevant pull request to report this to, but there's also a small bug in `pgbinary_star.f90`. In lines 206 and 225, the format specifier should be `f0.2`, not `f3.2`. At the moment the mass of a point companion is always shown as `*** Mo` because the format cannot fit in the required digits.

matthiasfabry:
Thank you Ryosuke for this find! line 225 was fixed by @Debraheem, and I updated line 206.
I will continue making improvements and adding features to `pgbinary` over the coming months. I would love to hear if you have suggestions or requests.

Issue #607, opened 2023-11-29
predictive mixing segfault


Debraheem:
See the mailing-list thread: https://lists.mesastar.org/pipermail/mesa-users/2023-November/014916.html

An example (which crashes near timestep 19200, with a photo at timestep 19000 provided) can be downloaded here. 
[example_predictive_mixing_segfault.tar.gz](https://github.com/MESAHub/mesa/files/13505214/example_predictive_mixing_segfault.tar.gz)

where a 10Msun  r23.05.1 MESA model with the attached inlist segfaults at He core ignition (roughly when the convection appears in the core). 

The inlist controls adopted were:
    predictive_mix(1) = .true.
    predictive_zone_type(1) = 'burn_He'
    predictive_zone_loc(1) = 'shell'
    predictive_bdy_loc(1) = 'any'
    predictive_avoid_reversal(1) = 'he4'

                                      bad k for do1_mlt        3471        3470
../private/predictive_mix.f90, Line:  822, Message: non-zero op_err
ERROR STOP 1

Error termination. Backtrace:
#0  0x13af52147
...
It seems predictive mixing is the source of this issue, in which the subroutine is trying to access a zone which does not exist. where nz = 3470, the routine is trying to access zone 3471, interior to the center cell. 

This segfault can be avoided by turning predictive mixing off, or (hint?) setting 'make_gradr_sticky_in_solver_iters = .true.' Although, by reducing the mesh and time resolution of the model, I was also able to avoid this segfault.

I'm not familiar with nor do I use predictive mixing. Perhaps someone more familiar with it knows what is happening?

MESA version : r23.05.1, sdk version: mesasdk-aarch64-macos-23.7.2
computer: (arm) Macbook pro, M1 max, OS: Monterey 12.6

Debraheem:
The best solution to this is probably to just prevent users from using TDC with predictive mixing or present a warning to users.

PR #639, opened 2024-03-29
Prevent TDC with predictive mixing

Debraheem:
First stab at addressing gh-issue-[607](https://github.com/MESAHub/mesa/issues/607).

 I added an error that gets printed if predictive mixing is called while TDC is on. We could probably flag it earlier so you can't set both controls at the same time on start up? Open to any opinions here, should we disallow users from using these controls together, or just give them a big warning?

### DIFF
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -2301,6 +2301,15 @@
       ! Note that TDC does not respect the ``alt_scale_height`` option, and continues to use ``h = P / rho g``
       ! even if that flag is set.
 
+      ! We caution that combining different mixing models in a stellar evolution
+      ! calculation might lead to physically inconsistent solutions,
+      ! because the different models have been developed separately,
+      ! and their underlying assumptions might not be compatible with each other.
+      ! Examples include combining the newly implemented time-dependent local limit
+      ! convection model ``TDC`` with an overshooting model, or combining ``TDC``
+      ! with other models for chemical composition gradients
+      ! (predictive mixing or convective premixing), rotation, etc. (MESA VI)
+      
       ! ::
 
     MLT_option = 'TDC'
@@ -2773,6 +2782,9 @@
 
       ! Set to .true. to enable this set of parameters
 
+      ! ``predictive_mix`` is incompattible with ``convective premixing``
+      ! and mlt_option = ``TDC``, due to conflicting assumptions.
+      
       ! ::
 
     predictive_mix(1) = .false.
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -2781,9 +2781,6 @@
       ! ~~~~~~~~~~~~~~
 
       ! Set to .true. to enable this set of parameters
-
-      ! ``predictive_mix`` is incompattible with ``convective premixing``
-      ! and mlt_option = ``TDC``, due to conflicting assumptions.
       
       ! ::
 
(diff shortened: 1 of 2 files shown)


pmocz:
There is possibly a serious issue in [`do_predictive_mixing()`](https://github.com/MESAHub/mesa/blob/main/star/private/predictive_mix.f90#L209) in the stopping criterion for its [`search_loop`](https://github.com/MESAHub/mesa/blob/main/star/private/predictive_mix.f90#L363). 
Note that in this loop we are increasing `k_bot_mz = k_bot_mz + 1`  when `outward`=false, and the stopping criterion is in [line 491](https://github.com/MESAHub/mesa/blob/main/star/private/predictive_mix.f90#L491) when  `k_bot_mz == s%nz-1`.   But the search will fail if we start the search at `k_bot_mz =s%nz-1` because the value is incremented by 1 and then the stopping criterion cannot be hit. This is what is happening in the user issue.  What should the search_loop return in this case? set `k_bot_mz = s%nz-1` and exit the search loop?

Debraheem:
> What should the search_loop return in this case? set `k_bot_mz = s%nz-1` and exit the search loop?

that seems like it could work. Would it also make sense to change `k_bot_mz == s%nz-1)` on line 491 to `k_bot_mz >= s%nz-1)`?

pmocz:
> > What should the search_loop return in this case? set `k_bot_mz = s%nz-1` and exit the search loop?
> 
> That seems like it could work! Would it also make sense to change `k_bot_mz == s%nz-1` on line 491 to `k_bot_mz >= s%nz-1`?

It wouldn't quite fix it because [`eval_mixing_coefficients()`](https://github.com/MESAHub/mesa/blob/05b02d5b8af7f0028463591fb0091f4f4721fa01/star/private/predictive_mix.f90#L380C13-L380C31) operates on the index first, and goes out-of-bounds.

It seems to me that the intention of this function is to return a search result within [1, s%nz-1]. Would @evbauer know more about `do_predictive_mixing()`?

It seems to me there is also an issue at the other end of the search. If `k_top_mz` is initially 1 (and `outward`=true) then the `search_loop` in `do_predictive_mixing()` would decrease the index to 0 and not be able to exit at the stopping criterion in [line 490](https://github.com/MESAHub/mesa/blob/05b02d5b8af7f0028463591fb0091f4f4721fa01/star/private/predictive_mix.f90#L490C12-L490C47)

fxt44:
my recollection from mesa VI is TDC and predictive mixing should not be used together, although i can't seem to find the email thread on this at the moment.

rich townsend, as the primary author of predictive mixing, should weigh in on this issue as well the do loop limits.

pmocz:
Sounds like we should make TDC and predictive mixing a run mode that is not allowed, as @Debraheem has nicely done!

@rhdtownsend what are your thoughts on how to fix the out-of-bounds issue in the `search_loop` in `do_predictice_mixing()`?

wmwolf:
From a conceptual standpoint, is there a reason that predictive mixing and TDC are mutually exclusive? I use both in novae since the time steps are necessarily short during the ramp-up of the convective runaway, so convective pre-mixing introduces unnaturally fast mixing in boundary regions. Predictive mixing does not instantaneously mix material, but it helps prevent the convective shell from shattering into tons of tiny sub-shells.

I get that there is a numerical issue here, but if the two are not logically inconsistent with each other, I'd rather put up some guardrails rather than forbid the combination altogether. I agree; it would be good to get @rhdtownsend's opinion.

Debraheem:
Nice tweak @pmocz! @rhdtownsend can you give it a quick look over before we merge?

PR #642, opened 2024-04-05
Ignore `max_allowed_nz` if ≤ 0

warrickball:
As [suggested](https://github.com/MESAHub/mesa/issues/621#issuecomment-1959094641) in issue #621, this PR allows users to ignore `max_allowed_nz` by setting it zero or less, in effect allowing the mesh to grow arbitrarily large, and in keeping with many other controls that are ignored when negative. Consensus seems to be that this should *not* be the default, although we may want to increase it from the current default of 8000.

I'm not sure if we want to strictly compare to -1 or perhaps advise users to use that, which would leave us the option of using particular values of `max_allowed_nz` to control meshing behaviour. I think generally we'd want more description controls anyway (e.g. set a string, not a specific integer) but I thought I'd throw the idea out there before users start using all sorts of negative values in their inlists.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -18,6 +18,8 @@ Backwards-incompatible changes
 New Features
 ------------
 
+``max_allowed_nz`` is now ignored if the value is less than or equal to zero.
+
 
 .. _Bug Fixes main:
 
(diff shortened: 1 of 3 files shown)


mjoyceGR:
will check this today/tomorrow, sorry for the delay!

mjoyceGR:
added documentation of this `controls.defaults` but forgot [ci skip] initially. I just cancelled the test workflow manually and pushed another minor modification (with [ci skip]) sooo I'm going to assume this is fine now. Thanks @warrickball for introducing this feature

PR #644, opened 2024-05-13
Add lanl oplib type1 opacity tables, new abundances, new lowT tables

Debraheem:
This pr exists to introduce new aag21, and magg22 abundances into MESA, and associated lowT Fa05 opacities. It also introduces the four 1194 table OPLIB opacities tables computed for gs98, agss09, aag21, and mb22 mixtures. 

This pr still needs some further documentation and changelog before being merged. We have tested the independently in a separate branch, as they are not the MESA default, however it could be worth running one more test to confirm.


warrickball:
Not sure if you've already noticed but MESA [won't compile](https://testhub.mesastar.org/EbF%2Fadd_lanl_OPLIB_type1_opacity_tables/commits/c41ae3c). Error message for `cannon_conroy_intel` is the same I just got when trying to submit a test run:

```
************************************************

/n/holyscratch01/conroy_lab/ebauer/mesa-git-tests/tmp.MZeHhPO8sg/kap
installing mesa kap data.

create kap_data
tar xfJ kap_data.tar.xz
tar: This does not look like a tar archive
tar: Skipping to next header
tar: Exiting with failure status due to previous errors
```

Debraheem:
Yeah I guess the file was corrupted somehow. I know this might rub some the wrong way since the current opacity files are old, but I am just going to rebuild it completely this time with the current sdk version on my arm computer. I understand that the results will change, but they haven't been rebuilt since 2017, and me peacemealing the file to only add new data is what caused this issue in the first place.

fxt44:
> rebuild it completely this time

this will be fun on the test suite.

Debraheem:
testing is here: https://testhub.mesastar.org/EbF%2Fadd_lanl_OPLIB_type1_opacity_tables/commits/ad20f28

docs/source/kap/overview.rst
```diff
@@ -57,10 +57,27 @@ High temperature :math:`(T \gtrsim 10^4\,\rm K)`
...
+:math:`0.0\leq Z \leq 0.2`. The set of tables to be used are
 selected by the option :ref:`kap/defaults:kap_file_prefix`.
 
+A direct comparison between the Type 1 format of OPAL/OP tables and 
+the OPLIB tables are shown in the figure below
+taken from Figure 1 in Farag et al. 2024. Further comparisons
+between OP/OPAL/OPLIB can be found in Farag et al. 2024.
+
+
+.. figure:: type1_format.pdf
```
evbauer, line 73: Figures will display better inline as pngs rather than pdfs.

docs/source/kap/overview.rst
```diff
@@ -57,10 +57,27 @@ High temperature :math:`(T \gtrsim 10^4\,\rm K)`
...
+   :alt: EOS blends in solar-like compositions
```
evbauer, line 74: This label should refer to Opacity, not EOS, right?

evbauer:
This looks concerning: https://testhub.mesastar.org/EbF%2Fadd_lanl_OPLIB_type1_opacity_tables/commits/ad20f28/test_cases/star/rsp_Type_II_Cepheid

evbauer:
Why do we end up having different results after regenerating the old kap tables? Is it just because the kap preprocessor isn't bit-for-bit? I spent a while looking at differences between the data files between the last release and this branch for `a09` and `gs98` mixtures, and it looks mostly like few percent level differences possibly driven by different rounding and the fact that our old style kap tables only store 3 decimal places. So it's not obvious that there's anything terribly wrong with these newly generated tables, but it's uncomfortable that this drove some test suite failures.

Weirdly, I also noticed that the previous Type II data files sometimes contained duplicate and/or missing tables when compared to the ones Eb just generated. So maybe these newly generated ones are a strict improvement!

evbauer:
Thanks! I think this mostly addresses my concerns. Note that `radiative_levitation` and `rsp_save_and_load_file` are showing the usual inconsistency between SDK and ifort tests, so there's no regression on those. That leaves only `ppisn` and `rsp_Type_II_Cepheid`.

`ppisn` is a fairly difficult test that might just need more help converging?

The problem with `rsp_Type_II_Cepheid` is concerning in that we're legitimately getting different results on different platforms. This looks like it could be revealing a bit-for-bit problem with RSP that could have been lurking but never revealed until the opacity tweaks caused the model to do something slightly different?

evbauer:
Oh, sorry, RSP is known not to be bit-for-bit across platforms, even on just SDK builds. We already skip the restart and checksum checks for RSP test cases. Looking at a diff of output between successful and unsuccessful runs shows differences in the initial model before even taking the first step:

```
            P(days)         growth                                                                  P(days)         growth
  0       0.16449E+01     0.16484E+00                                                     0       0.16449E+01     0.16484E+00
  1       0.11483E+01     0.57032E-04                                               |     1       0.11483E+01     0.57056E-04
  2       0.83165E+00    -0.53670E+00                                                     2       0.83165E+00    -0.53670E+00
                                                     nz         200                                                                          nz         200
                                                  T(nz)    2.0004613628876873D+06   |                                                     T(nz)    2.0004613628791873D+06
                                          L_center/Lsun    1.3600000000000000D+02                                                 L_center/Lsun    1.3600000000000000D+02
                                          R_center/Rsun    7.2457680346835551D-01   |                                             R_center/Rsun    7.2457680347158349D-01
                                          M_center/Msun    5.4152292487481979D-01                                                 M_center/Msun    5.4152292487481979D-01
                                              L(1)/Lsun    1.3600000000000000D+02                                                     L(1)/Lsun    1.3600000000000000D+02
                                              R(1)/Rsun    9.5245487199735681D+00   |                                                 R(1)/Rsun    9.5245487200400945D+00
                                              M(1)/Msun    5.5000000000000004D-01                                                     M(1)/Msun    5.5000000000000004D-01
                                               v(1)/1d5    4.5000000000000000D+00   |                                                  v(1)/1d5    4.4999999999999991D+00
                                             tau_factor    1.5000000000000000D-03                                                    tau_factor    1.5000000000000000D-03
                                               tau_base    6.6666666666666663D-01                                                      tau_base    6.6666666666666663D-01

                                        set_initial_age    0.0000000000000000D+00                                               set_initial_age    0.0000000000000000D+00
                                                set_age    0.0000000000000000D+00                                                       set_age    0.0000000000000000D+00
```

It just happens to be the case that Cannon's version fails while the others succeed. While ideally we would fix this inconsistency someday, that might be beyond the scope of this particular PR. Perhaps we should loosen the tolerance of this test.

Debraheem:
On ppisn, it is just sensitive. I have some changes coming in the future that will make this test_suite more robust, but for now, just removing the meshing from the surface zones like I did in this commit should be enough, see https://github.com/MESAHub/mesa/commit/e30386f1452f55a16e0415517cb2592474d568b4.

evbauer:
Great. Let's go ahead and make that change to `ppisn` on this branch as well then. I think testing should be good to go at that point.

rsmolec:
RSP results may indeed by sensitive to opacities, however I would be surprised with 1-1.5% differences if the changes are just to some minor numerics and not real changes in opacity data. Of courses changes like using linear vs. spline interpolation may lead to significant differences and if it is the case 1-1.5% is ok.

It may also be important how tests are constructed and what is actually compared. Test cases like rsp_Cepheid do not compare linear periods. They also do not compare non-linear periods - we would need to run hundreds of models to get those. Instead, the model is initialised, computes 10 cycles, and the period (peak-to-peak) is compared then. So it is in fact most sensitive to the eigenfunction which we use to initialize nonlinear calculation, which is affected by opacity. For such cases I expect much larger sensitivity. I'm not sure whether any of the tests compares just the linear periods, the ones that are computed based on static model only and are stored in LINA_period_growth.data.

Debraheem:
The opacity data itself has changed when I regenerated it using the pre-processor. To what degree, I'm not fully aware but these difference manifest at the three decimal point level in the kap tables.

@evbauer are we sure we should loosen the tolerance? Is it better to just change the target period for the test case but keep the tolerance tight, so we can keep an eye on these test cases in the future?

warrickball:
Did someone say "bit-for-bit"? :grin: 

> While ideally we would fix this inconsistency someday, that might be beyond the scope of this particular PR.

It's long surprised me that RSP has been quite tolerant to the sort of bit-for-bit changes we've implemented elsewhere. The moment we have cause to wade in and a way of knowing if we're making progress, I'm happy to give the module a pass. I fully agree that this PR isn't the place for it but if we know we have a way of triggering different pass/fail on different platforms, it's an opportunity to fix any remaining issues.

evbauer:
> @evbauer are we sure we should loosen the tolerance? Is it better to just change the target period for the test case but keep the tolerance tight, so we can keep an eye on these test cases in the future?

I checked Cannon vs ScienceFurnace for the test [here](https://testhub.mesastar.org/EbF%2Fadd_lanl_OPLIB_type1_opacity_tables/commits/ad20f28/test_cases/star/rsp_Type_II_Cepheid), and the tested period differs by about 1% _for the same commit and same opacity data_ across those two different machines. So we either have to loosen the tolerance or fix all bit-for-bit issues in RSP.

Debraheem:
Wow, thanks for checking. I'm fine with keeping the loose tolerances and i think it this is ready for a merge now?

evbauer (approved):

rsmolec:
@evbauer can you run this model with RSP_do_check_omega = .false. in &controls. One should not do that, but I'm curious what will happen and wonder whether use of rand() may be a source of bit-for-bit issue. We use rand() to avoid some problems for restart of convection in model cells that were radiative. At one point Bill tried to get rid of that (RSP_do_check_omega = .false.) but it leads to some serious issues (it was a default in one of the releases, https://docs.mesastar.org/en/release-r24.03.1/known_bugs.html#rsp). Otherwise what kind of 'bad coding practices' may lead to bit-for-bit inconsistency?

warrickball:
> Otherwise what kind of 'bad coding practices' may lead to bit-for-bit inconsistency?

First, I'll point out that the bit-for-bit stuff is generally not "bad" coding practice in a typical sense. It just turns out that we have to coerce the compiler into always implementing the same thing for certain operations. One of these is `**X`, which we replace with `powX(...)` to make sure that the multiplications happen in the same order. I think we found that `Z**2` always seems to come out as `Z*Z` but higher powers start becoming unreliable and there are e.g. `**4` in `rsp_lina.f90`.

Another thing to catch is redefining constants that are already in the `const` module. This gets as goofy as replacing `/3d0` with `*one_third`, which seems ridiculous but has been what it takes to get bit-for-bit consistency. E.g. in `rsp_def.f90`, `P4=4.d0*PI` is defined but we could use `pi4` from `const`.

Rob Farmer wrote a linter (in `linters/mesa-linter.py`) that will report many potential issues. In `$MESA_DIR`, try `python linters/mesa-linter.py star/private/rsp_*.f90`.

Because I run the `ifort` build, I'm usually the one who's investigated these and I'm quite happy to have a pass through RSP if you're happy to review any pull request I make. The nuisance with bit-for-bit stuff is that we're never really sure which bits actually matter, so sometimes we change code for nothing. But at least we'll eliminate the *possibility* that it's an issue.

warrickball:
>  E.g. in `rsp_def.f90`, `P4=4.d0*PI` is defined but we could use `pi4` from `const`.

Also, in RSP's defence, I just glanced at the source to find an example of something else and found the same thing happening with `4d0*pi` being used instead of `pi4`, so it's not just RSP.

Debraheem:
> Because I run the ifort build, I'm usually the one who's investigated these and I'm quite happy to have a pass through RSP if you're happy to review any pull request I make. 

Awesome, thank you @warrickball, perhaps we can open a separate issue for discussing and solving the bit for bit inconsistency in RSP? I will merge this commit now.

evbauer:
> can you run this model with RSP_do_check_omega = .false. in &controls

@rsmolec it looks like `rsp_Type_II_Cepheid` already has `RSP_do_check_omega = .false.` in its current implementation.

I went ahead and generated a branch to test this some more, tightening the tolerance back up so that Cannon should report a failure again, but Cannon is offline for annual maintenance this week, so we might have to wait a while to reproduce the failure. https://testhub.mesastar.org/RSP_testing/commits

PR #646, opened 2024-05-13
Automatic differentiation for cubic interpolation across type-1 and type-2 kap tables

Debraheem:
This pr introduces automatic differentiation into the interpolating functions for cubic splines. This allows kap to call the derivative of the interpolating function for rho-T derivatives instead of interpolating the derivatives themselves. This allows for a significant improvement in the numerical derivatives produced by kap when using cubic interpolation.

While I personally recommend a shift toward cubic interpolation in kap as a future MESA default, some more verification is needed before then. For now, we can introduce this capability and recommend it to users to test so we can get feedback. 

We have successfully tested the cubic interpolation on the test_suite as MESA default.  
Everything looks relatively good outside of three test cases, two output_test txt file mismatches, and one with RSP failing to build the model. Otherwise, everything looks fine.  A new commit should test one more time right now to confirm, see https://testhub.mesastar.org/test_kap_cubic_splines/commits/2c0dcec. 

At the moment, this pr is missing documentation, an updated changelog, and needs some level of cleanup to prevent auto_diff conflicts before merging. I will likely update the solar calibration test_suite to include cubic interpolation as a default, since this test suite needs updated and cubic interpolation appears necessary for producing good (comparable to literature) solar models.


wmwolf:
It'd be interesting to see if this cuts down on solver iterations, retries, and/or timesteps. Is there a clear before/after pair of commits that I could use to pull some data from the testhub for such an analysis? One could do this with a lot of clicking, but I think it might be easier if I just dive into the database and grab the data programmatically.

Debraheem:
Here is the latest commit for testing cubic splines 'test_kap_cubic_splines' with auto_diff (and a somewhat recent version of main merged in): https://testhub.mesastar.org/test_kap_cubic_splines/commits/2c0dcec

I will remerge the latest version of main into this branch and re-run the test, and then we can compare directly to the testing on the main branch?

Debraheem:
@wmwolf you can compare https://testhub.mesastar.org/test_kap_cubic_splines/commits/834dee5 to a full test on main.

evbauer:
Heads up, I'm going to simplify this PR by getting rid of most of the changes to `kap/plotter`. Those were useful for developing an understanding of what's going on, but I don't think all of that should live there when it gets merged into `main`.

evbauer:
Ok, I think I've taken care of the cleanup part. Now I just think we need some documentation and a changelog entry?

Debraheem:
This pull request looks ready to merge. If there are no further comments, I will merge and begin the release process.

Issue #645, opened 2024-05-13
no burning at lowT, lower the default jina rate floor.


Debraheem:
See : https://lists.mesastar.org/pipermail/mesa-users/2024-May/015169.html

See also https://docs.mesastar.org/en/release-r24.03.1/reference/star_job.html#jina-reaclib-min-t9

"set jina reaclib rates to zero for T9 <= this. if this control is <= 0,
then use the standard default from rates. need <= 3d-3 for pre-ms li7
burning if change this, must remove old cached rates from
data/rates_data/cache"

We currently have `reaclib_min_T9 = 1d-2` (i.e. no reactions below 10 MK) as a MESA default, which misses be9 burning as well as li7 burning on the pre-ms, so these rates are effectively set to 0. This seems like an issue. Why don't we lower this to 2d-5 or something similar to see if it breaks anything? I think we should find a way to lower the floor if possible or communicate when it should and shouldn't be lowered more clearly to users. I was not aware of this until Masanobu raised it and it directly influences pre-ms modeling, but it could have unforeseen consequence we are not yet aware of.

fxt44:
let's do the experiment of lowering to at least 1d-3.

evbauer:
Does anybody what was the original motivation for having this control in the first place?

fxt44:
On May 16, 2015, at 3:49 PM, Bill Paxton <paxton@kitp.ucsb.edu> wrote:

I'm leaving this one to Frank to fix someday.
Meantime, I'm just providing a not-very-satisfactory workaround for the pre-ms and ms case.

As you can see from this lovely figure, we need to get down to 2e6 for Li and 4e5 for Deuterium.
But if we drop the lower logT limit for reaclib below 3e6 we run into very bad convergence problems during post ms burning.

Perhaps there's a nice general solution, but I don't see it now.

So the ad hoc nasty solution is to provide a control that let's you change the lower logT bound from the inlist.

when you change the bound, you need to clear out the rates cache (delete all from data/rates_data/cache).

I'm leaving the default unchanged (1e7).    For pre-ms and ms evolution where you want to use a large net and also included Li and Deut burning, you'll have to explicitly reduce the limit by setting the new control 'jina_reaclib_min_T9' in &star_jobs.    Note that the value is T9.  So you'll want something like jina_reaclib_min_T9 4d-4 to get down to logT ~ 5.6.

This new control is in 7574.  I'm running the test suite at the moment.

-Bill

fxt44:
[...]

PR #647, opened 2024-05-13
logarithmic interpolation in jina rates

Debraheem:
See: https://github.com/MESAHub/mesa/issues/645

We currently have the default reaclib_min_T9 = 1d-2 (i.e. no reactions below 10 MK) as a MESA default, I would like to bring this down so we are not missing key reactions like li7 and be9 burning.
 
I have tested with reaclib_min_T9 = 1d-4 (i.e. no reactions below 1d5 K) with no issues, see https://testhub.mesastar.org/EbF%2Flower_jina_rate_floor/commits/8555db8. The testing looks good. I would appreciate any feedback or further testing from any members of the community on this.

I will modify the changelog once we come to a conclusion on this.


fxt44:
i will test after the current test suite run finishes.

rates/private/rates_support.f90
```diff
@@ -282,6 +280,20 @@ subroutine do_make_rate_tables( &
...
+                      if (rattab(j, i) > 0.0_dp) then  ! Only take log of positive values
+                          rate_logR = log10(rattab(j, i))
+                      else if (rattab(j, i) == 0.0_dp) then
+                          rate_logR = -999d0
+                      else
+                          ! warning for rates set to -1
+                          ! like rni56ec_to_co56,rco56ec_to_fe56
+                          !write(*, '(a,i4,2x,a)') 'Warning: non-positive value for rate in ',  &
+                          !   j, trim(reaction_Name(reaction_id(j)))
+                          rate_logR = rattab(j, i) ! don't convert to log space.
```
pmocz, line 294: I could be misunderstanding what is going on here. Are you saying the data tables (in linear space) have some -1 entries? Then setting rate_logR = -1 could be a problem, no?

PR #653, opened 2024-06-06
fix links to SDK website

mathren:
Stumbled upon a few rotten links in the documentation FAQ and fixed them.


mathren:
Added a commit that fix the tutorial for extending MESA. The examples where still using `rsol` (now renamed to `rsun`), `safe_log10_cr` (now just `log10`) and `s%P` (now `s%Peos`).

Thanks to Christian Burt for finding these little inconsistencies

Issue #661, opened 2024-06-19
shmesa GNU sed not compatible with macOS default BSD sed


gautam-404:
**Describe the bug**
All `shmesa` functions that use the `sed` utility do not work on macOS. Mac default is the FreeBSD `sed` which functions slightly differently.  Functions affected by this are `shmesa_extras`, `shmesa_defaults` and `shmesa_change`.

Since `perl` is available on most systems, it can potentially replace the use of `sed`.

**To Reproduce**
1. On MacOS try the `shmesa extras` command
2. The error:
```
sed: illegal option -- z
usage: sed script [-Ealnru] [-i extension] [file ...]
        sed [-Ealnu] [-i extension] [-e script] ... [-f script_file] ... [file ...]
```

A trial implementation using `perl` that works on macOS:
```
shmesa_extras () {
    if shmesa_check_h_flag "$@"; then
        echo "Usage: shmesa extras [src/run_star_extras.f90]"
        echo "replaces run_star_extras.f90 with the full template"
        return 0
    fi

    local rse='src/run_star_extras.f90'
    if [[ ! -z $1 ]]; then
        rse=$1
    fi

    local contents=$(cat "$MESA_DIR/include/standard_run_star_extras.inc")
    if [[ -z $contents ]]; then
        echo "Cannot find $MESA_DIR/include/standard_run_star_extras.inc"
        return 1
    fi
    contents=$(perl -pe 's/:/\\:/g; s/\n/\\n/g' "$MESA_DIR/include/standard_run_star_extras.inc")

    backup_copy "$rse" "$rse".bak
    perl -i -pe "s|      include 'standard_run_star_extras.inc'|$contents|g" "$rse"
}
```
```
shmesa_change () {
    if shmesa_check_h_flag "$@"; then
        echo "Usage: shmesa change inlist parameter value [parameter value [parameter value]]"
        echo "Modifies one or more parameters in the supplied inlist."
        echo "Uncomments the parameter if it's commented out."
        echo "Creates a backup of the inlist in the corresponding .bak file."
        echo ""
        echo "Examples:"
        echo "  shmesa change inlist_project initial_mass 1.3"
        echo "  shmesa change inlist_project log_directory 'LOGS_MS'"
        echo "  shmesa change inlist_project do_element_diffusion .true."
        echo "  shmesa change inlist_project initial_mass 1.3 do_element_diffusion .true."
        return 0
    fi
    
    if [[ -z $1 || -z $2 || -z $3 ]]; then
[...]

PR #663, opened 2024-06-20
shmesa fix for issue #661

gautam-404:
fix for issue #661

### DIFF
--- a/scripts/shmesa/shmesa
+++ b/scripts/shmesa/shmesa
@@ -113,28 +113,26 @@ shmesa_change () {
     
     # Create a backup of the inlist before making any changes
     backup_copy "$filename" "${filename}.bak"
-    
-    local ESCAPE="s#[^^]#[&]#g; s#\^#\\^#g" # sed escape string
-    # iterate through parameter,value pairs in the supplied input 
+
     while [[ -n $1 && -n $2 ]]; do
         local param=$1
         local newval=$2
         shift 2
         
-        local escapedParam=$(sed "$ESCAPE" <<< "$param")
-        local search="^\s*\!*\s*$escapedParam\s*=.+$" 
+        local search="^\s*\!*\s*\b${param}\b\s*=.*$"
         local replace="    $param = $newval"
         
         # Check if the parameter is present in the inlist
-        if ! grep -q "$param" "$filename"; then
+        if ! grep -qw "$param" "$filename"; then
             echo "Error: Parameter '$param' not found in the inlist '$filename'."
             backup_restore "$filename"
             return 1
         fi
         
         # Update its value
-        sed -r -i -e "s#$search#$replace#g" "$filename"
+        perl -i -pe "s/$search/$replace/" "$filename"
     done
+
 }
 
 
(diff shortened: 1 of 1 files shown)


earlbellinger:
You are great! Thank you for working on this. 

Unfortunately there seems to be a bug. When I call `shmesa test` I get:

```
...
>>> test: shmesa change

SHMESA> DEBUG: Calling shmesa change with arguments: inlist_project pgstar_flag .false.
SHMESA> DEBUG: SHMESA> BACKING UP: inlist_project inlist_project.bak
SHMESA> DEBUG: Calling shmesa change with arguments: inlist_project initial_mass 1.2 Z 0.01 Zbase 0.01
SHMESA> DEBUG: SHMESA> BACKING UP: inlist_project inlist_project.bak
Error: Parameter 'Z' not found in the inlist 'inlist_project'.
SHMESA> DEBUG: SHMESA> RESTORING: inlist_project
```

Does `shmesa test` work on your machine?

earlbellinger:
Oh actually this might be a problem with the tester rather than with your fix. Before it tried to change an inlist parameter `Z` which doesn't exist. Normal `shmesa` just doesn't make a change whereas this fix causes it to not go through. It's fine with it throwing an error though. In that case we should just change the test to have the proper `initial_z` rather than just `Z` then.

gautam-404:
Thanks for pointing that out. This should work now. 

I am wondering if it is worth exploring the possibility to include functions like `shmesa_set`, `shmesa_rm`, `shmesa_get` to add, remove or echo inlist parameters. 
I've previously implemented this with Python. I am not sure if this is required or will be useful. Thoughts?

earlbellinger:
Yes, please do! I think those would be really great additions. The only gotcha I can think of is taking care of getting the right namelist group.

wmwolf:
The problem of identifying nameless groups is not trivial but it's doable. You can look at the source for [MesaScript](https://github.com/wmwolf/MesaScript) to see how it's done there (in MesaScript, you don't need to specify namelists; it handles sorting them for you). It might be hard to reproduce in `bash` alone, though. If I recall correctly, there is exactly one control that lives in multiple namelist groups. Maybe its `logs_directory` that is shared between `binary_controls` and `controls`, but I could be mistaken.

gautam-404:
I have achieved the desired results [here](https://github.com/gautam-404/MESA-PORT). But I guess Python does make things easier. Doing the same with shell script is going to be way more challenging.

earlbellinger:
Yes having dependencies is tough because then we have to support them - even `sed` isn't as cross-platform as I had thought 😂 

But I just realized we can do all of these in a pretty straightforward way: 

- `shmesa add` can be done if we just specify the group name and then add the parameter right after it. Then we can grep for e.g. `&controls` and put it as the first param in there. Syntax can be `shmesa add inlist group paramter value`, e.g., `shmesa add inlist controls time_delta_coeff 1` 

`shmesa get` can be a simple grep 

- `shmesa remove` could just add a `!` in front of the parameter. It would comment out all such instances which is bad if the same parameter name appears in multiple groups, but as Bill said there's only one example of that, so I think we could live with it... Alternatively it could also require specifying the group name, then it's doable with regex.

PR #669, opened 2024-06-26
adding parallel do simd

pmocz:
I am using Caliper (https://software.llnl.gov/Caliper/) to profile parts of the code and understand why scaling typically plateaus beyond 12 cores. Some of it has to do with overhead + small simulation size. But I am detecting some unparallelized code

Here, I've changed `!$omp simd` -->  `!$omp parallel do simd`  to ensure loops are not only vectorized but multi-threaded. The scaling is improved and the 2024 MESA summer school day 1 minilab1.1_5M and a couple seconds are shaved off runtime.

image

### DIFF
--- a/star/private/star_bcyclic.f90
+++ b/star/private/star_bcyclic.f90
@@ -75,12 +75,13 @@
 
          ierr = 0
          neq = nvar*nz
-         !$omp simd
+         !$OMP PARALLEL DO SIMD
          do i = 1,nvar*neq
             lblkF1(i) = lblk1(i)
             dblkF1(i) = dblk1(i)
             ublkF1(i) = ublk1(i)
          end do
+         !$OMP END PARALLEL DO SIMD
 
          if (dbg) write(*,*) 'start bcyclic_factor'
 
--- a/star/private/star_solver.f90
+++ b/star/private/star_solver.f90
@@ -1016,11 +1016,11 @@
                call start_time(s, time0, total_time)
             end if
             
-            !$OMP SIMD
+            !$OMP PARALLEL DO SIMD
             do i=1,neq
                b1(i) = -equ1(i)
             end do
-            !$OMP END SIMD
+            !$OMP END PARALLEL DO SIMD
             
             if (s% use_DGESVX_in_bcyclic) then
                !$OMP PARALLEL DO SIMD
--- a/star/private/star_solver.f90
+++ b/star/private/star_solver.f90
@@ -1016,30 +1016,33 @@
                call start_time(s, time0, total_time)
             end if
             
-            !$omp simd
+            !$OMP PARALLEL DO SIMD
             do i=1,neq
                b1(i) = -equ1(i)
             end do
+            !$OMP END PARALLEL DO SIMD
             
             if (s% use_DGESVX_in_bcyclic) then
-               !$omp simd
+               !$OMP PARALLEL DO SIMD
                do i = 1, nvar*nvar*nz
                   save_ublk1(i) = ublk1(i)
                   save_dblk1(i) = dblk1(i)
                   save_lblk1(i) = lblk1(i)
                end do
+               !$OMP END PARALLEL DO SIMD
             end if
             
             call factor_mtx(ierr)
             if (ierr == 0) call solve_mtx(ierr)
             
             if (s% use_DGESVX_in_bcyclic) then
-               !$omp simd
+               !$OMP PARALLEL DO SIMD
                do i = 1, nvar*nvar*nz
                   ublk1(i) = save_ublk1(i)
                   dblk1(i) = save_dblk1(i)
                   lblk1(i) = save_lblk1(i)
                end do
+               !$OMP END PARALLEL DO SIMD
             end if
 
             if (s% doing_timing) then
--- a/star/private/star_solver.f90
+++ b/star/private/star_solver.f90
@@ -1016,11 +1016,11 @@
                call start_time(s, time0, total_time)
             end if
             
-            !$OMP PARALLEL DO SIMD
+            !$OMP SIMD
             do i=1,neq
                b1(i) = -equ1(i)
             end do
-            !$OMP END PARALLEL DO SIMD
+            !$OMP END SIMD
             
             if (s% use_DGESVX_in_bcyclic) then
                !$OMP PARALLEL DO SIMD


earlbellinger:
Excellent! On the testhub it looks like there's an RSP failure, but I can't immediately tell if it's related 
https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/commits/c5b2818

warrickball:
That's the commit, shared with `main`, where this branch starts. You really want [this commit](https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/commits/54709ce).

The failure is a long-running mixed result because it's failed on `bluebear_ifort` for agees.

warrickball:
Results generally look [good](https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/commits/54709ce). I've been comparing some results for `bluebear_ifort`, since this commit and the faster two of the three commits on the [previous commit](https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/commits/c5b2818) should be happening under very similar conditions (in particular, the same subset of CPU model).

@evbauer, how consistent should `cannon_conroy_intel` be? It seems to quite often be *slower* after these changes. e.g. for `12M_pre_ms_to_core_collapse` it took [~18m after these changes](https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/commits/54709ce/test_cases/star/12M_pre_ms_to_core_collapse) vs [~17m before](https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/commits/c5b2818/test_cases/star/12M_pre_ms_to_core_collapse). 

It seems quite clear that the code changes are an improvement. I'm not sure offhand if we need to specify which are the private variables in the `OMP` directive but I assume not in such a simple case.

warrickball:
To be fair, on `bluebear_ifort`, `wd_nova_burst` was a bit slower [after the changes (~6.6m)](https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/commits/54709ce/test_cases/star/wd_nova_burst) than [before (~5.5m)](https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/commits/c5b2818/test_cases/star/wd_nova_burst). It might just be short runs where the variance on the systems doesn't make the benefits as clear but I don't doubt this is an improvement.

warrickball (approved):

pmocz:
Am I reading this correctly? There is quite a large increase in the timing after my changes on `16M_conv_premix` on `cannon_conroy_intel` : [after](https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/commits/54709ce/test_cases/star/16M_conv_premix)  and [before](https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/commits/c5b2818/test_cases/star/16M_conv_premix) . This looks worrisome to me, but I am unsure why there is a large increase in timing for this test

pmocz:
> Am I reading this correctly? There is quite a large increase in the timing after my changes on `16M_conv_premix` on `cannon_conroy_intel` : [after](https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/commits/54709ce/test_cases/star/16M_conv_premix) and [before](https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/commits/c5b2818/test_cases/star/16M_conv_premix) . This looks worrisome to me, but I am unsure why there is a large increase in timing for this test

Although, I do see a large increase/blip in the timing of this test at an earlier time as well: https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/commits/16b6401/test_cases/star/16M_conv_premix

evbauer:
I just had cannon rerun `16M_conv_premix` on the [after](https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/commits/54709ce/test_cases/star/16M_conv_premix) commit, and this time the runtime looks totally normal.

I will say that Cannon occasionally has filesystem issues that cause very slow reads/writes. I don't know if that's what happened in this case, but at least I think we can rule out that your commit was to blame for the slowdown.

pmocz:
I reran more testing as well, and results are consistent with small speedup. To view history of each test problem, one can take a look at e.g.  https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/test_cases/star/1M_pre_ms_to_wd?utf8=%E2%9C%93&history_type=show_instances&start_date=2024-05-28&end_date=2024-06-27&computers=bluebear_ifort&status=&commit=Search

I believe this PR will be good to merge.

Debraheem (approved):
I think it would be nice to do one more empty "[ci optional]" commit to test the speed up on the optional inlists.I'm particularly curious how the longer test suites like pms_to_cc fare.

pmocz:
> I think it would be nice to do one more empty "[ci optional]" commit to test the speed up on the optional inlists.I'm particularly curious how the longer test suites like pms_to_cc fare.

Great idea! I'm running it now. Which test is it exactly? I'm seeing some nice speedup on some of the longer tests, e.g.: https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/test_cases/star/make_pre_ccsn_13bvn

Debraheem:
Nice, https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/test_cases/star/20M_pre_ms_to_core_collapse, am i right in saying that for these two cases the run time was cut in half?

pmocz:
> Nice, https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/test_cases/star/20M_pre_ms_to_core_collapse, am i right in saying that for these two cases the run time was cut in half?

Looks like it!

Debraheem:
Wow, even the [1M_pre_ms_to_wd](https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/test_cases/star/1M_pre_ms_to_wd) appears 2x as fast now!

fxt44:
rolling full test suites on b92a98b; m1 and intel macs.

pmocz:
> Wow, even the [1M_pre_ms_to_wd](https://testhub.mesastar.org/bugfix%2Fpmocz%2Fparallel-simd/test_cases/star/1M_pre_ms_to_wd) appears 2x as fast now!

Wait! It looks like the # threads used on a test is not consistent across runs. It's not shown in the columns by default. Sometimes it's 8 and sometimes it's 16. I think the speedup we see is consistent with just shaving off a few seconds

evbauer:
Nice catch on the different thread counts @pmocz!

A bit of background on why this is the case... A while ago Rob and I decided to try moving the longest running tests to the beginning of `do1_test_source` so that their jobs would launch first, improving the net turnaround time for complete testing of a commit. I also set Cannon up so that the first 10 tests get 16 cores, while the rest of the tests all run on 8. That way the really long-running ones like `ppisn` and `1M_pre_ms_to_wd` can run to completion somewhat faster.

In addition to the automated testing that picks up and tests every commit, Cannon has a separate scheduled nightly job to run the head of `main` (if it hasn't already been run), and do a couple extra things:
* ignore the `[ci optional]` flag and always run all optional inlists
* force push test logs to the logs server, even for tests that are passing

That way in theory we always have a recent reference point including what log info should look like for each test. This nightly run isn't time-sensitive, so I didn't bother to give the longest-running tests 16 cores for faster turnaround.

But if all this makes interpreting the timing history too complicated, maybe I should go ahead and make the nightly runs have the same core counts as the regular runs for consistency on Cannon.

fxt44:
> This PR should speed up Mesa by a small percent

we'll take it. first blood pmocz!

PR #670, opened 2024-07-01
adding missing openmp directives and making python files (in linters/) python 3.12+ compatible

pmocz:
Just minor missing openmp directives from my exploration of performance last week. My compiler seems smart enough that most of these don't make much difference in timings and must be happening already. But I thought I'd add them for completion and see if they affect timings in the test suite on other machines


star/private/star_bcyclic.f90
```diff
@@ -447,7 +448,7 @@ subroutine cycle_rhs( &
          nmin = 2
          op_err = 0
 !$OMP PARALLEL DO SCHEDULE(static,3) &
-!$OMP PRIVATE(ns,shift1,ipivot,shift2,k,dmat,dmatF,X,row_scale_factors,col_scale_factors,equed,i,okay,op_err)
```
pmocz, line 450: variable is not being used

mtx/public/mtx_solve_routines.inc
```diff
@@ -282,7 +283,8 @@
          real(dp) :: temp
          integer :: i,k
          info = 0
-         do i = 1, n
+         !$omp simd private(temp)
```
pmocz, line 286: Follow the general strategy of `!$OMP PARALLEL` outer loop and `!$omp simd` inner loop

const/test/src/test_const.f90
```diff
@@ -0,0 +1,52 @@
...
+      error stop 1
+   end if
+
+   call do_test_const
+
+contains
+
+   subroutine do_test_const
+
+      character(*), parameter :: fmt1 = "(a40, 3x, 1pe24.16)"
```
pmocz, line 22: rewrite in modern fortran

linters/check_columns.py
```diff
@@ -73,9 +74,9 @@ def print_options(options):
 def delete_command(options, filename):
     print("for i in", " ".join(options), end=";")
     print("do ", end="")
-    print('sed -i "/^\s*\!$i/d" */' + filename, end=";")
-    print('sed -i "/^\s*\!\s*$i/d" */' + filename, end=";")
-    print('sed -i "/^\s*$i/d" */' + filename, end=";")
+    print(r'sed -i "/^\s*\!$i/d" */' + filename, end=";")
```
pmocz, line 77: Need to use raw expressions, otherwise Python 3.12 complains

Debraheem (approved):
Looks good. I like the extra docs clean up as well!

PR #674, opened 2024-07-09
binary_tides for fully convective stars

matthiasfabry:
When doing tidal synchronization with `Hut_conv` on fully convective stars, the timescale was computed as `NAN`, which messes up `s% extra_jdot(k)` and makes `do_omega_mix` fail.


orlox (approved):
AOK to merge from me

PR #683, opened 2024-07-18
Adding test-mesa Github Action workflow

pmocz:
(Moving part of the build-linux github workflow into its own github action for re-usability)

Adding a github workflow to automatically run a small subset of problems (<6 hours) and upload to mesa-hub.


pmocz:
Adding in building with latest sdk version (24.7.1) and merging this now. Currently, running a subset of mesa_test test problems and pushing results is triggered on release or manually, to not use up too much of github action's free-tier resources

PR #684, opened 2024-07-19
Revert "Part 1: automating documentation/images for the test suite"

evbauer:
Reverts MESAHub/mesa#681 until we can get Read the Docs builds working.


pmocz:
Thank you so much for catching this, @evbauer ! That requirement is not needed. I accidentally left it in while experimenting.

PR https://github.com/MESAHub/mesa/pull/681  is actually a pretty small change. It adds one dev feature for future capability to auto-generate docs, that is not exposed to Users. The doc updates are mostly style and spelling fixes / cleanup, except for a couple cases where I moved readme code into files. 

I think the docs will build if I remove the redundant requirement.  If not, then we should revert this PR for now. But in terms of adding features for the next release there isn't anything major yet. That will be a future PR which will require a lot of manualy effort first to re-format all the Test problem readme's

pmocz:
Oh! One more thing. Looks like readthedocs is unhappy with the tags, which a couple of the readme's have. I will remove these for now, and add as a feature in the future, post-release. We should probably add tags all at once to all the test problems

evbauer:
I'm seeing a lot of "Requires-Python >=3.9" in some of the failures. Perhaps we could try updating the top level `.readthedocs.yaml` file to bump the python version from 3.8 to 3.9 or higher?

pmocz:
> I'm seeing a lot of "Requires-Python >=3.9" in some of the failures. Perhaps we could try updating the top level `.readthedocs.yaml` file to bump the python version from 3.8 to 3.9 or higher?

That must be why readthedocs doesn't like the tags!

For now I just took out the tags  in (https://github.com/MESAHub/mesa/pull/687  https://github.com/MESAHub/mesa/pull/686 ) so users will see virtually no change before the release. Let's add them in a future PR after the release. 

The docs are building now, so I will close this PR because it is no longer needed.

evbauer:
Awesome! Thanks for the quick turnaround on getting this passing. I'm excited to see this new feature get fleshed out in due course!

PR #688, opened 2024-07-23
Updating starspot test_suite problem

pmocz:
adding figure


star/test_suite/starspots/inlist_starspots
mjoyceGR, line 1: built and ran it on my machine and can confirm it's not behaving as expected. This is being weird without making even the cosmetic changes of replacing x_ctrl 1 and 2 with inlist parameters named xspot/fspot. Can investigate more thoroughly tomorrow

pmocz: I thought so too. The star is in a different part of the HR diagram than the stars in Figure 15 of the MESA VI paper: https://arxiv.org/abs/2208.03651v

I just ran the `starspot` test problem with Mesa `mesa-r22.11.1-rc1` and the HR diagram looks similar to that with the current version of Mesa
image

This leads me to suspect maybe there is a small error/typo in the version of the `inlist` or `run_stars_extras.f90` for this test problem that was uploaded to the repo?

pmocz: We tracked this down! My bad for making the x-axis range odd, and @mjoyceGR found out we need to also  increase the max age to reproduce plots in the paper

PR #695, opened 2024-07-26
Adding test_suite docs Python autogenerate capabilities

pmocz:
Adding capability to autogenerate test_suite figures with `matplotlib`. I see a good chunk of test problems have specialized figures made with Python (hopefully those scripts exist somewhere on people's machines and we can add them to the repo to make all tests reproducible!)

With this PR:

`./each_test_run -u TEST_NUMBER` will now also run your python plotting scripts if they are included in the test problem directory and automatically update figures in the docs if that figure is referenced in `README.rst`.   (plotting script has to be named `plot.py`, and should output to a folder `plt_out`)

An example is shown with the `1.3M_ms_high_Z` test problem. The doc pages looks like: 
image

Reminder: It is best to save figures as `.svg` for crisp rendering on the web!


pmocz (commented):
merging in to have example ready for mesa-palooza event tomorrow to update test problems

PR #696, opened 2024-07-30
num: make test src f2008 standard compliant

VincentVanlaer:

PR #702, opened 2024-08-05
Make `num` comply (mostly) with Fortran 2008 standard

warrickball:
This is the continuation of @VincentVanlaer's PR #696 (thanks Vincent!), which already made `num/test` comply with the Fortran 2008 standard. The rest of `num` is now compiled with a Fortran 2008 flag so the aim is to reduce the number of warnings before merging this into `main`.

We should make sure we can still build with `ifort` before merging.


astero/private/adipls_support.f90
```diff
@@ -682,8 +681,7 @@ subroutine read_and_store(iriche, iturpr, cgrav)
             write(*,*) 'read_and_store failed in read_fgong_file'
             call mesa_error(__FILE__,__LINE__)
          end if
-         call fgong_amdl( &
```
pmocz, line 685: This is a bug. The second input variable `nn` (intent(in)) does not have a defined value at this point in the code. It should be `nn_in`

warrickball:
I think so? I'm not sure if we want to wait until after the release. I personally don't mind but wouldn't mind at least another dev chipping in.

pmocz:
> I think so? I'm not sure if we want to wait until after the release. I personally don't mind but wouldn't mind at least another dev chipping in.

Probably best to merge this in right after release actually, since we'd like to get the release out very soon and even though we tested this branch it doesn't add too much to users (yet, without the other successive PRs which should be left for after the release)

pmocz:
Merging now that we already have the 24.08.1 release branch

PR #704, opened 2024-08-08
Utils warn cleanup

VincentVanlaer:
I noticed some warnings in the C code in the utils module regarding the way strncpy was used. I ended up refactoring the code to make it less complex, and applied a formatter (in a separate commit).


pmocz:
Adding this into a mesa branch to test before merging into main

PR #712, opened 2024-08-09
mtx: drop quad precision support

VincentVanlaer:
I noticed some warnings in the mtx module which came from the quad precision code. After some messing around with the code, I noticed that none of it was currently in use. Therefore, I removed all of the quad precision code. If the dev team wants to keep this around for future use, feel free to close :)


warrickball:
I think we might need @orlox to weigh in here. I recall the discussion of quad-precision sums for advanced burning in Sec. 5 of [MESA III](https://ui.adsabs.harvard.edu/abs/2015ApJS..220...15P) but I'm not sure if that's still used nor if it warranted bundling all the quad-precision linear algebra. I noticed [MESA II](https://ui.adsabs.harvard.edu/abs/2013ApJS..208....4P) mentions quad-precision linear algebra in Sec. B.6, when discussing the separate solution of the angular momentum profile. So maybe these were once used in the rotation (and maybe binary?) routines but aren't any more.

fxt44:
quad precision linear algebra is mostly a leftover from the mesa III experiments. for nuclear burning, only summing the contributions to each isotope in quad is absolutely necessary. quad jacobians, hence quad linear algebra, turned out to not be as important (but we did check!)

VincentVanlaer:
Rebased on main

PR #714, opened 2024-08-13
bugfix nn_in

pmocz:
Since we are waiting after impending release to get https://github.com/MESAHub/mesa/pull/702 in, let's get this small bugfix in before the release

### DIFF
--- a/astero/private/adipls_support.f90
+++ b/astero/private/adipls_support.f90
@@ -683,7 +683,7 @@
             call mesa_error(__FILE__,__LINE__)
          end if
          call fgong_amdl( &
-            cgrav, nn, iconst, ivar, ivers, glob, var, data, aa, nn, ierr)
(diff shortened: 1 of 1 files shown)


astero/private/adipls_support.f90
```diff
@@ -683,7 +683,7 @@ subroutine read_and_store(iriche, iturpr, cgrav)
             call mesa_error(__FILE__,__LINE__)
          end if
          call fgong_amdl( &
-            cgrav, nn, iconst, ivar, ivers, glob, var, data, aa, nn, ierr)
```
pmocz, line 686: If you look above this line, `nn` is uninitialized and this should be `nn_in` since this function argument has `intent(in)`

warrickball: Agreed. `nn_in` is populated by `read_fgong_file`, even though the variable name is misleading.

warrickball (approved):

PR #715, opened 2024-08-14
[Draft]: auto diff net rates

pmocz:


net/private/net_eval.f90
```diff
@@ -356,73 +369,121 @@ subroutine eval_net_approx21_procs(n,just_dxdt, ierr)
...
          ierr = 0
 
          g => n% g
 
          num_isos = g% num_isos
-         
+        
+        ! here we define our new auto_diff variable for reaction rates.
+        !auto_diff_real_2var_order1
+        n% rate_screened_ad %val = n% rate_screened
```
Debraheem, line 382: @pmocz  this is where we define the rate_screened and its derivates. When auto_diff is implemented into rates, we can just set one value `n% rate_screened_ad %val = n% rate_screened`, instead of all three calls. Then our derivatives won't be singular values for `n% rate_screened_ad %d1val1`, rather they will carry the entire interpolating function inside`n% rate_screened_ad` which can be differentiated at any moment to yield the derivative. This helps later on when these derivatives are used to calculate the derivative of the equilibrium links.

PR #716, opened 2024-08-21
bugfix dP to d_P in create_initial_model

pmocz:
This is a bug in the RK4 solver of `create_initial_model` that has been in Mesa since before the code was added to github. `dP` needs to be `d_P`. The variable `dP` was never defined, and got interpreted as `dp` because Fortran is case-insensitive.

Luckily this bug does not seem to have a large effect in the solver convergence. Tested here: https://testhub.mesastar.org/bugfix%2Fpmocz%2Fquickfixes/commits/1aec194

### DIFF
--- a/star/private/create_initial_model.f90
+++ b/star/private/create_initial_model.f90
@@ -334,11 +334,11 @@
 
             ! 4th order rk step
             dy1=dydP*d_P
-            call derivs(cs,P+dP/2d0,S,y+dy1/2d0,dydP)
+            call derivs(cs,P+d_P/2d0,S,y+dy1/2d0,dydP)
             dy2=dydP*d_P
-            call derivs(cs,P+dP/2d0,S,y+dy2/2d0,dydP)
+            call derivs(cs,P+d_P/2d0,S,y+dy2/2d0,dydP)
             dy3=dydP*d_P
-            call derivs(cs,P+dP,S,y+dy3,dydP)
+            call derivs(cs,P+d_P,S,y+dy3,dydP)
             dy4=dydP*d_P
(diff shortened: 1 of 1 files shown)


evbauer (approved):

rhdtownsend:
I wonder if this bug is the reason why MESA-Web doesn’t converge to a starting model for certain choices of the initial mass. I’ll do some digging...

> On Aug 21, 2024, at 8:32 AM, Philip Mocz ***@***.***> wrote:
> 
> This is a bug in the RK4 solver of create_initial_model that has been in Mesa since before the code was added to github. dP needs to be d_P. The variable dP was never defined, and got interpreted as dp because Fortran is case-insensitive.
> Luckily this bug does not seem to have a large effect in the solver convergence
> You can view, comment on, or merge this pull request online at:
>   https://github.com/MESAHub/mesa/pull/716
> Commit Summary
>     • 1aec194 fix dP to d_P
> File Changes 
> (1 file)
>     • M star/private/create_initial_model.f90 (6)
> Patch Links:
>     • https://github.com/MESAHub/mesa/pull/716.patch
>     • https://github.com/MESAHub/mesa/pull/716.diff
> —

--

pmocz:
Thanks @rhdtownsend I'm curious to see what you find. It's part of the story. I'm also seeing I can sometimes get convergence for non converging models if I choose a different initial guess for the temperature. Is there a way to chose it more smartly? https://github.com/MESAHub/mesa/pull/716/files#diff-ed13c8f5578322afea3296aa7fab7414804b9a67c99c95a41b17e90a95cbc102R484

pmocz:
I'll merge this into main so we can all start trying it out

Debraheem:
A good time to try regenerating the zams models.

PR #721, opened 2024-08-23
[skip ci] fix and expand file_device and file_extension documentation

mathren:
Regarding this mailing list request: https://lists.mesastar.org/pipermail/mesa-users/2024-August/015307.html

Documentation was incorrectly suggesting `pdf` files can be produced directly from pgplots and the option to produce `ps` and `eps` were instead not documented.

### DIFF
--- a/star/defaults/pgstar.defaults
+++ b/star/defaults/pgstar.defaults
@@ -87,14 +87,24 @@
 
       ! file_device
       ! ~~~~~~~~~~~
-      ! file_extension
-      ! ~~~~~~~~~~~~~~
 
-      ! 'png' is Portable Network Graphics format; can also use 'pdf'
+      ! 'png' is Portable Network Graphics format
+      ! can also use:
+      ! 'ps' (in combination with file_extension = 'ps') for PostScript
+      ! or:
+      ! 'vcps' (in combination with file_extension = 'eps') for Encapsulated PostScript
 
       ! ::
 
     file_device = 'png'
+
+      ! file_extension
+      ! ~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


pmocz (approved):

warrickball:
I just had a quick play around with this myself and I'm not sure it's clear that `file extension` is just what is literally used in the filename, has no bearing on the format and is not limited to a string set of options. So although users probably _want_ to use an extension associated with the relevant device, it's not required.

If we only intend to write files with the relevant file extensions, we should remove the `file_extension` option and write code that selects the extension based on `file_device`. From what little I know about PGPLOT, it theoretically supports other options for `file_device` but I can't see us (or our users) wanting anything but `ps`, `vcps` or `png`. (I briefly tried to find a list of supported devices but didn't manage to find one that mentioned `png`, which is clearly supported.)

pmocz:
> I just had a quick play around with this myself and I'm not sure it's clear that `file extension` is just what is literally used in the filename, has no bearing on the format and is not limited to a string set of options. So although users probably _want_ to use an extension associated with the relevant device, it's not required.
> 
> If we only intend to write files with the relevant file extensions, we should remove the `file_extension` option and write code that selects the extension based on `file_device`. From what little I know about PGPLOT, it theoretically supports other options for `file_device` but I can't see us (or our users) wanting anything but `ps`, `vcps` or `png`. (I briefly tried to find a list of supported devices but didn't manage to find one that mentioned `png`, which is clearly supported.)

Great idea! Removing the `file_extension` option will minimize confusion. I created a PR here: https://github.com/MESAHub/mesa/pull/722

PR #722, opened 2024-08-25
removing file_extension user option

pmocz:
Inspired by @warrickball 's comment: https://github.com/MESAHub/mesa/pull/721#issuecomment-2308515534

this PR removes the `file_extension` option from inlists for pgstar/pgbinary because it is redundant with `file_device`. The file extension is now deduced from the pgplot device that is set.

### DIFF
--- a/binary/defaults/pgbinary.defaults
+++ b/binary/defaults/pgbinary.defaults
@@ -87,15 +87,14 @@
 
       ! file_device
       ! ~~~~~~~~~~~
-      ! file_extension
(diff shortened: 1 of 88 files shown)


Debraheem:
We might need to remove these extensions in the dev test cases as well. Otherwise, if this is working, perhaps we can merge?

Debraheem (approved):

pmocz:
looks like all instances of `file_extension` are removed, so I will merge this

PR #723, opened 2024-08-26
mtx: don't vendor lapack

VincentVanlaer:
It is only used for non-sdk builds, and other required libraries (such as HDF5) are not included for non-sdk builds as well. Furthermore, this prevents the vendored version from being out of date with the SDK version.


pmocz:
This makes sense to me. Just have the sdk take care of lapack, since that's what we test and support, and remove 3rd party source code from Mesa. We should merge this PR if others agree

warrickball:
I agree in principle but this will break my ifort build until I switch it over to using the system LAPACK, which I should do anyway but might not have much time for ~2-3 weeks.

warrickball (approved):

PR #724, opened 2024-08-26
Add new build system

VincentVanlaer:
# Main commit message

Reasons for replacing the previous build system were:

- the previous build system had a lot of duplication between the modules, both in terms of make files and shell scripts.
- parallelisation was limited. The new build system parallelises better within a module, and between modules.
- build scripts and output from the build ended up in the same folder. With the new system, all build output files are stored in the folder  ``build`` in the repository root, making cleaning build files equivalent to deleting this folder

For the details of the implementation, see the newly added documentation (Developing -> Build system).

Some further notable changes:

- makedepf90 has been replaced by a perl script that scans the fortran source files. This allows for better flexibility in construction the make files.
- Currently, `pymesa` does not work with the new build system, as the new system does not provide a way to build shared libraries for every module.
- Dependencies are being discovered using pkg-config. This change makes it easier to build MESA without the SDK, as linux distributions typically ship with pkg-config files.

# Draft status

Draft status of this PR will be removed once a decision has been mode about #712 and #723, after which I will rebase.

# Reviewing this PR

I recommend to first take a test drive of the new system, especially if you use some special setup (e.g. a non-sdk build). Install instructions haven't changed. I have written relatively detailed documentation on how the new system works, which I recommend reviewers of this PR to read before diving into the Makefiles and scripts. Let me know if something is unclear or missing in the documentation! The first few commits are just #712 and #723, so those can be skipped (if #712 and #723 don't end up being merged, it is an easy change to add the files back). The other commits are truly part of this PR, with the first two introducing the new build system. The other commits are porting of individual modules.

# Remarks

- About half of the modules have been ported.
- There is a lot of output on each run for installed files to `$MESA_DIR/lib` and `$MESA_DIR/include`. Once all modules have been ported, this will no longer be needed for a normal install, except for the data files. Internally, the different modules that have been ported don't use those files.
- I generate version numbers using `git describe --tags` which generates (in my opinion) nicer version numbers than just using the commit hash. However, this requires the version tags to be on the main branch (currently running this command on `main` gives `r15140-2084-ge29ed74d` rather than something like `r24.08.1-NNNN-ge29ed74d`).
- It would be beneficial to add GitHub actions for the following cases
  - Building without SDK
  - Building all helper programs which are not built by default
- As mentioned in the main commit message, `pymesa` currently doesn't work. This is because I think it would make more sense to build one single shared library after the main build as a separate step, rather than a bunch of little libraries which need to be able to refer to each other. I intend to do this anyway for the output of the star/binary module, such that building the actually star binary can proceed faster, and allow for things such as experimenting with link-time optimization.

# Future iterations

I have thought about some future iterations of the build system after this, but I have not implemented those things since otherwise I would keep iterating forever on this build system trying to design the perfect build system. Hence, writing this down to at least have it documented somewhere.

1. Use a single make invocation to build and test all modules. Currently, each module is built in a separate make invocation, which limits parallelisation and true dependencies between modules (once the mod files have been generated, the dependent MESA modules could start building). Furthermore, if you change something in an include file which is included in another module, make might not immediately pick it up. Combining all modules into one make invocation would prevent such surprises in the development. In order to not cause excessive rebuilds, it would be necessary to use Fortran submodules, such that no public modules are dependent on their implementation. All of this requires a full set of generated make files, as to not overlap the variables and targets used by the different modules.
2. With such a system in place, it may be beneficial to replace the generated make files by [ninja](https://ninja-build.org/). The main issue with the system from 1. (and in fact the current system as well) this tries to solve, is that if you remove an include file, make will error out. This is because the include file is in the dependencies of the module. From what I understand, by using dynamic dependencies in ninja, it will not error out in such a case. However, this very much untested. In general, I think it is possible to make a build system in this way that exactly compiles what is needed, and doesn't error if cleaning and building wouldn't cause an error as well. Maybe there are easier ways to fix that in the end than using a different command runner.

I have probably forgotten some things I wanted to mention, but this should be most of it.


warrickball:
I've just started chipping away at this and started off with the error
```
mkdir -p build
make/gen-folder-deps "make" const utils math mtx auto_diff num interp_1d interp_2d chem eos forum > build/depend
fatal: No names found, cannot describe anything.
make: *** [make/version.mk:2: data/version_number] Error 128
make: *** Deleting file 'data/version_number'
```
for both `make` and `./install`, which emanates from `git tag`. I couldn't for the life of me fetch any tags so I just tagged the `HEAD` commit on your `build-system` branch and that got the build going.

I'll see how far I get with the SDK (the build _feels_ faster) and then give it a try with my system libraries, which includes GCC 14.2.1, and then the ifort build on the local cluster. It'll take me at least 2-3 weeks to work through but is my MESA priority.

VincentVanlaer:
> I've just started chipping away at this and started off with the error
> 
> ```
> mkdir -p build
> make/gen-folder-deps "make" const utils math mtx auto_diff num interp_1d interp_2d chem eos forum > build/depend
> fatal: No names found, cannot describe anything.
> make: *** [make/version.mk:2: data/version_number] Error 128
> make: *** Deleting file 'data/version_number'
> ```
> 
> for both `make` and `./install`, which emanates from `git tag`. I couldn't for the life of me fetch any tags so I just tagged the `HEAD` commit on your `build-system` branch and that got the build going.
> 

I noticed this was an issue with CI, but worked locally. I think at some point tags for the releases got moved and renamed, and the main branch therefore contains no tags when you do a clean checkout. It should be fixed now (still have to cleanup the mess I made of the commits for testing)

pmocz:
> > I've just started chipping away at this and started off with the error
> > ```
> > mkdir -p build
> > make/gen-folder-deps "make" const utils math mtx auto_diff num interp_1d interp_2d chem eos forum > build/depend
> > fatal: No names found, cannot describe anything.
> > make: *** [make/version.mk:2: data/version_number] Error 128
> > make: *** Deleting file 'data/version_number'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > for both `make` and `./install`, which emanates from `git tag`. I couldn't for the life of me fetch any tags so I just tagged the `HEAD` commit on your `build-system` branch and that got the build going.
> 
> I noticed this was an issue with CI, but worked locally. I think at some point tags for the releases got moved and renamed, and the main branch therefore contains no tags when you do a clean checkout. It should be fixed now (still have to cleanup the mess I made of the commits for testing)

I apologize, I should have recognized that error above was due to missing tags. Yes, we unfortunately had to rename tags and move them around to tips of benches get readthedocs to auto-detect the versions, render the right docs, and create banners. Sounds like this caused a headache, but I am glad it is sorted out. At some point I'd like to go back to branch-based docs builds but that will require renaming branches to follow PEP440 standards to get readthedocs to auto-detect versions

pmocz:
Looking good! I am able to install without problems. What is the status and plan for this PR, @VincentVanlaer ?

VincentVanlaer:
This can be merged as is in principle, since it builds. However, I would not yet merge this to main (maybe to a separate branch), but first:

- port the remaining modules (I am happy to do that, but if you want have a go at it, please do so :smiley: )
- fix the `pymesa` issue (this will be easier once all of the modules have been ported)
- ask for some volunteers on the users list to test on their machines to see if something breaks

I would like to have a review of the new documentation and the system in general. What are show stoppers, things that should be improved, ...

warrickball:
I'm hopefully getting back on top of enough other things that I've had a look back at some MESA things, and this is my MESA priority. I just checked out out your `mesa` fork afresh, switched to the `build-system` branch and ran `./install` as usual, which finished successfully. I was then able to run a work folder. Nice! :tada: 

For anyone else who tries all this, note that a top-level `make` doesn't build everything (yet). You still need to run `./install.sh` as before.

I'd like to push to get the whole of MESA under the new build before we merge. I don't think it makes sense to have some modules build one way and other modules build another. 

The first thing I want to try is to update another module, to see how that process works. I'm not sure if the modules have to be updated in order because of their dependencies but looks like `colors` is next in `install.sh` and is a reasonable place to start anyway because it's small. I plan to mimic how `chem` appears to have been ported (which is mostly deleting scripts :ok_hand:).

A lower-priority goal is to see how easy it is to switch away from the SDK as the build backend but I don't think that has to be in this PR.

Debraheem:
🚀

VincentVanlaer:
> I just pulled your branch again and now seem to be hitting an issue in the GYRE build when installing (with `./install` or `make`):

Ah right, that's because I forgot to add a pkg-config file to `make/pkgconfig` for lapack95 (fortran bindings for lapack) which gyre requires.

pmocz:
For building on Mac, `realpath` does not support the `-s` option (in `helpers.mk`). I can do a `brew install coreutils` and use `grealpath` instead. But I am wondering if there is a simpler solution that introduces less dependency / more portable. Maybe we need to write our own shell function?

VincentVanlaer:
> For building on Mac, `realpath` does not support the `-s` option (in `helpers.mk`). I can do a `brew install coreutils` and use `grealpath` instead. But I am wondering if there is a simpler solution that introduces less dependency / more portable. Maybe we need to write our own shell function?

It is worth noting that this transformation is not required, it just makes the resulting path a bit cleaner to look at in log messages and such.

warrickball:
I now seem to quite consistently run into this error when I run `./install`:

```
LIB_TOOL libstar.a
MAKEDEPF90
TEST_COMPILE ../src/run_star_extras.f90
TEST_COMPILE ../../../star/job/run_star.f90
TEST_COMPILE ../src/run.f90
make: *** No rule to make target '../../../lib/libgyre_mesa.a', needed by '../star'.  Stop.

FAILED
```

VincentVanlaer:
I merged the mesa frontend for GYRE with GYRE itself into a single static library. This also seems to depend on features of `ar` that are not consistent across linux distros (let alone with another OS). I will probably revert that change and just allow for multiple outputted libraries.

rhdtownsend:
If you’re doing this sort of thing, be careful that you don’t end up including libforum into whatever libraries you create. There’s some reference counting that goes on in libforum, relating to the HDF5 library, and this breaks if there are multiple copies of libforum that end up in the final executable.

cheers,

Rich

> On Nov 14, 2024, at 5:14 PM, Vincent Vanlaer ***@***.***> wrote:
> 
> 
> I merged the mesa frontend for GYRE with GYRE itself into a single static library. This also seems to depend on features of ar that are not consistent across linux distros (let alone with another OS). I will probably revert that change and just allow for multiple outputted libraries.
> —

--

VincentVanlaer:
> If you’re doing this sort of thing, be careful that you don’t end up including libforum into whatever libraries you create. There’s some reference counting that goes on in libforum, relating to the HDF5 library, and this breaks if there are multiple copies of libforum that end up in the final executable.

Thanks for the tip! I will double check that, but I think it should be fine. What I intend to do is the same thing MESA does currently, namely building GYRE with `FORUM=no` and then copying over `libgyre_mesa.a` and `libgyre.a`.

warrickball:
I just got back to a successful build and run of the `work` folder so I'll try to have a look at porting some more modules.

VincentVanlaer:
I tried to get the testhub to run the test suite, but I the compilation fails and the logs only show the help text, and no errors: https://testhub.mesastar.org/VincentVanlaer%2Fbuild-system/commits/8baecdd. Can someone who has access have a look at this?

evbauer:
Sure! Here's what I'm seeing:
```
Entering /n/netscratch/conroy_lab/Lab/ebauer/mesa-git-tests/tmp.awksxmgd1n.
MESA_DIR = /n/netscratch/conroy_lab/Lab/ebauer/mesa-git-tests/tmp.awksxmgd1n
./install
mkdir -p build
make/gen-folder-deps "make" const utils math mtx auto_diff num interp_1d interp_2d chem eos forum colors rates neu net gyre kap ionization atm turb star_data star astero binary adipls > build/depend
./install: line 210: syntax error near unexpected token `then'
./install: line 210: `if [[ $? -ne 0 ]] then'
Leaving  /n/netscratch/conroy_lab/Lab/ebauer/mesa-git-tests/tmp.awksxmgd1n

Entering /n/home02/ebauer/mesa-helios-test.
Changed MESA_DIR back to /n/netscratch/conroy_lab/Lab/ebauer/mesa-git-tests/tmp.awksxmgd1n.

Failed in compiling MESA.
```

evbauer:
I think it just need a semicolon or newline before `then` at line 210?

VincentVanlaer:
Thanks!

It worked locally (and in github actions), so that's probably because of the older version of bash on CentOS 7 then. I fixed it in the other branch, but no logs are begin uploaded it seems.

evbauer:
Hmmm, yeah the build job failed when trying to submit build.log to the log server for some reason. Not sure why that happened, but here's the tail end of the build output where it failed:
```
pgstar.defaults:2524:34:

Warning: Change of value in conversion from 'REAL(8)' to 'REAL(4)' at (1) [-Wconversion]
ar rcs -o ../build/star/lib/libstar.a   ../build/star/other/other_accreting_state.o ../build/star/other/other_adjust_mdot.o ../build/star/other/other_adjust_mlt_gradt_fraction.o ../build/star/other/other_after_set_mixing_info.o ../build/star/other/other_after_solver_setmatrix.o ../build/star/other/other_alpha_mlt.o ../build/star/other/other_am_mixing.o ../build/star/other/other_astero_freq_corr.o ../build/star/other/other_brunt.o ../build/star/other/other_brunt_smoothing.o ../build/star/other/other_build_initial_model.o ../build/star/other/other_cgrav.o ../build/star/other/other_close_gaps.o ../build/star/other/other_d_mix.o ../build/star/other/other_diffusion.o ../build/star/other/other_diffusion_coefficients.o ../build/star/other/other_diffusion_factor.o ../build/star/other/other_energy.o ../build/star/other/other_energy_implicit.o ../build/star/other/other_eps_grav.o ../build/star/other/other_eval_fp_ft.o ../build/star/other/other_eval_i_rot.o ../build/star/job/other_extras.o ../build/star/other/other_gradr_factor.o ../build/star/other/other_j_for_adjust_j_lost.o ../build/star/other/other_kap.o ../build/star/other/other_mesh_delta_coeff_factor.o ../build/star/other/other_mesh_functions.o ../build/star/other/other_mlt_results.o ../build/star/other/other_momentum.o ../build/star/other/other_momentum_implicit.o ../build/star/other/other_net_derivs.o ../build/star/other/other_net_get.o ../build/star/other/other_neu.o ../build/star/other/other_opacity_factor.o ../build/star/other/other_overshooting_scheme.o ../build/star/other/other_pgstar_plots.o ../build/star/other/other_photo_read.o ../build/star/other/other_photo_write.o ../build/star/other/other_pressure.o ../build/star/other/other_rate_get.o ../build/star/other/other_remove_surface.o ../build/star/other/other_rsp_build_model.o ../build/star/other/other_rsp_linear_analysis.o ../build/star/other/other_screening.o ../build/star/other/other_set_pgstar_controls.o ../build/star/other/other_solver_monitor.o ../build/star/other/other_split_burn.o ../build/star/other/other_surface_pt.o ../build/star/other/other_timestep_limit.o ../build/star/other/other_torque.o ../build/star/other/other_torque_implicit.o ../build/star/other/other_wind.o ../build/star/private/adjust_mass.o ../build/star/private/adjust_mesh.o ../build/star/private/adjust_mesh_split_merge.o ../build/star/private/adjust_mesh_support.o ../build/star/private/adjust_net.o ../build/star/private/adjust_xyz.o ../build/star/private/alloc.o ../build/star/private/atm_support.o ../build/star/private/auto_diff_support.o ../build/star/private/brunt.o ../build/star/private/conv_premix.o ../build/star/private/create_initial_model.o ../build/star/private/ctrls_io.o ../build/star/private/diffusion.o ../build/star/private/diffusion_procs.o ../build/star/private/diffusion_support.o ../build/star/private/do_one_utils.o ../build/star/private/element_diffusion.o ../build/star/private/eos_support.o ../build/star/private/eps_grav.o ../build/star/private/eps_mdot.o ../build/star/private/evolve.o ../build/star/private/evolve_support.o ../build/star/private/gravity_darkening.o ../build/star/private/history.o ../build/star/private/history_specs.o ../build/star/private/hydro_alpha_rti_eqns.o ../build/star/private/hydro_chem_eqns.o ../build/star/private/hydro_energy.o ../build/star/private/hydro_eqns.o ../build/star/private/hydro_momentum.o ../build/star/private/hydro_riemann.o ../build/star/private/hydro_rotation.o ../build/star/private/hydro_rsp2.o ../build/star/private/hydro_rsp2_support.o ../build/star/private/hydro_temperature.o ../build/star/private/hydro_vars.o ../build/star/private/init.o ../build/star/private/init_model.o ../build/star/private/ionization_potentials.o ../build/star/private/kap_support.o ../build/star/private/magnetic_diffusion.o ../build/star/private/mass_utils.o ../build/star/private/mesh_adjust.o ../build/star/private/mesh_functions.o ../build/star/private/mesh_plan.o ../build/star/private/micro.o ../build/star/private/mix_info.o ../build/star/private/mod_typical_charge.o ../build/star/private/net.o ../build/star/private/neu.o ../build/star/private/overshoot.o ../build/star/private/overshoot_exp.o ../build/star/private/overshoot_step.o ../build/star/private/overshoot_utils.o ../build/star/private/paquette_coeffs.o ../build/star/private/phase_separation.o ../build/star/private/photo_in.o ../build/star/private/photo_out.o ../build/star/private/pre_ms_model.o ../build/star/private/predictive_mix.o ../build/star/private/profile.o ../build/star/private/profile_getval.o ../build/star/private/pulse.o ../build/star/private/pulse_cafein.o ../build/star/private/pulse_fgong.o ../build/star/private/pulse_gr1d.o ../build/star/private/pulse_gsm.o ../build/star/private/pulse_gyre.o ../build/star/private/pulse_osc.o ../build/star/private/pulse_saio.o ../build/star/private/pulse_utils.o ../build/star/private/read_model.o ../build/star/private/relax.o ../build/star/private/remove_shells.o ../build/star/private/report.o ../build/star/private/rotation_mix_info.o ../build/star/private/rsp.o ../build/star/private/rsp_build.o ../build/star/private/rsp_def.o ../build/star/private/rsp_eval_eos_and_kap.o ../build/star/private/rsp_lina.o ../build/star/private/rsp_relax_env.o ../build/star/private/rsp_step.o ../build/star/job/run_star_support.o ../build/star/private/set_flags.o ../build/star/private/solve_omega_mix.o ../build/star/private/solver_support.o ../build/star/private/star_bcyclic.o ../build/star/private/star_history_def.o ../build/star/private/star_job_ctrls_io.o ../build/star/private/star_private_def.o ../build/star/private/star_profile_def.o ../build/star/private/star_solver.o ../build/star/private/star_utils.o ../build/star/private/starspots.o ../build/star/private/struct_burn_mix.o ../build/star/private/timestep.o ../build/star/private/turb_info.o ../build/star/private/turb_support.o ../build/star/private/winds.o ../build/star/private/write_model.o ../build/star/public/star_def.o ../build/star/public/star_lib.o ../build/star/job/run_star.o ../build/star/other/pgstar_decorator.o ../build/star/private/pgstar_ctrls_io.o ../build/star/private/pgstar_support.o ../build/star/private/pgstar_hist_track.o ../build/star/private/pgstar_kipp.o ../build/star/private/pgstar_dpg_dnu.o ../build/star/private/pgstar_trho.o ../build/star/private/pgstar_tmaxrho.o ../build/star/private/pgstar_hr.o ../build/star/private/pgstar_r_l.o ../build/star/private/pgstar_r_teff.o ../build/star/private/pgstar_l_teff.o ../build/star/private/pgstar_l_r.o ../build/star/private/pgstar_l_v.o ../build/star/private/pgstar_logl_teff.o ../build/star/private/pgstar_logl_r.o ../build/star/private/pgstar_logl_v.o ../build/star/private/pgstar_logg_teff.o ../build/star/private/pgstar_logg_logt.o ../build/star/private/pgstar_history_panels.o ../build/star/private/pgstar_trho_profile.o ../build/star/private/pgstar_mode_prop.o ../build/star/private/pgstar_power.o ../build/star/private/pgstar_production.o ../build/star/private/pgstar_abundance.o ../build/star/private/pgstar_dynamo.o ../build/star/private/pgstar_mixing_ds.o ../build/star/private/pgstar_summary_burn.o ../build/star/private/pgstar_summary_profile.o ../build/star/private/pgstar_summary_history.o ../build/star/private/pgstar_summary.o ../build/star/private/pgstar_color_magnitude.o ../build/star/private/pgstar_profile_panels.o ../build/star/private/pgstar_grid.o ../build/star/private/pgstar_network.o ../build/star/other/sample_pgstar_plot.o ../build/star/private/pgstar_full.o
cd test; ../../build/net/bin/checker | grep -Ev "^(write)|(create rate data for)|( read )|( write )|(                           number not already in cache:)" > ../../build/net/check-results
diff -b ../build/net/check-results test/test_output

!! MESA install failed, see above or build.log for details !!
```

VincentVanlaer:
Because building takes place in parallel, the error can end up earlier in the log file. That seems to be the case here, so the snippet is not the all the useful (although it is good to see that it managed to run the test for the star module)

evbauer:
Ah! Thanks for cluing me in. Here's the full output of the build job this time.
[build.txt](https://github.com/user-attachments/files/18560112/build.txt)

VincentVanlaer:
Thanks! I suspect the errors are due to bugs in the version of pkg-config on CentOS 7.

From the logs it seems that certain tests fail to link, with missing symbol errors in the `num` module for symbols in the `mtx` module:

```
/n/holystore01/LABS/conroy_lab/Lab/ebauer/mesasdks/mesasdk-24.7.1/bin/../lib/gcc/x86_64-pc-linux-gnu/13.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: /n/netscratch/conroy_lab/Lab/ebauer/mesa-git-tests/tmp.QxqcFMKJBA/num/private/mod_newton.f:637:(.text+0x10391): undefined reference to `__mtx_lib_MOD_multiply_xa'
/n/holystore01/LABS/conroy_lab/Lab/ebauer/mesasdks/mesasdk-24.7.1/bin/../lib/gcc/x86_64-pc-linux-gnu/13.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: /n/netscratch/conroy_lab/Lab/ebauer/mesa-git-tests/tmp.QxqcFMKJBA/num/private/mod_newton.f:635:(.text+0x103e0): undefined reference to `__mtx_lib_MOD_block_multiply_xa'
```

This does not seem to happen for the `num` module itself, but for certain dependent modules of the `num` module (`kap` and `atm` are examples). `mtx` does provide these functions, so that doesn't seem to be the issue. For static libraries, there is an additional requirement that dependencies are given later than the libraries that depend on them. From the logs it looks like that is not honored. Here is an example for the `atm` test binary:

```
-latm -lkap -lcrmath -lmtx -llapack -lblas -lforum -lhdf5_fortran -leos -lnum -linterp_1d -lchem -linterp_2d -lauto_diff -lmath -lutils -lconst
```

Note that `-lmtx` is given before `-lnum`, while `num` depends on `mtx`. When I do the compile locally, `mtx` is after `num` as expected. From what I can gather this was fixed in pkg-config [0.28](https://gitlab.freedesktop.org/pkg-config/pkg-config/-/blob/d97db4fae4c1cd099b506970b285dc2afd818ea2/NEWS#L80-89).

evbauer:
Nice! Thanks for digging through that and diagnosing it. I've actually had to turn off automated testing on the Cannon cluster for now. If I manage to bring it back up later this week, I'll look into whether I can get the testing to load in an updated version of `pkg-config`. Otherwise, we might need somebody else to throw their cluster at testing this branch.

@pmocz @warrickball @wmwolf anybody want to throw your cluster at testing this branch?

warrickball:
I've just started a local build, which is what I've been doing so far without running (or reporting on) the test suite. I definitely want to get `bluebear_ifort` going on this because it's a non-SDK build but it'll probably be a few week(end)s before I'll get around to it.

VincentVanlaer:
I'll see if I can get the test suite to run on the cluster of our institute, because it will take way to long on my laptop otherwise :smiling_face_with_tear:

PR #725, opened 2024-08-27
(minor) cleanup warnings in binary

pmocz:
cleaning up unused variables and some other compiler warnings in `binary`

changed a `dP` to `d_P` to avoid possible conflict with `dp` (since Fortran is case-insensitive)

Goal is that we will add more (best-practice) compile flags to avoid future bugs, but first I need to clean up warnings before committing the flags because there are a lot.

### DIFF
--- a/binary/private/binary_mdot.f90
+++ b/binary/private/binary_mdot.f90
@@ -502,7 +502,7 @@
          use binary_wind, only: eval_wind_xfer_fractions
          type (binary_info), pointer :: b
 
-         real(dp) :: fixed_xfer_fraction, actual_mtransfer_rate
+         real(dp) :: actual_mtransfer_rate
          integer :: ierr
 
          actual_mtransfer_rate = 0d0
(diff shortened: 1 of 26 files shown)


binary/private/binary_mdot.f90
```diff
@@ -735,7 +735,7 @@ end subroutine get_info_for_ritter
       real(dp) function calculate_kolb_mdot_thick(b, indexR, rl_d) result(mdot_thick)
          real(dp), intent(in) :: rl_d
          integer, intent(in) :: indexR
-         real(dp) :: F1, F3, G1, dP, q, rho, p, grav, hp, v_th, rl3, q_temp
+         real(dp) :: F1, F3, G1, d_P, q, q_temp
```
pmocz, line 738: It's best to change this from `dP` to `d_P`. I'm not sure how the (case-insensitive) Fortran compiler handles with with `dp` also being defined

PR #732, opened 2024-09-11
Do not guard calls mesa_error() with report_ierr

pmocz:
The user flag `report_ierr = .true.` is used as a debugging tool to report diagnostics on internal errors in Mesa but not stop the code. Flipping the switch to true should not change the outcome of the simulation. Therefore, we should not call `mesa_error()` inside `if(report_ierr == .true.)` if statements. Either always just halt the code, or use guards like `if(stop_for_bad_nums == .true.)`. 

There are just a handful of cases in the code where this pattern was not followed. This is dangerous because it can cause non-reproducible results when someone is trying to debug their run switches `report_ierr = .true.` on. Particularly in simulations that chain more than one inlist. For example, running `./run_all` for the `20M_pre_ms_to_core_collapse` test problem with `report_ierr = .true.` was yielding a different result than switching diagnostics off because one of the inlists in the middle of the chain was ending prematurely and the next inlist restarted the simulation from the last photo written.

**TL;DR:** fix so that switching on diagnostics reporting does not change outcome of the simulation

### DIFF
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -11216,7 +11216,7 @@
       ! stop_for_bad_nums
       ! ~~~~~~~~~~~~~~~~~
 
-      ! If true and report_ierr is also true, then stop for bad numbers (NaNs or infinity).
+      ! If true and ``report_ierr`` is also true, then stop for bad numbers (NaNs or infinity).
       ! this replaces old control stop_for_NaNs
 
       ! ::
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -11216,7 +11216,7 @@
       ! stop_for_bad_nums
       ! ~~~~~~~~~~~~~~~~~
 
-      ! If true and ``report_ierr`` is also true, then stop for bad numbers (NaNs or infinity).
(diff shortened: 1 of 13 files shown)


star/defaults/controls.defaults
```diff
@@ -11216,7 +11216,7 @@
       ! stop_for_bad_nums
       ! ~~~~~~~~~~~~~~~~~
 
-      ! If true and report_ierr is also true, then stop for bad numbers (NaNs or infinity).
+      ! If true and ``report_ierr`` is also true, then stop for bad numbers (NaNs or infinity).
```
warrickball, line 11219: At a glance this doesn't look like what we've changed the behaviour to. Instead it looks like the correct description of the new behaviour (which I prefer and support) is
```suggestion
      ! If true, then stop for bad numbers (NaNs or infinity).
```
I.e., `stop_for_bad_nums` now stops whether or not `report_ierr` is true because the `stop_for_bad_nums` checks our outside the `if (report_ierr)` blocks.

pmocz: Oh! Very good point! I'll make the change.  `stop_for_bad_nums` can simply function independently from having diagnostics switched on (as it already does in most cases)

PR #736, opened 2024-10-01
Refactor surface corrections to re-use same loop regardless of `radial_only`

warrickball:
The surface correction loops over the modes currently work by always deriving the radial correction and then proceeding onto the non-radial corrections if `radial_only` is `.false.`. This PR consolidates the radial and non-radial loops by using `radial_only` to decide whether to loop over *ℓ* up to 0 (if `.true.`) or 3 (if `.false.`).

I also cleared out trailing whitespace in that file while I was at it.

The `astero` test `surface_effects` passed locally.

### DIFF
--- a/astero/private/astero_support.f90
+++ b/astero/private/astero_support.f90
@@ -9,7 +9,7 @@
 !   by the free software foundation; either version 2 of the license, or
 !   (at your option) any later version.
 !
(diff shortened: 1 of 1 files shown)


pmocz (approved):
Nice simplification of the code 🚀

PR #739, opened 2024-10-08
Changes to lift the dependence of turb on star_data

rhdtownsend:
While making some enhancements to the thermohaline code (which is part of the turb module), I discovered a bad code smell: the turb module depends on the star_data module. This doesn't make architectural sense; turb handles macrophysics, and shouldn't know anything about stars.

The dependency comes through turb/private/tdc_support.f90; the convert() and unconvert() functions need to know how many variables are stored in auto_diff_real_star_order1%d1Array. This number is currently stored in the auto_diff_star_num_vars parameter defined in star_data/public/star_data_def.inc. 

A functionally equivalent way to do this, which also fixes the architectural sense, is to use the actual declared length of auto_diff_real_star_order1%d1Array rather than auto_diff_star_num_vars. 

This PR makes this change, plus a few others that accompany it:

* Move star_data from the LIBS_MESA_MICRO definition in utils/makefile_header (where it shouldn't be) to the LIBS_MESA_STAR_SUPPORT definition
* Rearrange the build order in ./install so that turb builds just after atm, and star_data builds after all the micro/macrophysics modules (turb included).

I'd appreciate a quick turnaround on the PR, as it's needed in order for the new thermohaline stuff to work.


pmocz (approved):
Looks good to merge once all tests pass: 
 https://testhub.mesastar.org/remove_bogus_turb_dependence/commits/621a3b6

Issue #651, opened 2024-06-04
need test case for rotating core-collapse progenitor?


mathren:
The attached figure shows a (not-so-great) resolution test for a Z=0.001, 40Msun star initially rotating at ~0.6 critical which experiences chemically homogeneous evolution. The figure only shows the profile of specific angular momentum, density, and velocity of the inner 5Msun and the numbers in the legend are the number of zones at this particular snapshot.

This is a rather convoluted setup (damping spurious velocities in the layers not in sonic contact in the core and trying to add resolution for a bunch of gradients), but very similar to the setup [here](https://github.com/mathren/CHE_net_comparison). The point that worries me is the "spiky" infall velocity profile (more visible in the blue, but it's coming up in the orange and hidden a bit by the dashed linestyle). 

I have been noticing this in many rotating models (even with less extreme rotation rates), but I just double checked that the test cases that go to core-collapse in r24.03.1 instead exhibit smooth velocity profiles, however they are also non-rotating.

I suspect this has to do with the rotation playing with the lagrangian hydrodynamics. 
I would like to know if any of the developers has suggestions on how to fix this (@Debraheem maybe?), and if/once fixed, if there is interest in molding this into a test case for rotating core-collapse progenitors:

![image](https://github.com/MESAHub/mesa/assets/49067756/87ca3a25-f871-4cc4-8749-c545d51d4ab3)

Debraheem:
Thanks for documenting this. What network did you use for this run in particular? Can you overplot the logT profile for this run? Are you using split_burn, if so what is the min T for splitting hydro and burning? I'll try to reproduce this locally, but if you have a working model directory I can run, that would make it easier to compare.

mathren:
The zenodo should be close enough to reproduce the issue (and that one I got to onset of core-collapse as defined there with many nets and resolutions).

This is with `approx21_plus_cr56.net` (classic case of "do what I say, not what I do", but in my defense, it's all testing). We do use `split_burn` from logT=9 onwards, and hydro (as in `v_flag`) from the very beginning.

I can produce the plot, but I expect logT to be as smooth as logRho here (and we are nearing `fe_core_infall`, the issue is roughly in the Si-rich layer, so logT~9.3 or so)

mathren:
added bottom panel with temperature (for the inner 5Msun):


![image](https://github.com/MESAHub/mesa/assets/49067756/63624d03-17a9-4c02-873d-55c45ebccc5c)

[...]

PR #755, opened 2024-11-30
add new rotating 40Msun to cc test_suite for testing

Debraheem:
It had been brought up many times that we should have a test case that demonstrates the evolution of a rotating model to CC. Here is a new test case for a 40Msun rotating model that evolves from pms to cc. This is a tweaked version of the model shared in this issue https://github.com/MESAHub/mesa/issues/651, from which this pr is sponsored. Since it is not always so easy to hit cc, this test_suite uses u_flag for the evolution to CC. I welcome and encourage anyone to modify this test_suite how they see fit and take ownership! One could even try using u_flag earlier if you'd like, although you might see some pulsation development at the surface.


Debraheem:
fyi, I switched off u_flag and stuck with v_flag in the last commit, although I kept the options for u_flag available. The model seems to work fine, although I turned off drag_coefficient in inlist_to_cc as damping the waves generated from burning seems problematic and might even crash the model.

Debraheem:
This test case just needs some minor tweaks and some clean up and it should be ready for a merge. @aurimontem , please let me know if you have any opinions.

Debraheem:
build failure is lfs related, just ignore.

PR #759, opened 2024-12-12
Introduce mass ratio

matthiasfabry:
Small QOL improvement for binary runs.

In binary simulations, the mass ratio is important enough to have a history column available out of the box in mesa (and it required for plotting with `pgbinary`).
I understand the definition of `q` is not universal, some fields use `m1/m2`, others `m2/m1` (I chose the latter, typically the initially less massive over the initially more mass star). Hence I also implemented `min(m2/m1, m1/m2)` as an "observational" mass ratio.


matthiasfabry:
Working on this reminded me to at a certain time strip the copying of `binary_history` into `(star_)history` and its infrastructure, since that is not required for `pgstar` plotting anymore.

orlox (approved):
aok

PR #760, opened 2025-01-06
fortitude linting

pmocz:
Using https://github.com/PlasmaFAIR/fortitude to fix warnings that often lead to code errors, if not careful.

Fixing: T051, S061, T041, T002 ...


kap/private/op_eval_mombarg.f90
```diff
@@ -909,7 +909,7 @@ subroutine compute_kappa_fast(k,fk, logT_cntr, logRho_cntr,lkap_ross_cell, dlnka
       real(dp):: epa_mix_cell(1648), amu_mix_cell ! Number of electrons per atom, mean molecular weight, density and temperature as a function of ite (temp index) and jne (density index) from the OP mono data.
       real(dp) :: delta(1648)
       real(dp) :: lgamm_cell(nel) !interpolated log kappa Rossland and log gamma_k in each cell (k = element index).
-      real(dp) :: dv = (1.0552976117319748d0 - 0.00010565516589892675d0)/nptot !v(u(1)) - v(u(nptot))/nptot
+      real(dp), parameter :: dv = (1.0552976117319748d0 - 0.00010565516589892675d0)/nptot !v(u(1)) - v(u(nptot))/nptot
```
warrickball, line 912: We define this same parameter repeatedly with the same name and value. Does it make more sense as a module-level variable or something?

star/private/gravity_darkening.f90
```diff
@@ -99,7 +99,8 @@ function GD_coeff(omega,inclination,C1) result(coeff)
     real(dp), intent(in) :: omega, inclination
     real(dp), pointer :: C1(:)
     real(dp) :: coeff, coeff_eval(6), safe_omega, safe_incl
-    integer :: ict(6)=[1,0,0,0,0,0], ierr
+    integer, parameter :: ict(6)=[1,0,0,0,0,0]
+    integer ierr
```
warrickball, line 103: While we're at it, how about
```suggestion
    integer :: ierr
```

star/private/pgstar_kipp.f90
```diff
@@ -107,11 +107,12 @@ subroutine do_Kipp_Plot(s, id, device_id, &
 
          integer :: ix,k
          real :: xleft,xright,now
-         real :: dxmin=-1.d0
+         real :: dxmin
 
          include 'formats'
 
          ierr = 0
+         dxmin=-1.d0
```
warrickball, line 115: Might as well comply with the surrounding style.
```suggestion
         dxmin = -1.d0
```

warrickball: Actually, in `pgstar_rti.f90` (the next file), this is instead marked as a `parameter`. These should be consistent and I suspect it's `pgstar_rti.f90` that's at fault.

star/private/pgstar_rti.f90
```diff
@@ -93,7 +93,7 @@ subroutine do_rti_Plot(s, id, device_id, &
 
          integer :: ix,k
          real :: xleft,xright,now
-         real :: dxmin=-1.d0
+         real, parameter :: dxmin = -1.d0
```
warrickball, line 96: This isn't a `parameter` in the previous file. Is that deliberate? This looks like it might be something that should be defined to a computed positive value later.

kap/private/op_eval_mombarg.f90
pmocz: Excellent idea!

star/private/pgstar_kipp.f90
pmocz: Agree, it should not be a parameter. I will mark is a `save` to keep functionality exactly as before, since that is what the compiler is doing implicitly

PR #637, opened 2024-03-27
w_div_w_crit_roche update

matthiasfabry:
While playing with very short period binaries using the single rotating star fp/ft corrections, I noticed the that the sigmoid function that is used to smooth very large `j_rot` was interpreted incorrectly. I fixed this in the documentation.
I also cleaned up some obsolete `ctrls`, and took the liberty of updating the polynomial fits to the more accurate ones of Fabry+2022.

PR #773, opened 2025-01-31
Tidal deformation

matthiasfabry:
Following on #637, this (rather large) overhaul of `hydro_rotation.f90` includes support for computing the stellar structure corrections of tidally distorted stars. This is especially useful when modeling contact binaries.

The setup is as follows:

I used procedure hooks (akin to `binary_other_torque` for tidal J_dot calculations) to communicate the binary calculated fp, ft, and i_rot values for the stellar layers.
Then these values are "blended" with the single-rotating-star corrections according to the synchronicity of the component to the orbit. I implemented a user hook `binary_other_tidal_deformation_switch_function` if one does not agree with this treatment.

The user can still use the `other_fp_ft` and `other_i_rot` hooks to bypass all default calculations (and blending).


binary/defaults/binary_controls.defaults
```diff
@@ -49,9 +49,9 @@
 
 
 ! initial_eccentricity
-! ~~~~~~~~~~~~~~~~~~~~
 
 ! Initial eccentricity of the system
+! ~~~~~~~~~~~~~~~~~~~~
```
warrickball, line 54: I'm pretty sure Sphinx will emit a warning if the underline isn't as long as the heading.
```suggestion
! ~~~~~~~~~~~~~~~~~~~~
```

binary/make/makefile
```diff
@@ -3,6 +3,6 @@
 # goal of QUIET is to make it easier to spot warnings and/or circular dependency
 # errors.
 
-QUIET = yes
+QUIET = no
```
warrickball, line 6: I think we want `yes` in production.
```suggestion
QUIET = yes
```

binary/other/mod_other_tidal_deformation_switch_function.f90
```diff
@@ -0,0 +1,56 @@
...
+   subroutine null_other_tidal_deformation_switch_function(id, k, omega_in, f_switch, ierr)
+      integer, intent(in) :: id, k
+      real(dp), intent(in) :: omega_in
+      real(dp), intent(out) :: f_switch
+      integer, intent(out) :: ierr
+      include 'formats'
+
+      write (*, 1) 'no implementation for other_tidal_deformation_function'
+      ! must set f_switch
+      ierr = -1
```
warrickball, line 51: Though it's quite common in the blank `other_*` routines, just a comment that `ifort` will warn that `f_switch` has no value here.

binary/private/binary_roche_deformation.f90
```diff
@@ -0,0 +1,357 @@
...
+   use binary_def
+   use binary_utils, only : eval_rlobe
+
+
+   implicit none
+
+   real(dp), parameter :: nudge = 1d-4
+   real(dp), pointer :: xvals(:), yvals(:), yvals_gtr_than_1(:), fpfunc1d(:), ftfunc1d(:), &
+         irotfunc1d(:), otherrfunc1d(:), afunc1d(:)
+   logical :: inter_ok = .false., dbg = .true.
```
warrickball, line 45: I'm guessing we don't generally want the debug output on, otherwise it isn't really debug output.
```suggestion
   logical :: inter_ok = .false., dbg = .false.
```

binary/private/binary_roche_deformation.f90
```diff
@@ -0,0 +1,357 @@
...
+                  fpfunc1d, num_xpts, testval, ierr)
+            write(*, 1) 'fp   test gave should be close to 0.6', testval
+            call interp_evbipm_db(xtest, ytest, xvals, num_xpts, yvals, num_ypts,&
+                  ftfunc1d, num_xpts, testval, ierr)
+            write(*, 1) 'ft   test gave should be close to 0.8', testval
+            call interp_evbipm_db(xtest, ytest, xvals, num_xpts, yvals, num_ypts,&
+                  irotfunc1d, num_xpts, testval, ierr)
+            write(*, 1) 'irot test gave should be close to 0.4', testval
+            inter_ok = .true.
+         end if
```
warrickball, line 84: I won't have a chance to look at this closely for a bit but do we want to `inter_ok` to `.true.` *inside* the debug block? Should it not be set to `.true.` in normal operation? I'd expect
```suggestion
         end if
         inter_ok = .true.
```

binary/private/binary_roche_deformation.f90
```diff
@@ -0,0 +1,357 @@
...
+      use const_def, only: mesa_data_dir
+      real(dp) :: xtest, ytest, testval
+      integer :: ierr
+      character(len=strlen) :: upstairs
+      include 'formats'
+      if (.not. inter_ok) then
+         upstairs = trim(mesa_data_dir) // '/roche_data/'  ! where fp/ft data lives
+         if (dbg) then
+            write(*, 1) 'starting interpolator setup'
+         end if
```
warrickball, line 60: Though it's cosmetic, we can just have
```suggestion
         if (dbg) write(*, 1) 'starting interpolator setup'
```

binary/private/binary_roche_deformation.f90
```diff
@@ -0,0 +1,357 @@
...
+      type (star_info), pointer :: s
+      type (binary_info), pointer :: b
+      integer :: ierr, j, this_star=0, other_star=0
+      real(dp) :: r_roche, a, m1, m2, ar, lq, eval, d_eval_d_ar
+
+      ierr = 0
+      call star_ptr(id, s, ierr)
+      if (ierr /= 0) return
+      call binary_ptr(s% binary_id, b, ierr)
+      if (ierr /=0 ) return
```
warrickball, line 249:
```suggestion
      if (ierr /= 0) return
```

star/private/hydro_rotation.f90
```diff
@@ -169,7 +169,57 @@ real(dp) function w_div_w_roche_jrot(rpsi, Mphi, jrot, cgrav, max_w, max_w2, w_d
...
+         integer :: ierr
+
+         include 'formats'
+
+         call star_ptr(id, s, ierr)
+         if (ierr /= 0) return
+
+         i_rot = 0d0
+         if (s% simple_i_rot_flag) then
+            i_rot = (2d0/3d0)*r00*r00
```
warrickball, line 190:
```suggestion
            i_rot = two_thirds*r00*r00
```

star/private/hydro_rotation.f90
```diff
@@ -169,7 +169,57 @@ real(dp) function w_div_w_roche_jrot(rpsi, Mphi, jrot, cgrav, max_w, max_w2, w_d
...
+            i_rot% d1Array(i_w_div_wc_00) = ir%d1val1
+            i_rot% d1Array(i_lnR_00) = ir%d1val2
+         end if
+
+      end subroutine eval_i_rot
+
+
       subroutine set_i_rot(s, skip_w_div_w_crit_roche)
+         use auto_diff
+
```
warrickball, line 222: What do we need from `auto_diff` here?

star/private/hydro_rotation.f90
```diff
@@ -243,44 +299,94 @@ subroutine check_omega(s, str)
       end subroutine check_omega
 
       subroutine update1_i_rot_from_xh(s, k)
+         use auto_diff
```
warrickball, line 302: As above, what do we need from `auto_diff` here?

warrickball:
Not a blocker for me but would it be possible to use this in the test suite somewhere, be that in a new test or tweaking an existing one?

Debraheem:
Certainly this needs a test_suite. I would love to test it out as well.

binary/private/binary_roche_deformation.f90
matthiasfabry: you're absolutely right!

star/private/hydro_rotation.f90
matthiasfabry: a simple dev remnant, good catch

matthiasfabry:
As a first test, `double_bh` contains a contact phase, plus it has strong tides, so this is a natural environment for the tidal deformation corrections. I incorporated a `pgbinary` window for good measure.

matthiasfabry:
the tests on [c79cdd6](https://testhub.mesastar.org/tidal_deformation/commits/c79cdd6) all pass! I'll leave it up here for anyone to test/review, but from my end this is ready to merge.

star/test_suite/ppisn/src/run_star_extras.f90
```diff
@@ -331,79 +333,78 @@ subroutine my_adjust_mdot(id, ierr)
...
-               if (s% q(j) > 0.999d0) then
-                  fp(j) = 1d0
-                  ft(j) = 1d0
-               else if (s% q(j) > 0.998d0) then
-                  alpha = (1d0-(s% q(j)-0.998d0)/(0.001d0))
-                  fp(j) = fp(j)*alpha + 1d0*(1-alpha)
-                  ft(j) = ft(j)*alpha + 1d0*(1-alpha)
-               end if
-            end do
+            if (s% q(k) > 0.999) then
```
pmocz, line 389: add `d0` 's after 0.999, 0.998, 0.001

matthiasfabry: good catch!

matthiasfabry:
I was able to resolve the merge conflicts, please feel free to test now. For some reason, the testhub reports that my [ci optional] call on [8fc5948](https://github.com/MESAHub/mesa/pull/773/commits/8fc594821e1b5f1799b6844e940e231e1c038b59) only completed partially (90 untested), but the merge commit [f9bd070](https://github.com/MESAHub/mesa/pull/773/commits/f9bd070d493c9f528447e34189dbee06700a1ddc) shows 105 passes and 1 untested.

warrickball:
Is this just waiting for testing? I'll submit a run on our cluster with the SDK now.

warrickball:
My full test suite runs (incl. "optional" inlists) seem to be consistently failing on [`ppisn`](https://testhub.mesastar.org/tidal_deformation/commits/25e0fd6/test_cases/star/ppisn). It passes on `ccalin046` but I suspect that's because it's not doing the optional parts of the inlist.

I'm not likely to have enough time to dig into why any time soon but I noticed that you have changed some stuff in `ppisn` as well as bits of `star`, so maybe this is just because `ppisn` is such a flaky test. Not sure who to summon to help with `ppisn`.

Debraheem:
The version of ppisn in the test_suite is a stripped down version of the one Rob, Pablo, Mathieu, and I used. It's missing some bells and whistles that stabilize the surface and prevent running into EOS issues. I've added back one, which limits the surface velocities so that zones cannot unrealistically exceed the speed of light or runaway. A future project for me is to re-enable convection during the pulsation, which should also help prevent runaway in the surface layers. 

It should complete successfully now.

warrickball:
`ppisn` still [passes on `bluebear`](https://testhub.mesastar.org/tidal_deformation/commits/c67c019/test_cases/star/ppisn) after turning rotation back on in c67c01919ffdff67de632ede2f531ecaa14c7f5e.

pmocz:
Thanks @Debraheem and @warrickball !

Issue #776, opened 2025-02-07
Limiting convective velocities with max_conv_vel_div_csound


JRGarza:
Hi, 
I noted the control `max_conv_vel_div_csound` is deffined in the documentation (for 23.05.1 and later versions), as well in ctrls_io, controls.defaults and star_controls but it is not used in the code. I notice this after trying to use the control to limit convection velocities below sound speed but did not see any difference, I am using Mihalas MLT option. Is there a reason for this, or it is just a remanent code from previous MESA versions?

Debraheem:
Thanks for opening this issue. Hmm, It looks like a remanent deprecated control, but it should be functional in theory, although I don't see it being used internally anywhere. It appears to be a topic of an open issue that was never addressed, so as of now there is no conv_vel limiter.

see https://github.com/MESAHub/mesa/issues/373

perhaps, we can revive it... but is it necessary? 

For now, this control should either be removed, or put to use.

Debraheem:
We are having git-lfs bandwidth issues at the moment, but when they are resolved, I can push a branch with a conv_vel limiter.

Debraheem:
Perhaps you could git checkout this branch https://github.com/MESAHub/mesa/tree/EbF/add_Cs_limit_to_TDC? Let me know if you have any thoughts. I'm not immediately sure how to limit the conv_vel in mlt self-consistently like we do in TDC. The hack might have to cut it.

JRGarza:
Thanks for the answer. For now I required the option for the 'Mihalas' option (as in my science case I have issues using TDC, but thanks for referring to the [add_Cs_limit_to_TDC](https://github.com/MESAHub/mesa/tree/EbF/add_Cs_limit_to_TDC) branch). I made a PR #781 for this control in case it is useful. As for the justification, the MLT prescriptions (at least Cox and Mihalas) are not valid for supersonic convection velocities. The argument in Cox and Giuli's book (chapter 14) states that turbulent pressure by non-supersonic convection is negligible and not taken into account for the derivation, contrary to the case of supersonic convection.

Debraheem:
I think the only difference is that you are using the actual sound speed at the cell face, where as I just used an estimate. I can correct that.

JRGarza:
> [add_Cs_limit_to_TDC](https://github.com/MESAHub/mesa/tree/EbF/add_Cs_limit_to_TDC)

[...]

Issue #373, opened 2022-02-17
Sound speed limiter on TDC?


adamjermyn:
@orlox suggested that we impose a (smooth) limit on the convection speed in TDC, and implemented this in the `tdc_gamma_no_cs` branch.

I'd just like to have a bit of discussion about the physics involved here before we merge that in. In particular, I'd like to hear from @rsmolec on whether or not there is actually physics limiting convection to the sound speed, and to how consistent any additional physics we might include to create such a limit is with the physics already in TDC.

wmwolf:
@rsmolec will certainly know better than I, but isn't it usually the case that if convective velocities approach the sound speed, some assumptions in MLT are already invalid? That's why I've always had some respect for MLT++... it knew that things were probably wrong, so it just swapped those wrong assumptions for other wrong assumptions that were numerically more stable.

adamjermyn:
Was MLT++ about supersonic convection? I always thought it was about locally super-Eddington convection, but I could well be confused...

My immediate reaction to 'MLT + supersonic' is that I get worried about the various uses of Cp, because the pressure fluctuations become large, but I don't think that changes anything quantitatively.

I also get worried that maybe the turbulent dissipation is no longer ~`vc^3` and instead something like `vc^3 * (vc/cs)^n` for some `n`. That wouldn't quite serve as a limiter, but would make it harder to get strongly supersonic flows...

But I'm definitely naive here...

wmwolf:
I think you're right that the primary trigger for MLT++ was super-Eddington rather than supersonic convective velocities. My recollection is that super-Eddington convective luminosity would usually coincide with supersonic convective velocities near the surface. Deeper in the star, though, I think it can still be super-Eddington without being supersonic.

So, my comment may be a non-sequitur for the scenario at hand.

fxt44:
generalizing adam's question just a bit, is there anywhere where supersonic convection takes place within stars?

for one, mach numbers in the oxygen and silicon shells at the onset of core-collapse peak at M~ 0.3 depending on mass.

rsmolec:
[...]

PR #780, opened 2025-02-09
add conv_vel cs limit to mlt/tdc

Debraheem:
To make use of deprecated controls `max_conv_vel_div_csound` and `max_conv_vel_div_csound_maxq` by request https://github.com/MESAHub/mesa/issues/776 and https://github.com/MESAHub/mesa/issues/373. This pr introduces a cs limiter, initially designed by adam/pablo for tdc. We can also introduce a limit for mlt, similar to the one present in previous releases. At the moment, this pr only includes the cs limiter for conv_vel inside TDC.

This pr has also highlighted some potential issues in the building of the pre_ms_model. We are unable to call `s% q(k)` inside mlt when building the pre_ms_model because it is not yet set. In fact, I've marked a couple statements inside the pms build routine which could be problematic. 

I don't mind removing `max_conv_vel_div_csound_maxq `, but ideally, we can call q(k) inside the mlt_routine during the pms model creation.

worst case, we don't introduce either of these controls, but in that case we should remove them all together.


Debraheem (commented):
Ideally, q and dq are set before this subroutine which calls mlt/tdc to create the pms model. 
```
       call look_for_brackets(lnd, dlnd, lnd1, lnd3, pre_ms_f, y1, y3, &
               imax, pre_ms_lrpar, rpar, pre_ms_lipar, ipar, ierr)
```

Debraheem:
This pr needs a changelog, and some testing, then it will be ready for a merge.

Debraheem:
Reconsideration, perhaps we can introduce a convective enthalpy flux limiter as well for TDC, as in Radek's thesis. This might be more appropriate than a strict convective velocity limter for TDC. It might help convergence on the pre-ms for TDC. https://acta.astrouw.edu.pl/Vol58/n3/pap_58_3_4.pdf. This would be useful for RSP as well.

star/private/turb_support.f90
```diff
@@ -201,12 +201,23 @@ subroutine Get_results(s, k, MLT_option, &  ! NOTE: k=0 is a valid arg
...
       grav = cgrav*m/pow2(r)
+      max_conv_vel = 1d99
       if (s% use_Ledoux_criterion) then
          gradL = grada + gradL_composition_term  ! Ledoux temperature gradient
       else
          gradL = grada
       end if
 
+      ! maximum convection velocity.
+      if (k>=1) then
```
Debraheem, line 212: Unfortunately checking `if associated(s% Csound_face) .and. associated(s% q)` still yields in backtrace errors from astero in the `fast_from_file` and other tests, see https://testhub.mesastar.org/EbF%2Fadd_Cs_limit_to_TDC/commits/0a89a8e.

 Using a check for `k >=1` seems to work better.

PR #800, opened 2025-03-18
remove implicit save attribute to binary variables

pmocz:
implicit save attribute can be a source of bugs, especially for restart

### DIFF
--- a/binary/private/binary_photos.f90
+++ b/binary/private/binary_photos.f90
@@ -132,6 +132,8 @@ contains
          b% CE_lambda2, b% CE_lambda2_old, &
          b% CE_Ebind1, b% CE_Ebind1_old, &
          b% CE_Ebind2, b% CE_Ebind2_old, &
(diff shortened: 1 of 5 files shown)


matthiasfabry:
Strange, none of these parameter had the keyword `save` tagged to them. Is this the default behavior of FORTRAN compilers? To assume any variable that is assigned at declaration is `save`?

warrickball:
AFAIK this is part of the Fortran standard, e.g. in Sec. 11.2 in [this document](https://wg5-fortran.org/N1601-N1650/N1601.pdf),

> (R1104) If an object of a type for which component-initialization is specified (R444) appears in the specification-part of a module and does not have the ALLOCATABLE or POINTER attribute, the object shall have the SAVE attribute.

matthiasfabry:
Huh! The more you know... This will likely save me (ha get it) tons of debugging time!

PR #806, opened 2025-04-07
use consistent values for constants

pmocz:
Sometimes constants get re-defined (to slightly different values) in parts of the code.  Making things consistent by calling the values from the `const` module.

### DIFF
--- a/eos/eosDT_builder/src/scvh_core.f90
+++ b/eos/eosDT_builder/src/scvh_core.f90
@@ -21,6 +21,9 @@
 ! ***********************************************************************
 
 module scvh_core
+
+   use const_def, only: clight, avo, kerg
+
    implicit none
 
    logical, parameter :: dbg = .false.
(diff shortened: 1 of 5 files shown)


eos/eosDT_builder/src/scvh_core.f90
```diff
@@ -637,10 +640,9 @@ subroutine interpolate_scvh(include_radiation, search_for_SCVH, logT, logRho, T,
       double precision :: dpressraddd, dpressraddt, deraddd, deraddt, dsraddd, dsraddt
 
 !..constants
-      double precision :: clight, ssol, asol, asoli3, avo, kerg, xka, mh1, mhe4, third
-      parameter(clight=2.99792458d10, ssol=5.67051d-5, asol=4.0d0*ssol/clight, &
-                asoli3=asol/3.0d0, avo=6.0221367d23, kerg=1.380658d-16, &
-                xka=kerg*avo, mh1=1.67357d-24, mhe4=6.646442d-24, third=1.0d0/3.0d0)
+      double precision ::  ssol, asol, asoli3, xka, mh1, mhe4
+      parameter(ssol=5.67051d-5, asol=4.0d0*ssol/clight, asoli3=asol/3.0d0, &
+                xka=kerg*avo, mh1=1.67357d-24, mhe4=6.646442d-24)
```
warrickball, line 645: I think these are defined in `const` too:
* `ssol` → `boltz_sigma`
* `asol` → `crad`
* `xka` → `cgas`

`mh1` and `mhe4` can probably be more informatively expressed in terms of `amu` or something.

warrickball: Not sure if these changes are ever actually run or if we would need to explictly rebuild the EoS tables, and if that would have knock on effects.

PR #808, opened 2025-04-14
Make pgplot color indices static

VincentVanlaer:
This is refactor of the pgstar colors that partially originates from the build system branch.

The first two commits (those effectively come from the build system branch) are just a cleanup of the color setting code. I replaced some custom code to map color names to rgb values using a file provided by pgplot with an equivalent function in pgplot itself. There was 

The next two commits then get rid of the dynamic mapping of color indices, and just make the list static, since that is what ends up happening in practice anyway. I have also lifted the actual named colors from the pgplot provided rgb.txt.

I would consider the first set of changes more or less of a bug fix, so I would like to see those go in. The second set is a refactor I would do, but not something I would consider necessary.

### DIFF
--- a/astero/private/pgstar_astero_plots.f90
+++ b/astero/private/pgstar_astero_plots.f90
@@ -127,6 +127,7 @@
 
          use utils_lib
          use const_def, only: dp
+         use pgstar_colors
 
          integer, intent(in) :: id, device_id
          real, intent(in) :: xleft, xright, ybot, ytop, txt_scale
@@ -399,6 +400,7 @@
 
          use utils_lib
          use const_def, only: dp
+         use pgstar_colors
 
          integer, intent(in) :: id, device_id
          real, intent(in) :: xleft, xright, ybot, ytop, txt_scale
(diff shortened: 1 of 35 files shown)


pmocz:
This is great!

VincentVanlaer:
The way pgplot works is that in order to draw something in a certain color, you have to

1. Map a color index to the color (the rgb values) with `pgscr`
2. Select that color index before you draw the thing you want with `pgsci`

At the initialization of pgplot, MESA sets up a bunch of color indices. This happens whenever a pgplot device is opened, so also when a plot is saved as png for example. The values of those indices where computed dynamically, and written to bunch of global variables (`clr_Red`, `clr_Green`, ...). These variables would always have the exact same values, and map to the exact same color, since that is all hard coded (except for white-on-black vs black-on-white). This PR makes those variables static instead.

When one of the panels selects a color, it does so by taking one of these global variables/parameters and pass them to `pgsci`. So in order to have dynamic colors, one either needs to have some flag to switch which parameter to use, or use the color indices directly (e.g. reading them from an inlist). But none of this has been affected by this PR (in fact, it makes it a bit easier to see what the values of those color indices are, and to keep them actually stable as colors are added).

pmocz:
Looks good! @VincentVanlaer ready to merge?

pmocz (approved):

PR #815, opened 2025-04-24
Restore high_T timestep namelist options

matthiasfabry:
During #471, when cleaning up the control namelist for deprecated items, I mistakenly removed two timestep options from the namelist. Here I reintroduce them.


pmocz (approved):

Debraheem:
if this is good, let's merge this during the dev meeting/

PR #816, opened 2025-04-28
rename pause to pause_flag since pause is a reserved fortran 77 keyword

pmocz:
It is best practice to avoid using Fortran reserved keywords for variable names. In best case, it causes poor syntax highlighting in the editor and can confuse some linters. In worst case, it may cause unexpected behavior depending on the compiler and language standard down the road (although things work for us)

So I am renaming `pause` to `pause_flag`. I know it can be a pain when inlist variable names are changed, but I think it is worth it.

### DIFF
--- a/astero/test_suite/astero_adipls/inlist_astero_adipls
+++ b/astero/test_suite/astero_adipls/inlist_astero_adipls
@@ -69,7 +69,7 @@
 
       ! top level controls
 
-         !pause = .true.
+         !pause_flag = .true.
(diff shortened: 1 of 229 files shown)


matthiasfabry:
I think this is a fair change, and I'm surprised the compiler even allows this?! Imagine naming a variable "return". Personally I've never used `pause`, only `pause_on_terminate` (and even that only occasionally).

pmocz:
I'm glad it's not a commonly used flag! It can get annoying for users to have variable names changed. But worth it in this case to future-proof

PR #818, opened 2025-05-08
Fixed namelist typo

JakeBHassan:
Changed "lextra" to "lxtra" in the documentation for extending MESA


pmocz:
Awesome find! Thanks, Jake!!

Looks there are some other commits in here that should not be here (looks like they are commits that belong to the 24.08.01 release branch), so I will close this PR and just apply your commit

In the future forking the main branch (our develop branch) instead of a release branch will prevent this issue

PR #819, opened 2025-05-14
exposing some star_utils functions to user

pmocz:
Exposing smoothing and kelvin-helmholtz timescale to users to be usable in custom `run_stars_extras`

### DIFF
--- a/star/private/star_utils.f90
+++ b/star/private/star_utils.f90
@@ -82,7 +82,7 @@
       public :: calc_ptrb_ad_tw
       public :: set_energy_eqn_scal
       public :: get_scale_height_face_val
-      public :: weighed_smoothing
+      public :: weighted_smoothing
       public :: get_kap_face
       public :: get_rho_face
       public :: get_chirho_face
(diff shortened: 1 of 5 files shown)


star/private/star_utils.f90
```diff
@@ -2813,12 +2813,12 @@ subroutine weighed_smoothing(dd, n, ns, preserve_sign, ddold)
...
 
-      end subroutine weighed_smoothing
+      end subroutine weighted_smoothing
 
 
-      subroutine threshold_smoothing (dd, dd_thresh, n, ns, preserve_sign, ddold)
+      subroutine threshold_smoothing(dd, dd_thresh, n, ns, preserve_sign, ddold)
 
-        ! Same as weighed_smoothing, but only smooth contiguous regions where |dd| >= dd_thresh
+        ! Same as weighted_smoothing, but only smooth contiguous regions where |dd| >= dd_thresh
```
warrickball, line 2821: The code looks fine as is but as a low-priority remark, this smells like the two functions could be one, possibly by implementing `dd_thresh` as an optional argument. If absent (`.not. present(...)`), setting `dd_thresh` to +∞ presumably has the same effect (though not sure about NaNs) as not setting `dd_thresh` at all.

pmocz:
> Additionally, eval_kh_timescale can be renamed to e.g. kh_time_scale for consistency.

I made sure the function exposed to users does not have `eval_` in it. But I'll leave the internal function as-in for now because it makes it easier to distinguish between the variable `kh_time_scale` and lots of other internal functions start with `eval_` so at some point we will have to make the style consistent

Issue #814, opened 2025-04-24
Opacity evaluation for cool material (Low T and R values below the limits of the kap_lowT table)


JRGarza:
**Context**  
I am running a simulation in which energy is injected into a stellar envelope to follow the evolution of the star and any expelled material with the Riemann hydro solver.  The opacity tables in use are

  kap_lowT_prefix  = 'lowT_fa05_a09p'
  kap_CO_prefix    = 'a09_co'
  kap_file_prefix  = 'a09'

**Bug description**  
When the expelled material cools below log T = 2.7 (the lower limit of the *lowT_fa05_a09p* tables) MESA calls for a retry because it attempts to evaluate an opacity outside the table’s range (see error output below).  Raising an error for log T < 2.7 in *lowT_fa05_a09p* is correct; however, MESA should never call *kap_lowT* at temperatures below its valid range.  

Inside *kap_eval* the code blends *kap_lowT* with the free‑streaming limit of the prescription by Poutanen (2017).  The blend uses the parameter  
&nbsp;&nbsp;R = ρ / (T/10⁶ K)³  
but, as far as I can tell, the implementation does **not** guarantee that **kap_frac_Compton → 1** as log T approaches log T_min.  Consequently MESA still tries to evaluate *kap_lowT* where it is invalid, or it does not have a table to evaluate at such conditions if **kap_frac_Compton = 0** (I assume this is also incorrect).

**How to reproduce**  
Reproduction is difficult with my exact setup; nevertheless, similar crashes may appear in CCSN test cases that use *u_flag* with no winds or other mass‑removal routines.

In issue #776 @Debraheem mentioned on the *opacity_min* control:

> [...] I've crashed many expanding models because the surface hits the opacity table boundary ...

This seems related.  The *opacity_min* is ineffective here because the opacity (and its derivatives) must exist **before** any floor value can be applied.

**Expected behaviour**  
At these low temperatures the blended fractions should be

  kap_frac_lowT   = 0  
  kap_frac_highT  = 0  
  kap_frac_Type2  = 0  
  kap_frac_Compton = 1  

so that the opacity and its derivatives remain well‑defined.  At present MESA instead returns kap = ‑10^99.

**Version**
 - MESA version: r23.05.1
 - MESA SDK VERSION: mesasdk-x86_64-linux-23.7.3
 - OS: Rocky Linux 9.5 (Blue Onyx)

I notice that the latest MESA version contains no significant changes to *kap_eval* module, so I assume this issue is still present.

[...]

PR #821, opened 2025-05-19
Removing Compton opacities for logR < -8 (and blending with the lowT_kap tables at lowT & towR)

JRGarza:
This PR addresses issue #814 by correcting the conditions under which the Compton opacity prescription is applied. Previously, the prescription was used when either T > T_max (~10^8 K) or R < R_min (~10^-8 g/cm³), leading to an unphysical blending between the Compton and lowT_kap opacity tables and resulting in unrealistically low opacities.

To better reproduce the behavior shown in Figure 2 of Paxton et al. (2011), the updated condition now applies the Compton opacity if either:
- T > T_max from the opacity tables (~10^8 K), or
- R > R_max from the tables (~10^-8 g/cm³)
437164351-1e08d4bf-6774-46c6-8337-b2b9016a6ca5-2

This change avoids the problematic blending and ensures a more realistic transition in the opacity regime.

Additionally, this PR introduces a `plot_regions` option to the `kap/plotter`, which helps visualize the regions in log(T)-log(Rho) space where different opacity sources are used. The attached region plot demonstrates how the updated Compton opacity boundaries now appear in this diagnostic tool.

kap_regions-1


Debraheem:
Very nice! I think you are missing the conductive opacities, which are not shown in your plotter, but are represented by the dashed blue line in the original MESA figure. that represents the approximate point at which the opacities are blended with conductive opacities. Compton opacities are not used at lowT in high density regimes, rather the molecular and atomic tables are interpolated laterally to the right with conductive opacities.

Debraheem:
Screen Shot 2025-05-19 at 11 03 14 AM
For reference, see also the top right panel from figure 2 of [Farag et al. 2024](https://ui.adsabs.harvard.edu/abs/2024ApJ...968...56F/abstract)

Debraheem:
I will look closer at your kap changes tomorrow.

pmocz:
Nice region plotter, @JRGarza ! It's helping me understand what the code is doing

I don't know if the new logic for Compton blending will be needed or fixes your original issue. I'm seeing the opacities remain quite similar if I extend the regions being plotted in `kap/plotter/`.
comparison

There is still a steep drop in opacities (by construction) in the left blue region of the plot. Of course, the code also crashes and returns nan below a minimum temperature of a few hundred Kelvin.

A fix would probably involve doing something different in that left blue region and/or below the min temperature cutoff.

pmocz:
We've found the main issue is not in the blending, but in the `lnfree_e`  values the EOS spits out. I've raised a new issue here: https://github.com/MESAHub/mesa/issues/839  and we will need a different solution. I'm closing this PR for now. It would be wonderful to add your region plotter to the codebase though @JRGarza in a separate PR if you are interested

Issue #824, opened 2025-05-22
pgstar kippenhahn colors are messed up


pmocz:
**Describe the bug**
The Kippenhahn diagram colors in pgstar have been messed up recently, as pointed out by @Debraheem 

**To Reproduce**

Run the `test_suite/20M_pre_ms_to_core_collapse/` problem.

Just the part:

```
cp standard_zams.mod zams.mod
do_one inlist_to_end_core_he_burn_header after_core_he_burn.mod
```

Make sure to set `pgstar_flag = .true.` in `inlist_common`

**Expected behavior**

Behavior looks good in:

commit 74ce895a838ba1d2369e24ea6c3e5c5b75dfe957
Author: pmocz <philip.mocz@gmail.com>
Date:   Tue Apr 22 13:27:23 2025 -0400

    [ci skip] fortitude check --fix on kap/preprocessor .f files

![Image](https://github.com/user-attachments/assets/e4d17913-2846-42eb-b368-19c14eb396b3)

And broke in:

![Image](https://github.com/user-attachments/assets/9f354617-6d77-461a-8d05-74519dc291e6)

commit dc8932fb5d14bc12aef866759cdf2172b26b04b4
Merge: 74ce895a d79f5d58
Author: Vincent Vanlaer <13833860+VincentVanlaer@users.noreply.github.com>
Date:   Wed Apr 23 11:16:54 2025 +0200

    Merge pull request #808 from MESAHub/VincentVanlaer/pgplot-static-colors
    
    Make pgplot color indices static

So the culprit is:

https://github.com/MESAHub/mesa/pull/808

The PR redefines some parameters, which are not consistent:

OLD:
```
colormap_offset      !   46
colormap_last      !   97
colormap_size      !   51
```

NEW:
```
   integer, parameter :: colormap_offset = 46
   integer, parameter :: colormap_length = 101
```

PR #825, opened 2025-05-22
Restore old colormap

pmocz:
This fixes issue: https://github.com/MESAHub/mesa/issues/824

The PR: 

https://github.com/MESAHub/mesa/pull/808

updated our colormap to use the full length 101 sample, whereas previously we used only every other value (length 51).

While the new code it technically correct -- I see no bug in rescaling the indices and applying offsets consistently-- it slightly changes the colors in diagrams users are already very familiar with.

The simplest and cleanest fix is just to store and use the length 51 reduced colormap, since those are the only color values we were previously using anyway


pmocz:
I've tested `test_suite/20M_pre_ms_to_core_collapse/` looks like before again (as described in: https://github.com/MESAHub/mesa/issues/824)

PR #829, opened 2025-05-28
Adding custom static analysis

pmocz:


linters/check_columns.py
```diff
@@ -1,11 +1,16 @@
...
+import sys
 import re
 import glob
 from collections.abc import MutableSet
 
-MESA_DIR = "../"
+try:
+    MESA_DIR = os.environ["MESA_DIR"]
+except KeyError:
+    MESA_DIR = "../"
```
warrickball, line 12: If you like, you can use either of these variants of Python's [`dict.get` method](https://docs.python.org/3/library/stdtypes.html#dict.get), which has an optional return value if the key isn't in the dict. The generic version with a dict would be
```suggestion
MESA_DIR = os.environ.get("MESA_DIR", "../")
```
The `os` module also has its own similar function [`os.getenv`](https://docs.python.org/3/library/os.html#os.getenv), 
```suggestion
MESA_DIR = os.getenv("MESA_DIR", "../")
```
though I have no preference and don't know any reason to prefer one over the either.

linters/check_columns.py
```diff
@@ -456,7 +465,16 @@ def check_profile():
...
-    check_history()
-    check_profile()
+    result1 = check_history()
+    result2 = check_profile()
+
+    result = result1 or result2
+    if result == 0:
+        print("All checks passed.")
+    else:
+        print("Some checks failed.")
```
warrickball, line 479: I would personally write this as
```suggestion
    if result1 or result2:
        print("Some checks failed.")
    else:
        print("All checks passed.")
```
The reversing of the order is because this way I can guarantee not needing any brackets in the logical expression. (I can explain further on request.)

linters/check_columns.py
```diff
@@ -456,7 +465,16 @@ def check_profile():
...
+
 
 if __name__ == "__main__":
-    check_history()
-    check_profile()
+    result1 = check_history()
+    result2 = check_profile()
+
+    result = result1 or result2
+    if result == 0:
```
warrickball, line 476: I believe this is more "Pythonic":
```suggestion
    if not result:
```

linters/check_columns.py
pmocz: Nice! This makes it cleaner :)

Issue #801, opened 2025-03-20
fe56_fake_factor


Debraheem:
Correct fe56_fake_factor.

mathren:
To elaborate: trying to blow up small network progenitors at the onset of core collapse, @AldanaGrichener, @Debraheem,  @AstroBarker, @aurimontem, @johndelker, @evanoconnor, and myself found that `fe56_fake_factor` default value changed by 3 orders of magnitude between ~2015/2016 (cf. [Farmer+16](https://ui.adsabs.harvard.edu/abs/2016ApJS..227...22F/abstract), where it was 1d-4, to 1d-7) and now and this was a surprise to all of us.

This is not a "bug", since this is an arbitrary factor changed arbitrarily, but it should have been better documented. Also, based on our experiments maybe the old value should be preferred as a default, because the compound electron captures occur at high T only, such large change in the rate affects the final `Ye` profile and thus the outcome of explosions.

We should  also doublecheck `min_T_for_fe56ec_fake_factor`

fxt44:
it was changed from 1d-4 to 1d-7 in august 2016 by rob/frank/bill because at the time the ni56 rate changed and fe56 was decaying much sooner than it should.

lots of things "should" have been better documented.  at the time the extreme pace of development far outstripped most attempts to document a rapidly evolving source code base. i still stand by that partitioning of the available resources.

Debraheem:
Understandable, I will do some further testing to see how the approx network compares to larger nets in the current version of mesa.

PR #837, opened 2025-06-26
[skip ci] add details on fe56_fake_factor

mathren:
Addresses issue #801


pmocz:
Thanks for adding clarification to this option, @mathren !

PR #842, opened 2025-07-20
lower overshoot_D_min

Debraheem:
See https://iopscience.iop.org/article/10.3847/2515-5172/adf06d. I am in favor of lowering the default overshoot_D_min if this pr passes testing with no issues.


Debraheem:
testing looks good, except for astero, likely due to a change in the chi^2. 

https://testhub.mesastar.org/lower_overshoot_D_min/commits/0ca9c0a

warrickball (approved):

Issue #843, opened 2025-07-22
`Orbit_win_flag = .true.` not showing Orbit plot (pgbinary)


vincent-bronner:
**Describe the bug**
I tried using pgbinary to show only the Orbit plot, but no plotting window is opened.

**To Reproduce**
1. Take the `binary/work` directory
2. Set `pgbinary_flag = .true.` in the `binary_job` section of the inlist
3. Set `Orbit_win_flag = .true.` in the `pgbinary` section of the inlist

**Expected behavior**
I expect that a plotting window is open showing the Orbit plot, but only the pgstar plots are shown. 

**Version**
 - MESA version: r24.08.1
 - MESA SDK VERSION:  mesasdk-x86_64-linux-24.7.1
 - OS: Ubuntu 22.04

PR #844, opened 2025-07-23
Add the orbit window device number

matthiasfabry:
Addresses #843 found during the MESA 2025 summer school.


pmocz:
Thanks for the quick fix! You can merge when ready

Debraheem:
maybe this is unrelated, but i noticed when i set "Orbit_file_flag = .true.", no such output is produced.

pmocz:
> Orbit_file_flag = .true.

It saves output now as a result of this fix!

PR #857, opened 2025-09-15
fix uninitialized variable in atm t_tau

VincentVanlaer:
The check should be on the value passed into the function, not the converted value which is only computed later.

### DIFF
--- a/atm/private/atm_t_tau_varying.f90
+++ b/atm/private/atm_t_tau_varying.f90
@@ -551,8 +551,8 @@ contains
        return
     end if
 
-    if (dlntau <= 0._dp) then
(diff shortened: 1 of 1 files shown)


warrickball:
Good catch!

LGTM. Tests are [passing](https://testhub.mesastar.org/atm%2Fttau-uninitialized-variable/commits/153df769eb4e6276633621dd56c032def12348d7). The linter is flagging trailing whitespace in a completely unrelated file (`binary/private/binary_disk.f90`), so I'm not worried about it but you're welcome to fix it.

I suggest we rebase and merge here since it's only one commit.

PR #859, opened 2025-09-22
kap: fix out of bounds array initialization

VincentVanlaer:
`num_kap_Zs` is set to 41, which matches the data files, while there are 48 elements in the assignment. This removes the 7 extra elements.


Debraheem (approved):
good catch, I must have made this mistake.

PR #860, opened 2025-09-22
Three component phase separation for ultramassive white dwarfs

mcastrotapia:
I added the files used in Castro-Tapia & Cumming (2025) to implement 3-species fractionation in white dwarf models. This includes the modified version of the phase_separation.f90 file and a directory in star/test_suite called wd_o_ne_3_phase to test the extension. Additionally, among the extensions of the phase_separation.f90, we added the isotope 22Ne to the previous O/Ne fractionation and fixed an inconsistency with the coordinate in mass of the solid core (see Appendix A of Castro-Tapia & Cumming, 2025).


evbauer:
It looks like we're also getting some relatively minor compilation failures: https://github.com/MESAHub/mesa/actions/runs/18114085086/job/51547218693?pr=860

The fortran linter is picking up a few other things (trailing white space, lines that are technically too long, .eq. would be better as ==, etc), but we can fix those later. I'm happy to go ahead with the squash/merge once we get this passing the compilation tests, and then we can continue with a new PR from the testing branch here.

evbauer:
Ah, I see now that we just lost a trailing underscore from `end subroutine update_model_` in the latest commit, so I think compilation should be fine as soon as we get that one character back in there.

PR #863, opened 2025-09-30
Three component phase separation for ultramassive white dwarfs, part 2

evbauer:
Updated code in phase_separation.f90 and a new wd_o_ne_3_phase test suite.


mcastrotapia:
Thanks for the fixes and for reviewing the PR! I will merge the branch into my fork, and I'll be attentive to comments/questions and updates.

star/private/phase_separation.f90
```diff
@@ -113,137 +114,216 @@ subroutine do_2component_phase_separation(s, dt, components, ierr)
...
                s% eps_phase_separation(k) = s% energy(k)
             end do
-
             ! loop runs outward starting at previous crystallization boundary
             do k = kstart,1,-1
                ! Start by checking if this material should be crystallizing
                if(s% phase(k) <= eos_phase_boundary) then
-                  s% crystal_core_boundary_mass = s% m(k+1)
-                  exit
+                  if (s% crystal_core_boundary_mass>s% m(k+1)) then
```
evbauer: What is this meant to catch? It seems there must be some subtle bug here I was missing.

star/private/phase_separation.f90
```diff
@@ -40,41 +39,48 @@ subroutine do_phase_separation(s, dt, ierr)
...
             call do_2component_phase_separation(s, dt, 'ONe', ierr)
+         else if(s% phase_separation_option == '3c') then
+            call do_2component_phase_separation(s, dt, '3c', ierr)
          else
             write(*,*) 'invalid phase_separation_option'
             stop
          end if
       end subroutine do_phase_separation
 
       subroutine do_2component_phase_separation(s, dt, components, ierr)
```
evbauer: If we keep the current code structure, we'll need to rename this subroutine to something other than "do_2component...", since it is now explicitly being used to do some 3 component cases as well. I had another branch a while ago that was attempting a different branch for 3-component distillation, but that is now somewhat stale and never got to a point of being ready to merge in. I'll need to reconcile the naming scheme somehow...

star/private/phase_separation.f90
```diff
@@ -40,41 +39,48 @@ subroutine do_phase_separation(s, dt, ierr)
...
          s% phase_sep_mixing_mass = -1d0
          s% eps_phase_separation(1:s%nz) = 0d0
-
-         if(s% phase(s% nz) < eos_phase_boundary) then
-            s% crystal_core_boundary_mass = 0d0
-            return
+         if(s% phase(s% nz) < eos_phase_boundary) then !!! prevent to move the core size inwards if the core is suddently "melted" leaving everything liquid under phi<0.9
+            if (s% crystal_core_boundary_mass>0d0)then
+               s% crystal_core_boundary_mass=s% crystal_core_boundary_mass
+               return
```
evbauer: I see why you might need something like this, but I'm not sure this is the best approach in general. One alternative I've experimented with is just redistributing the phase separation heating over a significant fraction of the interior. That way you avoid transient local heating spikes that might melt the core, but still get the net right amount of energy deposited in the interior, which seems to work for capturing the effect on cooling rate. @mcastrotapia do you think that could be a viable alternative here?

star/private/phase_separation.f90
```diff
@@ -83,18 +89,14 @@ subroutine do_2component_phase_separation(s, dt, components, ierr)
...
             end if
          end do
-
          XC = s% xa(net_ic12,k_bound)
          XO = s% xa(net_io16,k_bound)
-         XNe = s% xa(net_ine20,k_bound)
-         ! Check that we're still in C/O or O/Ne dominated material as appropriate,
-         ! otherwise skip phase separation
-         if(components == 'CO'.and. XO + XC < 0.9d0) return
-         if(components == 'ONe'.and. XNe + XO < 0.8d0) return  ! O/Ne mixtures tend to have more byproducts of burning mixed in
```
evbauer, line 93: I believe we need to bring these lines back. Is there an argument for deleting them?

They shouldn't affect the new 3-componenent functionality that we're trying to add here, and I'd like to keep the behavior of the 2-component options unchanged unless there is a compelling reason to change them.

star/private/phase_separation.f90
```diff
@@ -359,42 +715,35 @@ subroutine update_model_ (s, kc_t, kc_b, do_brunt)
            stop
         end if
         s%fix_Pgas = .false.
-
         ! Update opacities across cells kc_t:kc_b (this also sets rho_face
         ! and related quantities on faces kc_t:kc_b)
         call set_micro_vars(s, kc_t, kc_b, &
-             skip_eos=.TRUE., skip_net=.TRUE., skip_neu=.TRUE., skip_kap=.FALSE., ierr=ierr)
+             skip_eos=.TRUE., skip_net=.TRUE., skip_neu=.TRUE., skip_kap=.TRUE., ierr=ierr)
```
evbauer: I think we need to revert this change. This is now skipping the opacity update that is the point of this call. In fact, as written, I think this call to `set_micro_vars` skips everything and does nothing at all.

star/private/phase_separation.f90
```diff
@@ -252,103 +332,379 @@ subroutine mix_outward(s,kbot,min_mix_zones)
...
+            if (is_bad(deltax1_sob_f(k,j,i))) then
+            write(*,*) 'deltax1_sob_f', i, j, k, deltax1_sob_f(k,j,i)
+            end if
+         end do
+      end do
+   end do
+   call interp_evbicub_db( &
+            x1_, x2_, x1_l, num_x1, x2_l, num_x2, &
+            ilinx, iliny, deltax1_sob_f1, num_x1, ict, fval, ier)
+    dx1_=fval(1)  ! delta_x1 from 2d interpolation
```
evbauer: Indentation of this block should be fixed. (Just a minor aesthetic note.)

star/test_suite/wd_o_ne_3_phase/src/run_star_extras.f90
evbauer, line 1: Note to self for later: update run_star_extras, rn script, etc to make this a fully-fledged test case and integrate it into do1_test_source for the test suite.

star/private/phase_separation.f90
evbauer: There's another block later on with the same indentation problem as well.

star/private/phase_separation.f90
```diff
@@ -113,137 +114,216 @@ subroutine do_2component_phase_separation(s, dt, components, ierr)
...
                s% eps_phase_separation(k) = (s% eps_phase_separation(k) - s% energy(k)) / dt
             end do
             s% eos_rq% Skye_use_ion_offsets = save_Skye_use_ion_offsets
             s% need_to_setvars = .true.
          end if
-
          ierr = 0
       end subroutine do_2component_phase_separation
 
       subroutine move_one_zone(s,k,components)
```
evbauer: The `move_one_zone` subroutine has developed a lot of different branches that I think make it rather confusing. I can see why it developed into what it is, but I'd probably like to refactor this into two separate routines: `move_one_zone_for_2comp` that basically just retains the old functionality (plus the nice addition of accounting for Ne22), and `move_one_zone_for_3comp` that implements all the new 3-component cases.

star/private/phase_separation.f90
mcastrotapia: I think I was trying to prevent the core from moving inwards in case the extra heating suddenly changes the energy in both phases and the phase parameter. However, I think it's not necessary since the lines where I added the following:

`if (s% phase(s% nz) < eos_phase_boundary) then !!! prevent to move the core size inwards if the core is suddently "melted" leaving everything liquid under phi<0.9`

            `if (s% crystal_core_boundary_mass>0d0)then
               s% crystal_core_boundary_mass=s% crystal_core_boundary_mass
               return
            else 
               s% crystal_core_boundary_mass = 0d0
               return
            end if
         end if`

 must be enough for that

star/private/phase_separation.f90
mcastrotapia: Okay, I agree, if you want me to help with that, please let me know :)

star/private/phase_separation.f90
mcastrotapia: I think it is a good alternative. At some point, I tried something similar, but I didn't have success, and I ended up doing this because it seemed easier and faster to implement.

star/private/phase_separation.f90
mcastrotapia: I deleted them when trying to let the code decide for itself if doing C/O or O/Ne in the case of 2-component phase separation, this is because in ultramassive WD models the core is usually O/Ne, but towards the outer layers there is a C/O dominated zone before reaching the H/He atmosphere (for example, Figure 2 of [Camisassa et al., 2019](https://www.aanda.org/articles/aa/pdf/2019/05/aa33822-18.pdf)).  In the 3-component, I was also letting the code check which were the 3 most abundant elements in the liquid zone to decide which phase diagram must be used.

star/private/phase_separation.f90
mcastrotapia: I agree, I was mostly having convergence problems with this when using a small timestep and (I think) with the heating in the solid phase for such timesteps. That's why it would be good to implement the heating term as you suggested in one of the previous comments

star/private/phase_separation.f90
mcastrotapia: Okay :)

star/private/phase_separation.f90
mcastrotapia: That sounds good!

evbauer:
We'll also need to add some documentation of the new options in controls.defaults.

star/private/phase_separation.f90
evbauer: The bigger concern here is when you get all the way out to the interface with the He envelope, at which point I want to make sure phase separation ends. We can return to make a final decision on this once I refactor into separate 2-component and 3-component logic.

star/private/phase_separation.f90
evbauer: Ok, I introduced the heat redistribution in https://github.com/MESAHub/mesa/pull/863/commits/5519a13a3f8ac3a89fac035ce66d67ec138f9a7f. Let's see if that works instead.

star/private/phase_separation.f90
```diff
@@ -322,14 +457,298 @@ real(dp) function blouin_delta_xne(Xin)
...
+        ! setup interpolation table for tau sob eta
+        if (components=='CONe') then
+           open(unit=iounit, file='CONe_deltaO.dat', action='read',status='old')
+        else if  (components=='NeOMg') then
+           open(unit=iounit, file='NeOMg_deltaO.dat', action='read',status='old')
+        else if  (components=='ONeNa') then
+           open(unit=iounit, file='ONeNa_deltaO.dat', action='read',status='old')
+        else if  (components=='COMg') then
+           open(unit=iounit, file='COMg_deltaMg.dat', action='read',status='old')
+        end if
```
VincentVanlaer: Use newunit:


```suggestion
        ! setup interpolation table for tau sob eta
        if (components=='CONe') then
           open(newunit=iounit, file='CONe_deltaO.dat', action='read',status='old')
        else if  (components=='NeOMg') then
           open(newunit=iounit, file='NeOMg_deltaO.dat', action='read',status='old')
        else if  (components=='ONeNa') then
           open(newunit=iounit, file='ONeNa_deltaO.dat', action='read',status='old')
        else if  (components=='COMg') then
           open(newunit=iounit, file='COMg_deltaMg.dat', action='read',status='old')
        end if
```

star/private/phase_separation.f90
```diff
@@ -322,14 +457,298 @@ real(dp) function blouin_delta_xne(Xin)
...
+        ! setup interpolation table for x1 x2 dx1
+        if (components=='CONe') then
+           open(unit=iounit, file='CONe_deltaC.dat', action='read',status='old')
+        else if  (components=='NeOMg') then
+           open(unit=iounit, file='NeOMg_deltaMg.dat', action='read',status='old')
+        else if  (components=='ONeNa') then
+           open(unit=iounit, file='ONeNa_deltaNa.dat', action='read',status='old')
+        else if  (components=='COMg') then
+           open(unit=iounit, file='COMg_deltaC.dat', action='read',status='old')
+        end if
```
VincentVanlaer: Use newunit:


```suggestion
        ! setup interpolation table for x1 x2 dx1
        if (components=='CONe') then
           open(newunit=iounit, file='CONe_deltaC.dat', action='read',status='old')
        else if  (components=='NeOMg') then
           open(newunit=iounit, file='NeOMg_deltaMg.dat', action='read',status='old')
        else if  (components=='ONeNa') then
           open(newunit=iounit, file='ONeNa_deltaNa.dat', action='read',status='old')
        else if  (components=='COMg') then
           open(newunit=iounit, file='COMg_deltaC.dat', action='read',status='old')
        end if
```

evbauer:
Converting the plain text `.dat` files to hdf5 has saved quite a bit of space. The previous files took up about 250 MB, and the hdf5 files now take up only 5.4 MB. This will probably save some time on file i/o as well. However, I'm now getting a crash in a call to `set_micro_vars` after 3 component phase separation gets underway in the test case, so I'll need to check if I've translated everything properly.

star/private/phase_separation.f90
evbauer: Happily, converting to hdf5 made this even cleaner. +1 for @rhdtownsend's forum HDF5 interface!

star/test_suite/wd_o_ne_3_phase/README.rst
```diff
@@ -0,0 +1,23 @@
...
+status.  Supplementary material like plots, plotting scripts, etc.,
+should go in a subdirectory ``docs``.
+
+For tests that are likely to serve as examples of MESA usage, provide
+additional information about key options or important caveats.  This
+assists users who are adapting the test case to their own science.
+
+The README file should end with a line that describes when the most
+recent significant changes to this test case occurred.  This gives
+both users and developers a sense of how "fresh" the information is.
```
evbauer: @mcastrotapia could you fill this in or send me anything you'd like to be included here?

evbauer:
I believe we're finally ready on this one! (assuming the tests pass)

Perhaps we should add a changelog entry?

star/private/phase_separation.f90
```diff
@@ -40,26 +41,35 @@ subroutine do_phase_separation(s, dt, ierr)
...
          else if(s% phase_separation_option == 'ONe') then
-            call do_2component_phase_separation(s, dt, 'ONe', ierr)
+            call separate_mix_and_heat(s, dt, 'ONe', ierr)
+         else if(s% phase_separation_option == '3c') then
+            call separate_mix_and_heat(s, dt, '3c', ierr)
          else
             write(*,*) 'invalid phase_separation_option'
             stop
          end if
+
```
VincentVanlaer, line 55: You will need a check for ierr here, otherwise `smooth_eps_phase_sep` will hide the error. Since neither functions actually set ierr to anything else than zero, the ierr parameter could also be removed.

star/private/phase_separation.f90
```diff
@@ -322,14 +464,351 @@ real(dp) function blouin_delta_xne(Xin)
...
+        ! setup interpolation table for tau sob eta
+        if (components=='CONe') then
+           filename = trim(phase_sep_data_dir) // '/CONe_deltaO.h5'
+        else if  (components=='NeOMg') then
+           filename = trim(phase_sep_data_dir) // '/NeOMg_deltaO.h5'
+        else if  (components=='ONeNa') then
+           filename = trim(phase_sep_data_dir) // '/ONeNa_deltaO.h5'
+        else if  (components=='COMg') then
+           filename = trim(phase_sep_data_dir) // '/COMg_deltaMg.h5'
+        end if
```
VincentVanlaer, line 617: I don't know how often these data reads are called. If it is on the order of 1 to 10s of times per run its probably fine. Otherwise we might want to keep this around in memory. This isn't a blocker for me to merge in since it is a new feature anyway (and disabled by default?) and leave it for a rainy day.

star/defaults/controls.defaults
```diff
@@ -7563,6 +7566,17 @@
 
     do_phase_separation_heating = .true.
 
+      ! smooth_phase_separation_heating
+      ! ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+      ! if true, redistribute all heating associated with phase
+      ! separation evenly throughout the inner half of the white dwarf
+      ! (prevents unphysical localized spikes at small time steps)
```
VincentVanlaer: Would be good to document the choice for the inner half (could be arbitrary, but I think it would nice for the user to know why).

VincentVanlaer (approved):
Left some small comments, which can be resolved easily or can be postponed. I think you should add a changelog entry, this feature certainly deserves it!

star/private/phase_separation.f90
evbauer: Good point! In principle you're definitely right that it would be better to keep these around in memory. I think they get called 100s of times per run. In practice reading these small files doesn't seem to be much of a bottleneck right now, especially after converting them to hdf5, so I think we'll leave it for a rainy day for now.

evbauer:
Thanks for the suggestions @VincentVanlaer! I've addressed them (other than the loading into memory suggestion, which I'd like to leave for later), and I added a changelog entry. I think this should be ready to merge at last.

I've had a bit of trouble getting the new test case to be robust to small changes in starting model generation when running all inlists, which is why there are several [ci optional] commits here while I was working that out. I've added a few things that will hopefully help it be more robust, but I'll keep an eye on the testing for a while too.

Issue #889, opened 2025-11-22
make planets crashing when Psurf approches 1  dyne/cm^2


Debraheem:
MESA user (Sho Shibata) has identified an issue in their planet atmosphere models where the model crashes when Psurf approaches 1 dyne/cm^2, whether from above or below. 

This could be an eos issue, however the same issue occurs whether adopting OPAL or ideal eos, and I cannot find any eos inconsistency using the eos plotter functionaly for the region of parameter space where the model crahes. 

The crash is associated with a sudden large Energy error and work terms in the energy eqn:

I paste below the (often) offending cell which tends to be right below the surface: 

 work_p1_ad  -9.5507730832923167E+032  0.26828140753901653     
 work_00_ad  -9.5350237754495172E+032  0.26841898552505711     
 dwork   1.5749307842799536E+030

                                                   scal           3    1.0479637705272757D-15
                                               residual           3    8.5871662670425371D+02
                                           sources*scal           3    2.2118570989433915D-45
                                            -dL_dm*scal           3    1.2719366498810434D+05
                            -d_turbulent_energy_dt*scal           3    0.0000000000000000D+00
                                         -dwork_dm*scal           3   -1.2633494836603357D+05
                                           -dke_dt*scal           3    0.0000000000000000D+00
                                           -dpe_dt*scal           3    4.5208160783135141D-06
                                                  gradT           3    1.8593330599337089D-02
                                                opacity           3    8.7632112625897787D-08
                                                   logT           3    2.4771212547196630D+00
                                                 logRho           3   -1.0865549844059307D+01
                                                      X           3    7.1938774594784982D-01
                                                      Z           3    0.0000000000000000D+00

 work_p1_ad  -9.1008358599149637E+032  0.25772003705015861     
 work_00_ad  -9.0853603765438656E+032  0.25785633094895571     
 dwork   1.5475483371098058E+030

                                                   scal           3    1.0479637705272757D-15
                                               residual           3    6.3416381491038203D+02
                                           sources*scal           3    2.1248105658035978D-45
                                            -dL_dm*scal           3    1.2477259656841007D+05
                            -d_turbulent_energy_dt*scal           3    0.0000000000000000D+00
                                         -dwork_dm*scal           3   -1.2413843275791570D+05
                                           -dke_dt*scal           3    0.0000000000000000D+00
                                           -dpe_dt*scal           3    4.3060816502279115D-06
                                                  gradT           3    1.7022163525980977D-02
[...]

PR #890, opened 2025-11-24
fix surface residuals blowing up to nan

pmocz:
fixes issue #889

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -153,6 +153,7 @@ Fixed a bug in ``other_mlt_results`` which resulted in the pre-ms model builder 
 
 Removed unused parameters: ``fp_error_limit``, ``fp_min``, ``ft_error_limit``, ``ft_min``, ``retain_fallback_at_each_step``.
 
+Fixed bug in residual calculation near P = 1, or T = 1
 
 .. note:: Before releasing a new version of MESA, move `Changes in main` to a new section below with the version number as the title, and add a new `Changes in main` section at the top of the file (see ```changelog_template.rst```).
 


Debraheem:
The conundrum here is I believe neither of these tests use the fixed P or fixed T surface BCs. Philip's fix might have exposed a separate fpe issue?

Debraheem:
Whoops i was confused, please ignore my previous statement. I have a potential fix, will push soon.

pmocz:
Looking at the blame, looks like MESA used to behave like the fix I proposed: https://github.com/MESAHub/mesa/commit/0b386fb3b90aacf044464de93e0f70eeee65fb69#diff-b19cc691aa401fbf61b1768ec894f2fbf16855fff010f56a8ad95194827e72e1L2007

You'll have to manually go to line 2007 of the hydro_eqns.f90 file to see it in the link above. We have the option of scaling the residual terms if we want. Looks like the old mesa just used to use a scaling factor of 1 though

I'm not sure what is causing the tests to fail. My guess is they are sensitive tests, and they might pass if we tweak some settings

Debraheem:
ahh, you're probably right, they are just sensitive. I can tweak them. But in the mean time maybe we can get the best of both worlds with my patch that just floors the scale when it drops below 1d0? 

I attach a plot below of the original form in the current main, the absolute form Philip changed to, and then my scaled floor patch. Maybe that's enough for now? Anything more and I can touch up the test cases...?

original form = lnX_bc/lnX1_ad - 1d0
absolute form = lnX_bc - lnX1_ad
floored scale form = (lnX_bc - lnX1_ad)/scale ; scale = max(1d0, lnX1_ad)

the scaled floor form reduces to absolute in the lowT low P limit and returns to original form in the highT highP limit.

residuals

Debraheem:
In reality, we don't really need to mess with the T BC because the eos stops at 1d2 K, but assuming the eos changes in the future, the scaled form will prevent running into a singularity.

PR #892, opened 2025-11-29
Pmocz/bugfix/nan

Debraheem:
Hello, I forgot, but it's probably smart to fix the scale over solver iterations by using start of step values. If this passes testing, I think we can merge. @warrickball @pmocz let me know if you're alright with this change


pmocz:
Yeah, that does make sense to me. I will merge (only the already-broken radiative levitation is failing)

Issue #80, opened 2020-08-17
Helium flash when using the Ledoux criterion 


annethoul:
Cannot pass helium flash when using the Ledoux criterion. Lots of retries (retry: max residual jumped -- give up in solver), stops due to timestep ( stopping because of problems dt < min_timestep_limit).
This is true for masses around 1 Msun. Other masses not tested. 

This is independent of the algorithm used for the convective boundaries. No problem when using the Schwarzchild criterion provided it is used from the TAMS onwards. This is strange since both criteria should give essentially identical models when using the CPM algorithm!

It is possible to go through helium flash with the Ledoux criterion by using the control 
convergence_ignore_equL_residuals = .true.
This is not convenient for the "one inlist fits all" aficionados.

Version tested: 14356. But note that this is an ancient issue, mentionned in a long mesa-dev thread initiated by me under the subject "Problems for He flash and Ledoux criterion" in July 2018 (version 10881).
I am assigning this to myself but considering my debugging success rate, it might be nice if others look also into it ;-)

warrickball:
I've just been forced back to this problem myself, exactly because I have a 1.5 Msun model that I need to get through the flash *with* Ledoux, and I'm only now digesting the inlist @adamjermyn shared and the animation of the 1.5 Msun star's evolution.

First, in the video, what happens around model 400, around the base of the RGB? The HR track does some loops that I think could be described as "crazy", taking Teff to ~2500 K before continuing.

Second, there's a *lot* going on in that inlist, including thermohaline mixing, hydrodynamics and convective premixing. Do you have any idea which of these are necessary to get through? I've done a run with `delta_lgL_He_limit = 3d-3` in isolation but that didn't get me through the flash. I *have* been using `varcontrol_target = 1d-4` until now but I might relax that to `1d-3` to see what happens. Basically, can we get through without switching on any "extra" physics?

To be honest I'm just non-systematically flipping all the switches and turning all the dials in the inlist and nothing's working. In the absence of progress, I think I'll move to activating `convergence_ignore_equL_residuals` during the flash, and consider re-opening this.

adamjermyn:
Not sure about the first one.

Thermohaline is necessary to match observed Li abundances. These settings run (at least for 1Msun) just fine without it. So feel free to chuck that part.

You can probably get rid of min_D_mix as well. That just prevents mesh refinement from producing garbage in the brunt profile.

I have gotten models to run without hydrodynamics, but don't have an inlist handy. Try it and see!

[...]

PR #897, opened 2025-12-07
EquL solver stability improvement

### LINKED ISSUE #563: Inability to get through helium flash for 1.05 - 1.4 Msolar
MESA version 23051

Comparing this inlist (attached as .txt) for a science case against `inlist_1.3M_ms_high_Z` from the `test_suite` case of the same name, I have systematically removed each difference in physics (replace with `basic.net`, use standard opacities, etc) to try to isolate the reason the science inlist fails. I have not found any change that enables successful passage through the helium flash 
[inlist_issue.txt](https://github.com/MESAHub/mesa/files/12129491/inlist_issue.txt)

However, introduction of the following control parameters _does_ enable passage through the he flash:
`  convergence_ignore_equL_residuals = .true.
   relax_use_gold_tolerances = .true.
` 

Despite confirmation from other devs that this control is needed to get models in this regime through the he flash (as of some point after release 10398 and the solver controls overhaul),
there is no documentation to accompany ` convergence_ignore_equL_residuals = .true.`

I think a few things need to happen:
1. the effect of this and similar controls needs to be documented. As far as I can tell, the meaning of this control is to shut off consideration of discrepancies in the luminosity equation entirely (?). 
   a. Corollary: is that safe??
   
2. It needs to be made loud and clear that this setting is necessary for (most) AGB models. I advocate for introducing it in `inlist_1.3M_ms_high_Z` even though it is _not_ required in that case, just so users are aware of it. They will almost certainly consult this `test_suite` case when trying to figure out their own convergence issues for AGBs. It has been pointed out to me that this setting is in the `1M_pre_ms_to_wd` `test_suite` case, but that is not the first place I personally would look for usable AGB settings.

3. We need a test case for AGBs (@giulia and I are doing this, stand by)

4. Ideally, we need a better solution than (even a well-documented) magic flag for dealing with this. If the era of tighter tolerances made getting through the he flash impossible without telling the solver to ignore one of the stellar structure equations(?)...that seems like something we should fix.  

end rant!

Debraheem:
This is either a legendary fix or a blunder. I believe I've addressed the issue causing us to have to resort to using `convergence_ignore_equL_residuals = .true.` . 

If you're curious look at the commit. 

`edit *** residual scaling seems to have no effect and was removed.`Just light some slight tweaks to the solver tolerances for when to give up and changing `convergence_ignore_equL_residuals = .true.` across the entire test_suite to `.false.`, and changed the MESA default for `use_gold2_tolerances = .true.`, as we were stuck using gold_tolerances before. All the models seem fine now, except the two planets test_cases, which already resorted to this, and the pisn test case which uses a different form of the T-gradient equation anyway. 

This fixes issue like https://github.com/MESAHub/mesa/issues/563. With this pr, I can run through the entire He flash with `convergence_ignore_equL_residuals = .false.` and gold2_tolerances. 

see also old issues: https://github.com/MESAHub/mesa/issues/80, https://github.com/MESAHub/mesa/issues/490

For example the 1M_pre_ms_to_wd test_suite now completes the he flash with LogE_err < -5, roughly 3 orders of magnitude lower than previous commits. (might just be because i switched from eps_grav to dedt form of E-equation)

(I also made some improvements to the make_co_wd test_case, substantially improving its robustness, as it was extremely sensitive)

The largest caveat is my setting of `make_gradr_sticky_in_solver_iters = .true.` by default. This was already on all the massive star cc test_cases and some others. This solver hack waits until the 4th solver iteration, and then after that point, if the zone was convective and becomes radiative (not the opposite), the solver will hold gradr fixed to prevent the sign from flipping again. In practice it has almost no observable effect outside of extremely short timescales, when burning (dlnE_dt) is competing with gradT (from mlt) through the T-gradient equation. We can set this back to `.false.` to see where failures show up, and in practice it could remain `.false.`. I personally like the idea of keeping it on by default, but I will submit a test commit with it set back to false to see if it has any impact and where. 

please verify my changes, give the pr a try. 

This pr still needs to address the photo issue in magnetic breaking, likely related to some Spruit Taylor initialization bug.

### DIFF
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -8776,7 +8776,7 @@
 
       ! ::
 
-    max_corr_jump_limit = 1d6
+    max_corr_jump_limit = 1d12
 
 
       ! resid_norm_jump_limit
@@ -8796,7 +8796,7 @@
 
       ! ::
 
-    max_resid_jump_limit = 1d10
+    max_resid_jump_limit = 1d12
 
 
       ! convergence_ignore_equL_residuals
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -8796,7 +8796,7 @@
 
       ! ::
 
-    max_resid_jump_limit = 1d6
+    max_resid_jump_limit = 1d10
 
 
       ! convergence_ignore_equL_residuals
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -2369,7 +2369,7 @@
 
       ! ::
 
-    make_gradr_sticky_in_solver_iters = .true.
+    make_gradr_sticky_in_solver_iters = .false.
     min_logT_for_make_gradr_sticky_in_solver_iters = 1d99
 
 
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -8776,7 +8776,7 @@
 
       ! ::
 
-    max_corr_jump_limit = 1d12
+    max_corr_jump_limit = 1d6
 
 
       ! resid_norm_jump_limit
@@ -8796,7 +8796,7 @@
 
       ! ::
 
-    max_resid_jump_limit = 1d12
+    max_resid_jump_limit = 1d10
 
 
       ! convergence_ignore_equL_residuals
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -2369,7 +2369,7 @@
 
       ! ::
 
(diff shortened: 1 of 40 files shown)


Debraheem:
This pr successfully showed that with sticky gradr and some tolerance changes, the test_suite can be made to pass with gold2 on. 

However, I am working on broader changes on separate branches and I have already added the solver tweaks into main in a separate pr, so I will close this pr for now. I will follow up with individual prs updating some of the specific test cases touched here to make them more robust, and a potential T limit to toggle sticky gradr. We can consider tightening to gold2 as a default at a later date.

PR #901, opened 2025-12-24
Update formatting for code literals/blocks (docs)

abhro:
- Add backticks for monospace formatting
- Add yaml/fortran syntax highlighting indicators to rst docs
- Remove paragraph text from code block and separate out formatting

### DIFF
--- a/docs/source/using_mesa/best_practices.rst
+++ b/docs/source/using_mesa/best_practices.rst
@@ -107,81 +107,95 @@ First, copy the default ``history_columns.list`` to your work directory
 
 and modify your local ``history_columns.list``
 
-.. code-block:: console
 
-  add
+* add
 
-      mixing_regions 20
-      burning_regions 20
+  .. code-block:: fortran
 
- change
+        mixing_regions 20
+        burning_regions 20
 
-      !log_center_T ! temperature
-      !log_center_Rho ! density
+* change
 
- to
+  .. code-block:: fortran
 
-      log_center_T ! temperature
-      log_center_Rho ! density
-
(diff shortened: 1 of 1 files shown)


VincentVanlaer:
They way I understood this was that essentially the comments form an rst file and hence can use all the syntax. Perhaps I am just too used to seeing ReST/Markdown in code comment blocks for all sorts of documentation generators. I am happy to just revert those added back ticks if it makes to more annoying to read.

abhro:
> They way I understood this was that essentially the comments form an rst file and hence can use all the syntax. Perhaps I am just too used to seeing ReST/Markdown in code comment blocks for all sorts of documentation generators. I am happy to just revert those added back ticks if it makes to more annoying to read.

Same here! I still do view config files and such in a text editor, and while backticks are more for the parser, they do add a bit of visual separator for names, so the syntax works on both sides of raw file and parsed docs. But also, that probably varies person to person, so a revert would be more than understandable if it leads to accessibility issues.

warrickball:
FWIW I take back my original comment:

> They way I understood this was that essentially the comments form an rst file and hence can use all the syntax. 

@VincentVanlaer is right: the intention *is* that these are basically ReST. I didn't properly read the [docs that I linked](https://docs.mesastar.org/en/latest/reference/format.html) ( :facepalm: ), which say

> After leading spaces and ! are stripped, this file should be valid reStructuredText (reST) so that it can rendered in the Sphinx docs. This file is written in the format it describes.

I also don't really mind which way we go but we should be consistent across the docs, and might want a separate discussion about what we want to do.

PR #907, opened 2025-12-27
Make the use of some `null` functions more consistent

VincentVanlaer:


Debraheem:
Is this just a simplification so that users can call the public star_data functions for making a pgstar decorator, as opposed to using a null other_* function? Is this easy for a random user to do? We generally point users to use the other_* functions, so  we just want it to be clear how one adds their own decorator. I'm not sure if it's good or bad to do this. We don't want to hide this from users, and I don't know anyone that looks inside star_data/public, but maybe we can advertise it somewhere?

PR #909, opened 2026-01-07
Features/nialljmiller/customcolors update0

nialljmiller:
- Moved data from colors/data to data/colors_data
 - Precalculate zero points on module load
 - SED data is cached at module load to remove file IO

### DIFF
--- a/star/test_suite/custom_colors/README.rst
+++ b/star/test_suite/custom_colors/README.rst
@@ -9,7 +9,7 @@ This test suite was tested against SDK 25.12.1
 This test suite case demonstrates the functionality of the MESA ``colors`` module, a framework introduced in MESA r25.10.1 for calculating synthetic photometry and bolometric quantities during stellar evolution.
 
 What is MESA colors?
-=======================
+====================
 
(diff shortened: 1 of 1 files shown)


colors/private/shared_funcs.f90
```diff
@@ -1,561 +1,536 @@
...
+      ! Perform trapezoidal integration
+      DO i = 1, n - 1
+         sum = sum + 0.5*(x(i + 1) - x(i))*(y(i + 1) + y(i))
+      END DO
 
+      result = sum
+   END SUBROUTINE trapezoidalintegration
 
+   SUBROUTINE SimpsonIntegration(x, y, result)
+      INTEGER, PARAMETER :: DP = KIND(1.0D0)
```
warrickball: Do we need to redefine `dp` here? I presume not, since it's been imported from `const` at the start of the module.

warrickball:
```suggestion
```

nialljmiller: Yep, DP is already provided by const. I’ve committed your suggestion. Thank you.

pmocz (approved):

PR #916, opened 2026-01-29
memleaks?

pmocz:
adding missing deallocates .. need to wait for tests to pass

I think modern fortran does handle this automatically to prevent memleaks, but explicit deallocates still help with Memory Fragmentation


pmocz:
You're right. I think we don't actually need these. I'll close the PR

PR #939, opened 2026-03-28
Colors Bugfix: use Fe/H instead of zbase

### LINKED ISSUE #938: Metallicity mismatch in colors module
Hi, I attempted to use the colors module to produce a spectrum for a zero-metallicity star. However, when I manually produced the spectrum from ATLAS models (using the same Teff and log(g)), I noticed the MESA colors one much more closely matched a solar-metallicity model than it did a low-metallicity model.

Image

I believe this may be a bug in the colors module. The `data_for_colors_history_columns` function in `mesa_dir/star/private/history.f90` uses Zbase, and passes it through several functions. It eventually reaches the `find_interval` function in `mesa_dir/colors/private/hermite_interp.f90`, which compares it to z_grid. If my understanding is correct, z_grid may refer to [Fe/H], whereas Zbase is the metal fraction. Consequently, setting Zbase to 0 would match it with [Fe/H] = 0 (i.e. solar metallicity).

If this is indeed the problem, Zbase would have to be converted to [Fe/H] first. There is also a check for `metallicity >= 0` in the `data_for_colors_history_columns` function of `mesa_dir/colors/private/colors_history.f90`, so that may also need to be changed if it uses [Fe/H].

Debraheem:
This pr addresses the colors bug motivated by Jake Hassan's investigation in https://github.com/MESAHub/mesa/issues/938. 

I think the main issue was that the colors module was using Zbase instead of the photospheric metallicity for interpolating the atmosphere tables. The solution was to switch to using metallicity, and add a control for the default reference metallicity used in the computation of the atmosphere tables, which i believe is gs98 for Atlas9. I've added docs for this change as well, but a changelog entry is still necessary, and potentially a "known bugs" entry. 

https://github.com/MESAHub/mesa/blob/9ee3f0a72521c70e8ea5a459ef696177950d132c/star/private/history.f90#L270-L271

And then in another location, negative metallicities were being rejected (I think since Z was being used instead of [Fe/H])

https://github.com/MESAHub/mesa/blob/9ee3f0a72521c70e8ea5a459ef696177950d132c/colors/private/colors_history.f90#L115

@nialljmiller , @mjoyceGR  perhaps you could sweep through these changes and see if they make sense?  

I've done some preliminary testing with a 5 Msun model that i stop on the pre-ms when it hits 5000K, for five different metallicities.  Below I show the SED normalized and not normalized to illustrate that the colors module is working, and i did this test with type_2_opacities = .true., the mesa default. 

metallicity_sed_comparison

metallicity_sed_comparison_normalized

One concern raised by this investigation is: what do we do when there is no H? Obviously the current atm tables are H atmosphere tables, but we might want to think about a bette fall back for no H than adopting the lowest metallicity table in the grid? I'm not sure. 

If this passes testing, we can merge it into the r25.12.1_colors_updated branch as well, which will ultimately becoming r26.3.1, or probably r26.4.1 now.

### DIFF
--- a/colors/README.rst
+++ b/colors/README.rst
@@ -22,7 +22,7 @@ How does the MESA colors module work?
 
 The module operates by coupling the stellar structure model with pre-computed grids of stellar atmospheres.
 
-1.  **Interpolation**: At each timestep, the module takes the star's current surface parameters—Effective Temperature (:math:`T_{\rm eff}`), Surface Gravity (:math:`\log g`), and Metallicity ([M/H])—and queries a user-specified library of stellar atmospheres (defined in ``stellar_atm``). It interpolates within this grid to construct a specific Spectral Energy Distribution (SED) for the stars current features.
+1.  **Interpolation**: At each timestep, the module takes the star's current surface parameters, Effective Temperature (:math:`T_{\rm eff}`), Surface Gravity (:math:`\log g`), and Metallicity ([Fe/H]), and queries a user-specified library of stellar atmospheres (defined in ``stellar_atm``). It interpolates within this grid to construct a specific Spectral Energy Distribution (SED) for the stars current features.
 
 2.  **Convolution**: This specific SED is then convolved with filter transmission curves (defined in ``instrument``) to calculate the flux passing through each filter.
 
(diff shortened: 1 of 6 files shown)


nialljmiller:
Yes, given the GitLFS issues I was trying to use the smallest usable set of the data. I dont think it would be too much more data to extend this. I think the data that is shipped with the colors module (filters and SEDs) could be considered further.

VincentVanlaer:
Closing since this has been merged to `r25.12.1_colors_updated`

PR #946, opened 2026-03-29
remove module level safe qualifiers across MESA

Debraheem:
See https://github.com/MESAHub/mesa/issues/942, this pr provides the module level variable declarations the "save" attribute across MESA.

### DIFF
--- a/math/public/math_lib_crmath.f90
+++ b/math/public/math_lib_crmath.f90
@@ -32,7 +32,7 @@ module math_lib
 
   character(LEN=16), parameter :: MATH_BACKEND = 'CRMATH'
 
-  real(dp), save :: ln10_m
+  real(dp) :: ln10_m
 
   interface safe_sqrt
      module procedure safe_sqrt_
--- a/astero/private/adipls_support.f90
+++ b/astero/private/adipls_support.f90
@@ -28,18 +28,18 @@
       implicit none
 
       ! args for adipls
-      integer, save :: i_paramset, ierr_param, i_inout, nn
-      real(dp), save, pointer :: x(:) => null()  ! (nn)
-      real(dp), save, pointer :: aa(:,:) => null()  ! (iaa_arg,nn)
-      real(dp), save :: data(8)
+      integer :: i_paramset, ierr_param, i_inout, nn
+      real(dp), pointer :: x(:) => null()  ! (nn)
+      real(dp), pointer :: aa(:,:) => null()  ! (iaa_arg,nn)
+      real(dp) :: data(8)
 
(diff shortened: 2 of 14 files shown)


VincentVanlaer:
> Does this mean that have the save attribute in the module level definitions is enough to encompass all routines in the module?
> 
> Also, I noticed when pushing this branch that:
> 
> Declaring module level save along with a local declaration level save resulted in a duplicate save error, so that tracks with your statement that the module level save declaration is likely enough?

No, this means that `save` is not needed at all at the module level. All variables at the module level (not procedures in a module) implicitly have save applied.

VincentVanlaer:
> With this new information, I wonder whether we should go the other way -- deleting all SAVE statements, as they are redundant and contribute (a small amount) to code bloat.

I agree

Debraheem:
bless you, thanks

math/public/math_lib_crmath.f90
```diff
@@ -32,7 +32,7 @@ module math_lib
 
   character(LEN=16), parameter :: MATH_BACKEND = 'CRMATH'
 
-  real(dp), save :: ln10_m
+  real(dp) :: ln10_m
```
warrickball, line 35: Obviously I'm very late to this, but why isn't this a `parameter`? It's value is set to ln(10) in `math_init` but never modified.

rhdtownsend:
Could it be because ln10_m is (potentially) initialized using the crmath library, and calls to the library can’t appear in initialization expressions?

> On Apr 23, 2026, at 5:59 PM, Warrick Ball ***@***.***> wrote:
> 
> 
> @warrickball commented on this pull request.
> In math/public/math_lib_crmath.f90:
> > @@ -32,7 +32,7 @@ module math_lib
> 
> character(LEN=16), parameter :: MATH_BACKEND = 'CRMATH'
> 
> - real(dp), save :: ln10_m
> + real(dp) :: ln10_m
> 
> Obviously I'm very late to this, but why isn't this a parameter? It's value is set to ln(10) in math_init but never modified.
> —
> Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android. Download it today! 

--

warrickball:
The init seems to be part of the MESA wrapper, like here:

https://github.com/MESAHub/mesa/blob/d2e43a4bbc6e26dc84dd5683dddf2616495e4d43/math/public/math_lib_crmath.f90#L106-L114

I'll try to have a go at defining it as a `parameter` over the weekend and see what happens.

VincentVanlaer:
We could easily replace it by the actual value, since crlibm/crmath is (modulo bugs) fully deterministic.

PR #955, opened 2026-04-03
R25.12.1 colors updated interp optimization

nialljmiller:
- Interpolation techniques have been optimised to remove data structures that were not needed.
 
 
 - Added ``sed_per_model``.  If .true., each SED output file is suffixed with the model number, preserving the full SED history rather than overwriting.
 
 
 
- Added a more intelligent parser for the colours data file paths with this logic: 

```
  clean up the input path

  if path starts with "./" or "../":
      use it as-is

  else if path starts with "/":
      if that absolute path exists:
          use it as-is
      else:
          treat it as relative to mesa_dir

  else:
      treat it as relative to mesa_dir
```

- Added memory aware flux cube loading where, if the flux cube is too big to fit into ram, the individual CSV files are loaded. This is much slower for IO but stops it from crashing the run. This is the logic:

```
if no flux_cube.bin:
    use per-file SED loading

read n_teff, n_logg, n_meta, n_lambda from file

try allocate cube[n_teff, n_logg, n_meta, n_lambda]:
    if success:
        load cube from file
    if fail:
        use per-file SED loading
```

### DIFF
--- a/colors/private/hermite_interp.f90
+++ b/colors/private/hermite_interp.f90
@@ -1,20 +1,5 @@
 ! ***********************************************************************
-!
-!   Copyright (C) 2025  Niall Miller & The MESA Team
-!
-!   This program is free software: you can redistribute it and/or modify
-!   it under the terms of the GNU Lesser General Public License
-!   as published by the Free Software Foundation,
-!   either version 3 of the License, or (at your option) any later version.
-!
-!   This program is distributed in the hope that it will be useful,
-!   but WITHOUT ANY WARRANTY; without even the implied warranty of
-!   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
-!   See the GNU Lesser General Public License for more details.
-!
-!   You should have received a copy of the GNU Lesser General Public License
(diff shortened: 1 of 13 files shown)


colors/private/hermite_interp.f90
```diff
@@ -307,20 +214,20 @@ subroutine hermite_interp_vector(x_val, y_val, z_val, &
...
+      ! Precompute Hermite basis functions (same for all wavelengths)
       h_x  = [h00(t_x), h01(t_x)]
       hx_d = [h10(t_x), h11(t_x)]
       h_y  = [h00(t_y), h01(t_y)]
       hy_d = [h10(t_y), h11(t_y)]
       h_z  = [h00(t_z), h01(t_z)]
       hz_d = [h10(t_z), h11(t_z)]
 
-      ! loop over wavelengths — the hot loop
+      ! Loop over wavelengths — the hot loop
```
rhdtownsend: It looks like there's still some potential optimizations here. Given that wz, wzd etc don't change wrt lambda, could you move the lambda loop to be on the inner level?

PR #960, opened 2026-04-09
colors/test: add SED and magnitude unit test

nialljmiller:
Unit test for the Colors module. 

The test initialises the module against the Kurucz2003 atmosphere grid and Johnson filter set (the defaults)
Then it runs the colors module for these sets:

- Teff = 5778 K, log g = 4.44, [M/H] = 0.0
- Teff = 15000 K, log g = 4.00, [M/H] = 0.0
- Teff = 4000 K,  log g = 2.00, [M/H] = 0.0

For each case it prints Mag_bol, Flux_bol, Interp_rad, and all Johnson band magnitudes (Vega system) at full double precision. 
For the first case it also prints 20 evenly sampled wavelength/flux pairs from the interpolated SED.

I expect we will need to replace the test answers after some updates
`./rn > test_output`


Debraheem:
Perhaps in addition to those 3 tests

add three groups that check the different dimensions of the tables?

Fixed Teff, fixed log_g, varying metallicity
Fixed Teff, varying log_g, Fixed metallicity
varying Teff, fixed log_g, Fixed metallicity

If this is asking a lot, we could do this after the release.

nialljmiller:
> Perhaps in addition to those 3 tests
> 
> add three groups that check the different dimensions of the tables?
> 
> Fixed Teff, fixed log_g, varying metallicity Fixed Teff, varying log_g, Fixed metallicity varying Teff, fixed log_g, Fixed metallicity
> 
> If this is asking a lot, we could do this after the release.

Good idea! This has been added.

Debraheem:
Awesome, and just to be sure, you might want to quickly plot them jus to double check everything is doing what you think? you don't have to add that to the pr though. But it's good to double check.

Debraheem:
Double check https://github.com/MESAHub/mesa/actions/runs/24213255395/job/70687307621, the build is failing because tmp.txt doesn't match your expected test_output.

Debraheem:
sorry one more thing. Can you make the output double precision? This let's us test if changes later incur minor peturbations to the results.

Debraheem:
I am on mac, I have a couple meetings, but I can pull and test locally later this evening if you can't get this to build by then!

Debraheem:
Interesting so there is slight precision differences between the compilers at the 14 decimal level. Can you add a comment on this in the testing output and fortran so we know this for the future?

Debraheem:
Also, if you want to address the merge conflict in colors test now that I merged your colors data into the updates_and_fixes.

Debraheem:
Maybe we make an issue as well, to follow up on the test cases and confirm why there are differences between both arm and linux a little above precision.

nialljmiller:
> Maybe we make an issue as well, to follow up on the test cases and confirm why there are differences between both arm and linux a little above precision.

Yes I think this is the best option. I cant seem to get the linux and mac outputs to agree past the 14th digit

PR #963, opened 2026-04-11
Bugfix/lgXne_Si timestep controls

Debraheem:
@carlnotsagan found this bug, we might want to double check for other composition driven timestep controls.

From carl:
Describe the bug
Some timestepping controls use the wrong isotope to check for return. This can result in timestep limiters never satisfying the min set before being allowed to increase. Thanks to Maria Mutz for pointing this out to me!

To Reproduce
Run a CCSN progenitor model with all fuel depletion flags on and MESA will write that the timestep is being limited by retry: delta_lg_XNe_cntr hard limit even once depleting beyond delta_lg_XNe_cntr_min.

Expected behavior
The timestep controls should shut off once the min has been satisfied.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -81,9 +81,10 @@ Colors
 Other
 ~~~~~
 
-- realigned first row of headers in terminal output after changing ``lg_Lnuc`` to ``lg_Lnuc_tot``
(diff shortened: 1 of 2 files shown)


Debraheem:
Need two reviews to merge.

VincentVanlaer (approved):
I looked through the other checks, but did not find any that also needed changes.

PR #964, opened 2026-04-12
r26.4.1 changes

VincentVanlaer:
These are all the changes from the new release branch that still need to pushed to main


docs/source/changelog.rst
```diff
@@ -4,9 +4,6 @@ Changelog
 
 .. warning:: As of r24.08.1, building MESA now requires Python (3.5 or newer) be installed.
 
-.. note:: This section describes changes present in the development version of MESA (``main`` branch) relative to the most recent release.
```
Debraheem, line 7: might consider returning this line before merging, unless we feel it's unecessary.

VincentVanlaer: Good catch, we probably should

Debraheem:
@VincentVanlaer , I think this should be about ready to merge into main, aside from any changes to github actions we'd like to make.

PR #965, opened 2026-04-12
lock versions of python packages used in CI

VincentVanlaer:
This is to make sure that new releases of linters do not cause CI for unrelated PRs to fail.

I have opted for using `pip-tools` to do this since that is a tool I am familiar with, but I have no strong feelings w.r.t. which tool exactly we should use for this.


Debraheem (approved):

Debraheem:
I think this is fine to merge.

PR #966, opened 2026-04-14
Bugfix colors segfault

Debraheem:
for r26.4.1, This pr introduces a minor tweak to colors that prevents the flux cube from being loaded into memory when use_colors = .false. . Given where we placed the return statement, we now need to explicitly load the flux cube In colors/test/test_colors.f90, since it's running without a namelist. I add a comment that we might want a separate inlist for this test case, but I left it as a comment for now.

+ (and 1 minor cleanup in rsp, does not need to be included)

Let me know if this is fine to merge, once it compiles. Hopefully it passes testing as well.

### DIFF
--- a/colors/public/colors_lib.f90
+++ b/colors/public/colors_lib.f90
@@ -133,6 +133,7 @@ contains
       ierr = 0
       call get_colors_ptr(handle, rq, ierr)
       if (ierr /= 0) return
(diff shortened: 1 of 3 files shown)


nialljmiller (commented):
This looks good to me. Thank you for fixing it!

PR #970, opened 2026-04-18
refactor inlist reading

### LINKED ISSUE #957: Harmonize missing inlist/section in inlist handling
I was looking a bit deeper into the issues @mathren highlighted in #954 where missing a missing `pgbinary` inlist causes somewhat of messy exit. While the fix is not too hard (it even simplifies the code), I did notice that the same code is repeated in various parts of the code base, but with slight differences.

- `kap`, `eos`, and `colors` allow the first inlist file in a chain to be missing. Missing `kap` and `eos` namelist sections lead to a warning, but no hard error.
-  `binary_job`, `binary_controls`, `astero_search_controls`, `astero_pgstar_controls`, `controls`, `pgstar`, and `pgbinary` produce a hard error whenever the can't read a file or don't find their namelist sections (assuming the module is enabled).

I can see benefits for both cases: early warning to the user if they made a mistake vs. less unneeded noise.

Do we want to keep the current behaviour or pick one of the options? The difference for refactoring the code is limited.

VincentVanlaer:
This commit introduces utils_namelist, which abstracts some of the common aspects of reading inlists in MESA (error message, nested inlists, ...). Previously, various different namelist reading routines were copied and modified around the code base. These have accrued various differences over time, which has been fixed now.

The following behaviour has been changed:

- pgbinary and pgstar no longer make MESA error when they are missing from inlists
- it no longer matters where in the chain of inlist a namelist section is missing. It used to be that for certain section, only the first inline in a chain was allowed to have a missing section.
- checks and copying of options only happens once all inlists have been read
- failures when reading inlists will no longer dump a stack trace on the user

### DIFF
--- a/astero/public/astero_def.f90
+++ b/astero/public/astero_def.f90
@@ -839,73 +839,39 @@
 
 
       subroutine read_astero_search_controls(filename, ierr)
+         use utils_namelist, only: read_namelist, missing_namelist_error
          character (len=*), intent(in) :: filename
          integer, intent(out) :: ierr
+
          ! initialize controls to default values
          include 'astero_search.defaults'
-         ierr = 0
-         call read1_astero_search_inlist(filename, 1, ierr)
+
+         call read_namelist(filename, read_astero_search_file, "astero_search_controls", ierr, missing_namelist_error)
       end subroutine read_astero_search_controls
 
+      subroutine read_astero_search_file(unit, iostat, iomsg, extra_inlists, extra_inlists_mask)
+         use const_def, only: strlen, max_extra_inlists
 
-      recursive subroutine read1_astero_search_inlist(filename, level, ierr)
-         character (len=*), intent(in) :: filename
-         integer, intent(in) :: level
-         integer, intent(out) :: ierr
+         integer, intent(in) :: unit
+         integer, intent(out) :: iostat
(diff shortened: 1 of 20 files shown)


VincentVanlaer:
The way I have implemented this is different from the usual way MESA is structured. Rather than having the `utils_namelist` module be included with `utils_lib`, I have exposed it separately. The main reason for doing is that it allowed me to iterate quicker on the interface design, as changes to the `utils_lib` interface require most of MESA to be recompiled.

warrickball:
I just tried to run this through `mesa_test` on our cluster and [everything failed](https://testhub.mesastar.org/refactor-inlist-reading/commits/fb1f429) with messages like:
```
 [ERROR]: too many levels of nested star_job inlist files
 ierr from read_star_job inlist
rn1 failed
******************** pisn failed: non-zero exit code ********************
```
where you can replace `pisn` with any other test case.

I doubt I'll have time to investigate before you do but if you try to run any test case in your dev version you should be able to rule out this error. If you have already managed to run any test OK, let me know.

FWIW that's on our cluster running RHEL 8.10 using MESA SDK 26.3.2.

VincentVanlaer:
Should be fixed now!

Debraheem (approved):

PR #971, opened 2026-04-18
adjust documentation on momentum outer BC

Debraheem:
Abha Vishwakarma made a mailinglist post highlighting this comment in the MESA documentation which is incorrect/outdated. I've amended the docs here.

### DIFF
--- a/star/defaults/controls.defaults
+++ b/star/defaults/controls.defaults
@@ -4694,8 +4694,8 @@
       ! Tsurf_factor
       ! ~~~~~~~~~~~~
 
-      ! used when ``use_momentum_outer_BC``
-      ! ``T_surf`` is set to ``Tsurf_factor*T_black_body(L_surf,R_surf)``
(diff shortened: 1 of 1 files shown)


VincentVanlaer:
This may be a silly questions, but what is used for then?

Debraheem:
That is not a silly question, in fact, it does appear completely deprecated. It doesn't look to be used by anything at the moment, except for it's own bookkeeping, but we should double check. 

However, I'm trying to decide if it makes sense to bring back the BB outer bc, as it might be useful for a variety of things (such as ppisn), but i'm unsure. Maybe you have an opinion here?

VincentVanlaer:
Unless you want to add the BB boundary condition in the next few weeks, I would just remove it. Then we don't forget about it if we end up not adding that boundary condition.

Debraheem:
I think I'll just remove it unless someone objects. Less clutter.

PR #975, opened 2026-04-21
Bug in reverse rates mass exponent

### LINKED ISSUE #974: Question about compute_rev_ratio
### Discussed in https://github.com/MESAHub/mesa/discussions/969

<div type='discussions-op-text'>

<sup>Originally posted by **haakoan** April 17, 2026</sup>
Question about `compute_rev_ratio` for 2-to-2 reverse rates in `reaclib_support.f90`

I am running late-stage stellar burning simulations in 3D using the FLASH code, for core-collapse supernova progenitors.  
Initial conditions were constructed with MESA and I used approx21 for the MESA simulations and the FLASH
simulations. 
When comparing the end state of my MESA and FLASH simulations I found differences
in the final compositions right above the iron core. In general, I see an enhancement in 
sulfur in FLASH.

To understand where this difference comes from, I investigated the rates
relevant for this regime (temperatures around 1e9 K) 

Concretely at T = 2.0e9 for the rate r_ar36_ap_k39
I find a ratio of MESA / FLASH: `4.53e-3 / 6.37e-4 = 7.11`
Based on my understanding of these rates and how they are calculated,
the factor 7.11 is almost exactly the A-ratio `1 / (39/144)^(3/2) = 1/0.141`. 
The ratio stays essentially constant at ~7 across T = 2-4e9 K.
I see a similar pattern for `r_si28_ap_p31` vs `r_p31_pa_si28` 
(approx 6.9 at around T=1e9 K which matches `1 / (31/112)^(3/2)` for the Si28/P31 pair).

I am not very familiar with MESA, but I am trying to understand the
code and find where the difference is coming from.

In `rates/private/reaclib_support.f90` around line 214 (current `main`):

```fortran
tmp = product(mass(1:Ni))/product(mass(Ni+1:Nt))
rates% inverse_coefficients(1,i) = pow(tmp, 1.5d0*(Ni-No)) * (product(g(1:Ni))/product(g(Ni+1:Nt)))
```
The code cites Reichert et al. 2023, which does indeed have the Ni-No factor in the A-products.
Other papers, for example Smith et al. 2023 (arXiv:2210.09965), does not have this factor (see their appendix B)

For a 2→2 reaction like `K39(p,α)Ar36` (Ni = No = 2), `pow(tmp, 1.5·(Ni−No))`
evaluates to `pow(tmp, 0) = 1`.

If I'm reading things right, this line was changed in #632 (fix for #575) from
a constant exponent `1.5d0` to `1.5d0*(Ni-No)`. 

The similar code pynucastro does it differently. 
Smith et al. 2023 appendix B equation B16 writes the detailed-balance
ratio as two separate factors with *different* exponents:

- A-ratio: `(∏ A_reactants / ∏ A_products)^(3/2)` — **exponent always 3/2**, independent of (r-p)
- Thermal factor: `(m_u kB T / 2π ℏ²)^(3(r-p)/2)` — **exponent scales with (r-p)**

In `pynucastro/rates/derived_rate.py`, this is built up as a combination of
`A_nuc^2.5` and `1/A` factors that collapse to `A^(1.5)` regardless of
`net_stoich`, plus a separate `net_stoich`-dependent block that supplies the
thermal factor. The two factors are kept separate.

Question:

-  Is the current `pow(tmp, 1.5d0*(Ni-No))` line intentional, in the sense
   that there's a convention or a separate correction applied elsewhere that
   I haven't found? I did try to trace things through
   `compute_some_inverse_lambdas` in `reaclib_eval.f90` and through
   `net_approx21.f90` for the (α,p) cases I care about, and I didn't see an
   additional factors getting applied.

</div>

Debraheem:
This critical bugfix addresses a mistake in the mass exponent term in the reverse reaction rates, raised by @haakoan here: https://github.com/MESAHub/mesa/issues/974. 

I have added an additional unit test which tests changes to these reverse ratios in the future. I believe this is important enough to warrant being merged into the r26.4.1 release.


net/test/test_output
```diff
@@ -203,11 +203,10 @@
  test_one_zone_burn_const_P
                                       number of species          21
                                  large final abundances    1.0000000000000000D-02
-                                                    o16           1    7.1798810044269901D-01
-                                                   si28           2    1.6142410416068431D-01
-                                                    s32           3    6.0439378324441878D-02
-                                                   ar36           4    2.9692711889885499D-02
```
VincentVanlaer, line 209: Is there a reason these last two lines have been removed?

docs/source/changelog.rst
```diff
@@ -10,6 +10,8 @@ Changes in r26.4.1
 
 Important bug fix for ``r25.12.1`` raised by Jake B. Hassan: the colors module could select atmosphere metallicity from ``Zbase`` instead of photospheric ``[M/H]``. This could return solar-metallicity colors and SEDs for non-solar models. See :ref:`the known bugs entry <colors_zbase_bug>` and `gh-939 <https://github.com/MESAHub/mesa/pull/939>`_. We recommend that users using the new color module upgrade to this release.
 
+Important bug fix for ``r24.03.1``, ``r24.08.1``, and ``r25.12.1`` identified by Haakon Dahle: a bug in the reverse reaction rates module, introduced in ``r24.03.1``, used an incorrect mass exponent factor in some reverse detailed-balance rates. See :ref:`the known bugs entry <reverse_rate_mass_exponent_bug>`, `gh-974 <https://github.com/MESAHub/mesa/issues/974>`_, and `gh-975 <https://github.com/MESAHub/mesa/pull/975>`_. Users who rely on reverse reaclib rates should update to this release, or refer to the known bugs page for directions on how to fix it.
```
VincentVanlaer: What kind of options would cause one to hit the bug? For example, will these be pulled in by `add_isos_and_reactions`? Or does one need to do more?

rates/private/reaclib_support.f90
```diff
@@ -220,10 +221,9 @@ subroutine compute_rev_ratio(rates,winvn)
 
             ! see equation 21 in Reichert et al. 2023 (https://doi.org/10.3847/1538-4365/acf033)
```
VincentVanlaer: Should this reference be updated, since it contains the mistake?

net/test/test_output
Debraheem: i'm not sure why they were removed.

rates/private/reaclib_support.f90
Debraheem: yes!

docs/source/changelog.rst
Debraheem: the bug is primarily active at higher temperatures (GK or higher) where the network approaches NSE conditions and reverse rates become large and comparable to forward rates. However, the reverse rates can become important at intermediate temperatures as well, including helium burning. I would say this is a global bug, and is likely relevant to everyone. I will adjust the language. 

As far as which isos, i would just say everything is affected.

net/test/test_output
Debraheem: i believe it is because this test only prints abundances above 1d-2

Debraheem:
Any further reviews are appreciated as I would like this pr merged and tested asap, so we can correct it in this release.

docs/source/changelog.rst
```diff
@@ -10,6 +10,8 @@ Changes in r26.4.1
...
 Important bug fix for ``r25.12.1`` raised by Jake B. Hassan: the colors module could select atmosphere metallicity from ``Zbase`` instead of photospheric ``[M/H]``. This could return solar-metallicity colors and SEDs for non-solar models. See :ref:`the known bugs entry <colors_zbase_bug>` and `gh-939 <https://github.com/MESAHub/mesa/pull/939>`_. We recommend that users using the new color module upgrade to this release.
 
+Important bug fix for ``r24.08.1`` and ``r25.12.1`` identified by Haakon Dahle: a bug in the reverse reaction rates module, introduced in ``r24.08.1``, used an incorrect mass exponent factor in some reverse detailed-balance rates. This is most relevant in hot helium burning and in more advanced burning stages, especially carbon burning and later, where reverse detailed-balance rates become important. For the advanced-burning cases, this typically means temperatures of order ``1 GK`` and above. See :ref:`the known bugs entry <reverse_rate_mass_exponent_bug>`, `gh-974 <https://github.com/MESAHub/mesa/issues/974>`_, and `gh-975 <https://github.com/MESAHub/mesa/pull/975>`_. We recommend that all users affected by this issue apply the fix or update to this release.
```
haakoan: If it is important to you to get the name right, it is Haakon Andresen.

matthiasfabry:
I am by no means an expert in nuclear physics, but from your comments it seems both formulations exist in the literature? Is Reichert's Eq. 21 simply a typo? If so, I'd clarify that in a code comment so someone doesn't mistakenly reintroduce it by seeing the discrepancy again.

Debraheem:
> I am by no means an expert in nuclear physics, but from your comments it seems both formulations exist in the literature? Is Reichert's Eq. 21 simply a typo? If so, I'd clarify that in a code comment so someone doesn't mistakenly reintroduce it by seeing the discrepancy again.

yes, agreed, i believe it is a typo in their manuscript but not in their code. I made comments on this in reaclib_support.f90:

https://github.com/MESAHub/mesa/blob/6fdd62228e79a0362a10e0962f45f6e3dbd85d28/rates/private/reaclib_support.f90#L212-L238

Debraheem:
I compared the original and corrected reverse-rate implementations with BBQ hydrostatic one-zone burns using six MESA networks: approx21_plus_co56, mesa_80, mesa_151, mesa_206, mesa_330, and mesa_495. The sweep covered three physical test setups, expanded into five concrete burn cases: pure he4 at 0.1 GK and 0.2 GK with logRho = 4.0, each evolved for 100 Myr; a 50/50 by-mass c12/o16 mixture at 1 GK with logRho = 6.0, evolved for 1000 yr; and Z-split compositions at 3 GK and 6 GK with logRho = 8.0, each evolved for 1e5 s, where each network was initialized with equal total mass fraction in each nuclear charge Z present in that network, and that Z fraction was then split equally among the isotopes with that Z. These were fixed-duration burns, not runs stopped by an equilibrium criterion. From these runs I generated three plot sets in plots/bbq_sweep: by_temperature, which compares the final isotope mass fractions across networks for a single source state (new or old) and includes both grouped bar plots and multi-panel pie summaries of the top 15 final isotopes; compare_labels, which compares the final isotope distributions for new versus old for the same network and burn case; and eps_nuc, which compares the corresponding eps_nuc(t) histories versus age for new and old for the same network and burn case.

Because i made hundreds of plots I share a zip, but please if others will download and look at the results, and provide comment it would be appreciated!
[bbq_sweep.zip](https://github.com/user-attachments/files/26973576/bbq_sweep.zip)

when looking at the plots: `old = r26.04.1` and `new = r26.04.1 + mass_exponent_fix`

A coupe notable things, It seems low temperatures are relatively unaffected in abundance or eps_nuc, but higher temperatures near NSE conditions at 3GK and 6GK are most affected. This is good news in my opinion, as it means most of H/He/C/O burning is unaffected T<1GK. Only when reverse rates become important does this correction come into play.


As for the test_suite, I've returned the old flame speed and width targets to the `conductive_flame` test case, as the intial bugfix for the reverse rates which introduced this max_exponent bug is what caused us to change the flame targets in the first place.

Debraheem:
[495_comparison.pdf](https://github.com/user-attachments/files/26975015/495_comparison.pdf)
[approx_comparison.pdf](https://github.com/user-attachments/files/26975016/approx_comparison.pdf)
I share pdfs of two of the comparisons for approx21 and a 495 isotope network, under the conditions described above.

mathren (approved):
Great catch @haakoan and @Debraheem!

The one-zone-burner tests you shared are pretty convincing that this is a small, but noticeable effect for late evolution. In the interest of not blocking the release, I am approving this PR and I think it could be merged as is.

I would still run in parallel tests with full stars (maybe one mass that goes to core-collapse) which may show the effect is even smaller (not pushing too hard the models, if they fail because of any other reason, then this should become a science project on its own, and interested users should always run their own tests anyways). 

Based on the `bbq` tests, I don't expect tests with full stars to surprise us or introduce astrophysically significant issues.

Debraheem:
testing seems fine excluding the conductive_flame test. https://testhub.mesastar.org/bugfix%2Freverse_rate_mass_r26.4.1/commits/5eb8b12


I ran an optional test locally and found that ppisn now fails on my arm machine on this branch. I'm not surprised given how sensitive it is in its current form. I will try to address this, and run another test.

haakoan:
> Great catch @haakoan and @Debraheem!
> 
> The one-zone-burner tests you shared are pretty convincing that this is a small, but noticeable effect for late evolution. In the interest of not blocking the release, I am approving this PR and I think it could be merged as is.
> 
> I would still run in parallel tests with full stars (maybe one mass that goes to core-collapse) which may show the effect is even smaller (not pushing too hard the models, if they fail because of any other reason, then this should become a science project on its own, and interested users should always run their own tests anyways).
> 
> Based on the `bbq` tests, I don't expect tests with full stars to surprise us or introduce astrophysically significant issues.

I can add that in my testing against FLASH, I found differences around of 10-50%, in terms of mass fractions. However, it is not a clean test. We ran MESA models until core-collapse, went back 10-15 minutes and then started 3D FLASH simulations from the MESA profiles. Thus, differences in the convective activity can account for some of that difference. The rates used for approx21 in FLASH are also older than the MESA rates and I do see some differences in the raw rates too. To complicate the issue further, the discrepancies are model dependent.

 It is also not clear to me how the approx21 results would transfer to larger networks.

Debraheem:
testing has passed 
https://testhub.mesastar.org/bugfix%2Freverse_rate_mass_r26.4.1/commits/e7ad19d

PR #978, opened 2026-04-28
fix sdk caching for github actions

Debraheem:
1.  Both linux sdk github actions were writing to the same sdk. For now, I've changed the actions such that they generate and use their own sdk cache.
2. The mac os sdk was not being cached. A path for caching has been added.

### DIFF
--- a/.github/actions/install-mesa/action.yml
+++ b/.github/actions/install-mesa/action.yml
@@ -59,7 +59,9 @@ runs:
         path: |
           mesasdk-x86_64-linux-${{inputs.sdk}}.tar.gz
           mesasdk-aarch64-macos-${{inputs.sdk}}.pkg
-        key: ${{ runner.os }}-${{inputs.sdk}}-${{ env.DYNAMIC }}
+        # If this key changes, run the workflow once to seed the new caches and
+        # again to verify cache-hit behavior.
+        key: ${{ runner.os }}-${{inputs.sdk}}${{ env.DYNAMIC && format('-{0}', env.DYNAMIC) || '' }}
 
     - name: Get SDK ${{ runner.os }} '26.3.2'
       if: ${{ (steps.cache.outputs.cache-hit != 'true') && (  inputs.sdk == '26.3.2') && (runner.os == 'Linux') }}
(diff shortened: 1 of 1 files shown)


VincentVanlaer:
What is the issue with DYNAMIC and not DYNAMIC sharing the same cache? The SDK is the same for both of them.

Debraheem:
Yes they do should use the same sdk, but at the moment i think the way the actions work have it so that when the key is set for both the dynamic on and off builds, they were trying to both write to the cache simultaneously and I think this caused one of the build actions to fail. Ideally, only one builds the cache, and both read from it, i think. I can change change it to do that, but for now i had made both build one separately.

VincentVanlaer:
Hmm, I didn't think failures in writing caches would fail the whole action. Could you point me to one of the failed ones?

Debraheem:
you could be right, i had to use the oracle for this.

 It happened here on the ubuntu branch, but once i rebuilt it, the failure disappeared... https://github.com/MESAHub/mesa/actions/runs/25074572680

also, whatever i did might have also caused some ci failure as well : https://jenkins.flatironinstitute.org/blue/organizations/jenkins/mesa/detail/main/703/pipeline

we can revert this pr and force a rebuild on main?

Debraheem:
the jenkins error might be unrelated, since it's the first time jenkins has built main in a moment>?

Debraheem:
Screenshot 2026-04-28 at 9 03 49 PM

i have a screenshot of the post_job clean up, but i don't know if this was the main error. the initial error was some sort of file storage issue : 

`System.IO.IOException: No space left on device : '/home/runner/actions-runner/cached/2.334.0/_diag/Worker_20260428-195730-utc.log`

perhaps I misinterpreted the issue.

VincentVanlaer:
> It happened here on the ubuntu branch, but once i rebuilt it, the failure disappeared... https://github.com/MESAHub/mesa/actions/runs/25074572680

Looks more like a random failure in github actions

> also, whatever i did might have also caused some ci failure as well : https://jenkins.flatironinstitute.org/blue/organizations/jenkins/mesa/detail/main/703/pipeline

I think I have seen this before in Jenkins. So could be unrelated

Debraheem:
interesting, let me know if i should revert, and perhaps leave the mac os cache change, since mac os sdk wasn't being cached.

VincentVanlaer:
I would revert this since the r26.4.1 branch refactors the actions stuff. I looks like the cache isn't entirely in order there (wrong suffix for the filename), but it's better to work based on that.

PR #981, opened 2026-05-01
add report_max_infall_inside_fe_core to namelist controls defs

carlnotsagan:
report_max_infall_inside_fe_core is not defined, this keeps the user from enabling the old fe_core_infall mass limit. unsure of other of unexpected behavior as a result of this.


VincentVanlaer:
Looks like it got accidentally removed in 889f15286438a0c821ccad2c996d8954488caaf1.

VincentVanlaer:
Note to self: add changelog and potential know bugs

Debraheem:
Thanks for fixing my error.

PR #982, opened 2026-05-02
add call to compare to target for fe core infall limit

carlnotsagan:
add (or add back?) call to compare to target allowing users to use `atol` and `rtol` for getting closer to desired value of infall for fe core.


VincentVanlaer:
The change away from `compare_to_target` was done in SVN revision 10567, but there doesn't seem to be specific reason as to why.

Running the test suite to check if any tests need adjustments should this be merged in.

fxt44:
my recollection is it was removed as part of a cleaning session due to the perception that it was not used that much (at the time) and that the capability could be implemented in a run_star_extras. happy to see it return as a supported functionality.

VincentVanlaer:
Test suite has passed, so I'll merge this in.

PR #983, opened 2026-05-02
do not stop on missing special reactions

VincentVanlaer:
The case this intends to support is testing various nuclear reaction networks with custom rates. If you have an expansive set of custom reaction rates, you no longer have to remove the rates that are not present in the network you are currently testing.


Debraheem:
This looks fine to me. It relieves a pain point for users as well.

mathren (commented):
Agreed, this is a good change and there is a warning printed anyways.

PR #986, opened 2026-05-04
mlt-TDC face reconstruction and acceleration limiting

Debraheem:
Hello, this pr is a sort of draft of some changes we could consider incorporating. I'm curious for feedback.

This branch was birthed out of doing some testing with tdc pulsations, where I was benchmarking where TDC decoupled versus RSP. It turned out that using `use_face_values_eos_and_kap_mlt_tdc ` allowed mesa to converge significantly better on courser mesh, when the opacity is not well resolved at the cell face. TDC is uses the luminosity to construct gradr, and then gradr is used inside TDC to reconstruct L and solve for Lc. Because MLT/TDC are face centered. As an example, I believe this indicates averaging the cell centered opacity onto the cell face is less thermodynamically consistent than using the face T and rho to recalculate the opacity, which is what `use_face_values_eos_and_kap_mlt_tdc` does.
[light_curves_logL_vs_day.pdf](https://github.com/user-attachments/files/27372167/light_curves_logL_vs_day.pdf)
The only other solution I could find was to double or triple the mesh resolution in MESA to get similar smoothness to RSP at this particular resolution. 

This PR does a few bigger things:

1. (The biggest refactor) Adds an option for `use_face_values_eos_and_kap_mlt_tdc`, which uses the face_averaged primitives to computes the face eos and kap quantities using an additional eos call, instead of averaging the Pressure and opacity and other thermodynamic quantities on cell faces. This generates a more accurate reconstruction at the cost of an extra eos call. 

2. Adds the arnett acceleration limited convection back to MESA as an optional internally and locally solved inside turb. I don't know if people want this back in mesa or not, but we could consider it. If you turn this on, you can get through some difficult phases like the helium flash without solver tricks. However, we should consider whether we really want this back or not.

3. `use_TDC_Y_face_seeded_newton` was introduced which let's TDC skip the bisection logic and newton iterate to get a solution faster if the solution hasn't moved very far. This is helpful for zones where convection is not changing very much and another full root bisection can be wasteful. This control in my testing reduced the number of iterations in most TDC zones in half. I'm not sure this manifests into actual improvements, but it needs testing before I can decide if it's worth keeping.

4. Some testing in turb/test, to compare different growth and decay scenarios. 

I'm sure this could be refactored and improved, so i welcome any suggestions.

I had an llm go through the branch and describe what changed. 
[TDC_branch_changes.pdf](https://github.com/user-attachments/files/27372738/TDC_branch_changes.pdf)


Debraheem:
time_dependence_conv_vel_panels
Here is a plot showing some of the comparisons for convective velocity drawn from the unit tests.

VincentVanlaer:
General comment, I haven't looked at the changes in detail: would it make sense to split these changes up into separate PRs? That way we can keep the discussions on the three main changes you mentioned separate.

Debraheem:
> General comment, I haven't looked at the changes in detail: would it make sense to split these changes up into separate PRs? That way we can keep the discussions on the three main changes you mentioned separate.

yes

Debraheem:
I will reopen separate smaller prs.

PR #988, opened 2026-05-04
add optional flag to disable mlt_fallback in TDC adjust_mass

Debraheem:
This pr introduces a dev control which affects TDC. It allows zones touched by adjust mass to use TDC inste. I think this can be more consistent the falling back to mlt, and this control is needed to evolve TDC pulsation models with mass loss as well.

In my testing, pulsation models with mass loss generate noisy features and convection in the surface layers touched by adjust_mass, because of the mlt fallback. Not falling back fixes this issue.

### DIFF
--- a/star/private/turb_support.f90
+++ b/star/private/turb_support.f90
@@ -38,9 +38,8 @@ contains
 
    !> Determines if it is safe (physically) to use TDC instead of MLT.
    !!
-   !! Currently we only know we have to fall back to MLT in cells that get touched
-   !! by adjust_mass, because there the convection speeds at the start of the
-   !! step can be badly out of whack.
+   !! By default, fall back to MLT in cells that get touched by adjust_mass.
+   !! This can be disabled to let those cells use TDC.
    !!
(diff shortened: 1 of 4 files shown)


VincentVanlaer:
Is there a situation to prefer the old behaviour?

Debraheem:
I don't know, but for now I think an optional control we can test is a good solution.

star/private/turb_support.f90
```diff
@@ -38,9 +38,8 @@ module turb_support
 
    !> Determines if it is safe (physically) to use TDC instead of MLT.
    !!
-   !! Currently we only know we have to fall back to MLT in cells that get touched
-   !! by adjust_mass, because there the convection speeds at the start of the
-   !! step can be badly out of whack.
```
VincentVanlaer, line 43: I would keep part of the original explanation that explains why the fall back is there.

VincentVanlaer (approved):
Feel free to merge once the comment is updated!

PR #991, opened 2026-05-07
add banded solver, TDC Y-face seed, skip cox if include_mlt_corr is F

Debraheem:
This pr adds:

1.  a matrix solver option with `hydro_matrix_solver = 'banded'` support and solver plumbing. This is the same banded solver RSP uses, just repurposed for star. It is significantly faster than bcyclic for pulsation models, or really anything that is on a smaller mesh. Bcyclic can parallelize more and remains a better all purpose generalized solver for MESA.

2. `use_TDC_Y_face_seeded_newton` plus the Y-face start/previous-iteration seed path.
Small supporting TDC plumbing. `use_TDC_Y_face_seeded_newton` let's TDC skip the bisection logic and newton iterate to get a solution faster if the solution hasn't moved very far. This is helpful for zones where convection is not changing very much and another full root bisection can be wasteful. This control in my testing reduced the number of iterations in most TDC zones in half. This should be tested further, but this option should reduce the time spent in TDC ~ 1/2 . I have seen 5-10% speed up in some models during testing, but the speed up only becomes important when the timestep is limited by tdc/mlt and not eos/kap or other evolve step related overhead.

3. along with 2, when `include_mlt_corr_to_TDC = .false.`, we now skip the call to Cox inside TDC, since we don't need mlt Gamma. This is one less extra mlt call inside tdc per zone per iteration.

### DIFF
--- a/star/private/star_solver.f90
+++ b/star/private/star_solver.f90
@@ -114,7 +114,7 @@
          real(dp), dimension(:,:), pointer :: dxsave=>null(), ddxsave=>null(), B=>null(), grad_f=>null(), soln=>null()
          real(dp), dimension(:), pointer :: dxsave1=>null(), ddxsave1=>null(), B1=>null(), grad_f1=>null(), &
             row_scale_factors1=>null(), col_scale_factors1=>null(), soln1=>null(), &
-            save_ublk1=>null(), save_dblk1=>null(), save_lblk1=>null()
+            save_ublk1=>null(), save_dblk1=>null(), save_lblk1=>null(), band1=>null()
          real(dp), dimension(:,:), pointer :: rhs=>null()
          integer, dimension(:), pointer :: ipiv1=>null()
          real(dp), dimension(:,:), pointer :: ddx=>null(), xder=>null()
@@ -137,7 +137,8 @@
             min_corr_coeff, max_corr_min, max_resid_min, max_abs_correction
          integer :: nz, iter, max_tries, tiny_corr_cnt, i, &
             force_iter_value, iter_for_resid_tol2, iter_for_resid_tol3, &
-            max_corr_k, max_corr_j, max_resid_k, max_resid_j
+            max_corr_k, max_corr_j, max_resid_k, max_resid_j, &
+            band_kl, band_ku, band_ldab, band_diag
          integer(i8) :: time0
          character (len=strlen) :: err_msg
          logical :: first_try, dbg_msg, passed_tol_tests, &
@@ -151,6 +152,7 @@
          real(dp), pointer, dimension(:) :: equ1=>null()
          real(dp), pointer, dimension(:,:) :: equ=>null()  ! (nvar,nz)
          real(dp), pointer, dimension(:,:) :: AF=>null()  ! (ldAF,neq)
+         real(dp), pointer, dimension(:,:) :: band=>null()  ! (band_ldab,neq)
          real(dp), pointer, dimension(:,:,:) :: ublk=>null(), dblk=>null(), lblk=>null()  ! (nvar,nvar,nz)
          real(dp), dimension(:,:,:), pointer :: lblkF=>null(), dblkF=>null(), ublkF=>null()  ! (nvar,nvar,nz)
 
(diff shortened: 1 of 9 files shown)


star/private/star_solver.f90
```diff
@@ -1022,8 +1024,17 @@ logical function solve_equ()
                !$OMP END PARALLEL DO SIMD
             end if
 
-            call factor_mtx(ierr)
-            if (ierr == 0) call solve_mtx(ierr)
+            select case (trim(s% hydro_matrix_solver))
+            case ('', 'bcyclic')
```
VincentVanlaer: I don't think it makes sense to have a `''` option, since that only occurs if the user sets `hydro_matrix_solver` to an empty string,

star/private/star_solver.f90
```diff
@@ -1022,8 +1024,17 @@ logical function solve_equ()
...
+               call factor_mtx(ierr)
```
VincentVanlaer: We should rename this to `factor_bcyclic_mtx` and `solve_bcyclic_mtx`

star/private/star_solver.f90
```diff
@@ -1066,6 +1077,88 @@ subroutine solve_mtx(ierr)
          end subroutine solve_mtx
 
 
+         subroutine solve_banded_mtx(ierr)
+            integer, intent(out) :: ierr
+            integer :: info, k, i_equ, i_var, irow, jcol, row0, col0, iband
+            include 'formats'
```
VincentVanlaer: This can be removed as none of the write statements use any of the specific formats in formats

star/private/star_solver.f90
```diff
@@ -1902,8 +2006,9 @@ subroutine cleanup()
 
             nullify(equ, equ1, dxsave1,dxsave, ddxsave, B1, &
                      soln1, grad_f1, rhs, xder, ddx, row_scale_factors1,&
-                     col_scale_factors1, save_ublk1, save_dblk1, save_lblk1,&
+                     col_scale_factors1, save_ublk1, save_dblk1, save_lblk1, band1,&
                      B, soln, grad_f,ipiv1, ublk1, dblk1, lblk1, ublkF1,dblkF1, lblkF1)
+            nullify(band)
```
VincentVanlaer: Any reason for why this isn't in the previous `nullify`

turb/private/tdc.f90
```diff
@@ -47,12 +47,17 @@ module tdc
...
+   !! @param Y_face_guess Candidate superadiabaticity for the local solve.
+   subroutine get_TDC_solution(info, scale, Zlb, Zub, conv_vel, Y_face, tdc_num_iters, ierr, &
+         have_Y_face_guess, Y_face_guess)
       type(tdc_info), intent(in) :: info
       real(dp), intent(in) :: scale
       type(auto_diff_real_tdc), intent(in) :: Zlb, Zub
       type(auto_diff_real_star_order1),intent(out) :: conv_vel, Y_face
       integer, intent(out) :: tdc_num_iters, ierr
+      logical, intent(in), optional :: have_Y_face_guess
+      real(dp), intent(in), optional :: Y_face_guess
```
VincentVanlaer: Since it is `optional` would it not make more sense to just check for `Y_face_guess` with `present`?

Applies to all of these functions

star/private/turb_support.f90
```diff
@@ -325,13 +325,25 @@ subroutine Get_results(s, k, MLT_option, &  ! NOTE: k=0 is a valid arg
             scale = max(abs(s% L_start(k)), 1d-3*maxval(s% L_start(1:s% nz)))
          end if
 
+         have_Y_face_guess = s% use_TDC_Y_face_seeded_newton .and. s% doing_solver_iterations
+         if (have_Y_face_guess) then
+            if (s% solver_iter == 0) then
+               Y_face_guess = s% Y_face_start(k)
```
VincentVanlaer: Q: Is this only called once for each time `save_start_values` is called? If so, this if statement doesn't do anything practical, since `Y_face_start` is then the same as `Y_face`.

VincentVanlaer (approved):
Is there a test case you could enable this in? That way we don't break this option.

star/private/star_solver.f90
Debraheem: no

turb/private/tdc.f90
Debraheem: Yeah, that's a good idea.

star/private/turb_support.f90
Debraheem: Yeah, that's true, we could just use y_face here. Both are the same.

Debraheem:
> Is there a test case you could enable this in? That way we don't break this option.

Yes, I will be using this in the pulsation models I introduce to the test suite for TDC.

Debraheem:
RSP2 will likely use it as well, once it's up and running (the banded solver). The TDC_Y_face_seeded_newton can be turned on in any test case we want.

turb/private/tdc.f90
Debraheem: i actually removed it from being optional, because it's not useful as an optional control unless we call TDC with and without it, and at that point it would just clutter turb_support. I just use the value as a gate now and pass it in as a normal control because if Y_face(k) = 0d0, the seeded newton is skipped, so checking for Y_face <= 0d0 and is bad is a good enough check i think.

PR #997, opened 2026-05-14
fix low density/Pressure SCVH EOS tables 

### LINKED ISSUE #822: High values for the thermodynamic consistency metric dsp in blending involving FreeEOS, and as well in OPAL/SCVH tables
Hi,

Context: In my scientific problem, I am evolving the envelope of very massive stars as they expand due to the injection of additional heat. During the course of this evolution, as the outer part of the envelope expands and cools, the model began closing the blending zone at the lower temperature limit of FreeEOS. This led to abnormal behavior in the reported values of temperature and density.

Upon closer inspection, I found that this behavior correlates with unusually high values of the thermodynamic consistency metric `dsp`, which is defined as:

$$
\text{dsp} = -\rho \left( \frac{ds}{d\ln \rho} \right) \bigg/ \left( \frac{dp}{dT} \right) - 1
$$

This metric quantifies the consistency of the thermodynamic derivatives. Large deviations from zero may indicate that the EOS is no longer self-consistent in that regime.

<p float="left">
  
  
</p>

My `eos` controls for this run are:
```fortran
      use_simple_Skye_blends = .true.
      
      logT_min_for_any_Skye = 7.9
      logT_min_for_all_Skye = 8.0
      
      !logQ_max_FreeEOS_hi = 5.6d0
      !logQ_max_FreeEOS_lo = 4.0d0

      
      logT_max_FreeEOS_hi = 8.2d0 ! FreeEOS tabulations go up to here
      logT_max_FreeEOS_lo = 8.1d0

      logT_min_FreeEOS_hi = 3.0d0
       logT_min_FreeEOS_lo = 2.7d0
       logQ_min_FreeEOS_hi = -8.0d0
       logQ_min_FreeEOS_lo = -10.0d0
      
      logT_low_all_HELM = 2.2d0 ! HELM for lgT <= this
      logT_low_all_SCVH = 2.3d0 ! SCVH for lgT >= this
```

For the default EOS blending limits one still find inconsistent values for dsp in similar regions:

<p float="left">
  
  
</p>

Similarly if one sets the lower T blending for freeEOS higher as:
```fortran
      logT_min_FreeEOS_hi = 3.6d0
      logT_min_FreeEOS_lo = 3.5d0

```

<p float="left">
  
  
</p>

Or if one removes the OPAL/SCVH tables:
```fortran
     use_OPAL_SCVH = .false.
```

<p float="left">
  
  
</p>

Given this, I am unable to set arrangement of the EOS that won't produce any weird behaviour while crossing that region of low T and low Rho. I'm opening this issue to better understand why `dsp` becomes large in this part of the envelope and whether this behavior is expected at the edge of FreeEOS's valid temperature range.

### LINKED ISSUE #921: Failing to blend EOS around Z=0.04
When diffusion is turned on and Z drops below 0.04, the EOS changes. This leads to discontinuities in the star's evolution.

This was found while running a ~0.65 Msun model at Z a little larger than 0.04 with do_element_diffusion = .true.

Setting OPAL_SCVH = .false. removes this discontinuity.

Below is surf_cell_z vs. logTeff for model grid of a ~0.65 Msun star on the main sequence. Note that when surf_cell_z crosses 0.04, there is a sharp change in logTeff.

Image

**Version**
 - MESA version [r24.08.1]
 - MESA SDK VERSION [aarch64-macos-25.8.1.pkg)]
 - OS: [MacOS 15.6.1]

### LINKED ISSUE #995: OPAL/SCVH and Free_EOS fallback to HELM
Please see this thread on the mailing-list initially raised by Natasha: 
https://lists.mesastar.org/pipermail/mesa-users/2026-May/015873.html

She highlights two regions of concern 

Image

Image

Image

In the low density temperature regime, the mean molecular weight and likely electron fraction returned by the eos is that of an ionized gas, which is unphysical. This is likely because the tabulated eos table files (the data files themselves) for Free_EOS and OPAL_SCVH in MESA fall back to HELM in their tabulation, and return :

mu = 0.66667
log_free_e = -99.00000

The Free EOS table floor is logT = 3, at and below which HELM values are tabulated
The same is true for the SCVH routines below the logP = -0.6 boundary (logRho ~ 10 -11 ish) below which HELM values are tabulated in the table. 

These data tables get generated through MESA, so it looks like MESA is falling back to HELM before the table edge routines can activate for SCVH. Since the blend is something like:

```
if (logRho >= logRho5) then
   pure_opal_scvh
else if (logRho > logRho6) then
   blend
else
   pure_helm
end if
```
with 

```
logRho5 = -9.0d0
logRho6 = -9.99d0
```

Debraheem:
This pr is an attempt to address https://github.com/MESAHub/mesa/issues/995, which spawned from a mailing-list thread initially raised by Natasha Ivanova:

I believe the main concern was the mean molecular weight being that of an ionized mixture in the regions highighted in the plots below, when adopting both Free_EOS and OPAL/SCVH. 

https://lists.mesastar.org/pipermail/mesa-users/2026-May/015873.html:
eos_mu

Particularly in the low T-rho regime, as I'm not sure there is a better fallback than HELM in the high pressure regime (that would require an actual eos for that region).

For the SCVH tables, below `logP_min = -0.6` the SCVH coverage is now supplied analytically in
`eos/eosDT_builder/src/scvh_core.f90`. For each fixed `logT` row and each species, the tail uses the edge state and extrapolates to the ideal gas limit. 

```
logP0 = -0.6
logrho0 = logrho(logP0, logT)
logu0 = logu(logP0, logT)
logs0 = logs(logP0, logT)
R = 10**(logP0 - logrho0 - logT)
```

and the ideal low-pressure limit

```
logrho_ideal(logP) = logrho0 + logP - logP0
logu_ideal(logP) = logu0
S_ideal(logP) = S0 + R*ln(10)*(logP0 - logP)
logs_ideal(logP) = log10(S_ideal)
comp_ideal(logP) = comp0
```

The implementation uses this ideal value below the first lower-pressure grid
step and applies a local cubic correction between `logP0` and that point so the
tail is value-continuous and has the same `d/dlogP` as the restored processed
SCVH table at the edge. That keeps the generated eosDT interpolation support
smooth without editing the raw SCVH data files.

In this pr, I also noticed when regenerating the eos tables that there was a HELM fallback in
the high pressure region, and it produced noisy `mu` features
near the OPAL/SCVH support boundary in regenerated tables. I think this fallback didn't exist, 
when the tables where generated the first time, so i turned it off for the table generation.  see check_results :

was 
             `gamma1_opalscvh <= 1d0 .or. gamma1_opalscvh >= 2d0 .or. grad_ad_opalscvh >= 50d0) then`
now
             `gamma1_opalscvh <= 0d0) then`
 
 This might not be correct, but it was the only way for me to make the new eos smooth like the old eos tables, but maybe this expression should be adjusted.
 
 See the plots below
mu OPAL_SCVH only
messed_up_eos

  
After adding the smooth extrapolation to SCVH, i also noticed there was a bad row of partials on the
FreeEOS SCVH boundary at logT = 3 - 3.1. It turns out the fallback here were rows generated from the 
MESA EOS with `lnPgas`, `lnE`, and `lnS` in natural log form, but the FreeEOS data files
store those columns as base-10 logarithms. so i these just need to be converted in `eos/eosFreeEOS_builder/src/free_eos_table.f90`. (See the original issue below).
User attachment

I had to regenerate the eos for this pr.

- `eos/eosFreeEOS_data.tar.xz`
- `eos/eosDT_data.tar.xz`

I bumped the the FreeEOS table version and the eosDT table version to `52`.

A few plots from the new eos looks like this:

(a bit low res for the eos plotter)
eos_regions

eos_plotter_lnE_dlnT_v52_freeeos_fix
eos_plotter_mu_current_v52_freeeos_fix
eos_plotter_frac_HELM_current
eos_plotter_frac_OPAL_SCVH_current

The eosDT region definitions shifted very slightly in `eos/eosDT_builder/src/eos_regions_defs.dek` and
`eos/eosDT_builder/src/eos_regions_code.dek` keep the generated OPAL/SCVH support
out of the warm, very-low-density FreeEOS side. This prevents the SCVH
low-pressure floor from interacting with the FreeEOS boundary at
`logrho ~ -15`.

This PR probably needs some further attention validation and testing, maybe clean up.


evbauer:
Wow. Did you actually rerun the full FreeEOS preprocessor to regenerate all of the FreeEOS tables after the fix for the MESA fallback points? What version of FreeEOS did you use? Last time I tried this, I was having trouble getting an external FreeEOS installation to compile, run, and be compatible with the MESA preprocessor. But that was a while ago.

evbauer:
Rather than extrapolating "OPAL/SCVH" over to low density, why not just blend over to neutral ideal gas there? It seems like making a table that extrapolates toward ideal gas is maybe just a less-optimal version, since we can do it analytically with auto-diff now?

eos/eosFreeEOS_builder/src/free_eos_table.f90
```diff
@@ -564,7 +564,12 @@ subroutine mesa_eos_eval( logRho0, logT, mass_Frac, eos_result)
       end if
 
       eos_result(1:num_eos_basic_results) = res
-      eos_result(i_lnRho) = logRho
+
+      ! FreeEOS data files store these logarithms in base 10.
+      eos_result(i_lnPgas) = res(i_lnPgas)/ln10
+      eos_result(i_lnE) = res(i_lnE)/ln10
+      eos_result(i_lnS) = res(i_lnS)/ln10
+      eos_result(i_lnRho) = log10Rho
```
evbauer, line 572: Great catch! I think you're right about this. Do you know if it has an impact on anything?

In practice, I think we cut out most of the region where FreeEOS was failing and falling back to MESA with the "logQ_cut"-related boundary controls. Anyway, I've always been a bit uncomfortable with the idea that our "FreeEOS" tables that were meant to replace the older EOS might end up falling back to the older EOS without being very clear about that. So I like the fix, but I don't really like the existence of the thing that we're fixing here...

eos/eosDT_builder/src/create_eos_files.f90
```diff
@@ -178,7 +179,7 @@ subroutine Make_EoS_Files(Z_in, X_in, include_radiation)
                call do_stop('failed in helm_opal_scvh')
             end if
 
-            write (io_unit, '(f4.2,3(f10.5),7(1pe13.5),1(0pf9.5),4(0pf10.5),2(0pf11.5))') &
+            write (io_unit, '(f4.2,3(1x,f10.5),7(1x,1pe13.5),1x,0pf9.5,1x,f12.5,3(1x,1pe13.5),2(1x,0pf12.5))') &
```
evbauer: Why did the write format change?

eos/eosFreeEOS_builder/src/free_eos_table.f90
Debraheem: I've definitely seen models crash from this, in fact this issue from jaime comes to mind : https://github.com/MESAHub/mesa/issues/822

Debraheem:
> Wow. Did you actually rerun the full FreeEOS preprocessor to regenerate all of the FreeEOS tables after the fix for the MESA fallback points? What version of FreeEOS did you use? Last time I tried this, I was having trouble getting an external FreeEOS installation to compile, run, and be compatible with the MESA preprocessor. But that was a while ago.

No, I did not rerun the external FreeEOS code or regenerate the raw FreeEOS data. The raw FreeEOS part is the same. The confusing bit is that MESA’s eosFreeEOS data is not purely raw FreeEOS output. It also contains rows where the FreeEOS builder and falls back to the MESA EOS, marked by MESA = 1. This PR patches/re-stitches those fallback rows using the corrected MESA EOS behavior, including the SCVH low pressure extension, and fixes the log base for lnPgas, lnE, and lnS in those fallback rows. So this is not a new raw FreeEOS table generation. It is a correction to the MESA-packaged FreeEOS support tables in the regions where they were already using MESA fallback values.

I have a longer term goal of replacing this entire region with pure FreeEOS using FreeEOS3 and maybe chemEOS at lowT in a Free Energy formulation (your initial idea @evbauer), but I can't do this at the moment. Indeed, I also had the same issues as you regenerating the original tables last i tried, but I think I've learned enough now to give it another attempt soon. Perhaps with AI and some of @VincentVanlaer's compiler knowledge we can get FreeEOS to build raw tables again in the future.   


> Rather than extrapolating "OPAL/SCVH" over to low density, why not just blend over to neutral ideal gas there? It seems like making a table that extrapolates toward ideal gas is maybe just a less-optimal version, since we can do it analytically with auto-diff now?

This is close to what I did, but not quite the same as blending to the existing neutral ideal EOS at runtime.

In this PR the low-P SCVH extension is built during eosDT preprocessing. Below the processed SCVH edge at logP_min = -0.6, the code makes an edge-matched ideal-gas tail and writes that into the generated OPAL/SCVH support tables. The thermodynamic variables are pushed toward the ideal low-pressure behavior, but the H/He molecular and ionization fractions are taken from the SCVH table edge and then held fixed in pressure through the tail. See scvh_core.f90 around the new interp_vals_lowP_tail/tail_species logic, especially the calls that carry compv1/compv2 from the edge, and the later mu/logNe calculation.

So this is not exactly a pure neutral ideal blend. A runtime blend to the existing ideal EOS would be cleaner architecturally, but it would also change the EOS blending machinery and would give neutral-ideal composition unless we added a separate analytic treatment for H2/H/H+ and He/He+/He++. I did it in the preprocessor because it is a more local fix: the generated eosDT table stays smooth at the old SCVH edge, the region stops falling back to HELM, and the runtime EOS selection does not need a new component boundary.


on why i did it in the tables as opposed to doing it at runtime during the eos evaluations: 

My  fix did not need to be in the tables. The same tail formula could be evaluated analytically. However, in MESA’s current architecture, OPAL/SCVH at runtime is already reading in a preprocessed eosDT table. So putting the tail in the preprocessor is the lowest-impact place to fix the existing generated table and not introducing a bunch of other heuristics that have to computed at runtime.

The distinction is:

- Your suggestion, an analytic runtime version: add a new runtime branch/blend that detects the low-density SCVH edge and evaluates an analytic tail or neutral ideal EOS there.

- My current PR: do the analogous extension in the preprocessing stage, so the generated OPAL/SCVH eosDT tables have an analytic edge-matched tail. Runtime sees a smooth table and does not need new EOS selection logic.

The runtime version is possible, but it's a bigger change. If it used the existing neutral ideal EOS, it would not reproduce what I am doing either, since the current PR is keeping the ionization fractions anchored to the SCVH edge state (H2/H/H+ and He/He+/He++ fractions). The analytic approach at evaluation time could grab the SCVH edge state as a function of logT, then use the requested logP to carry those fractions into the tail and match energy/entropy/derivatives at the edge every time the routine is called. But then you are effectively moving the preprocessing logic into runtime, and recomputing that matching every call. At that point it seemed simpler to put the smooth extension into the table once and let the existing eosDT machinery handle it. And then there is no confusion in scenarios when someone looks at the Free_EOS data file in the data directory they arent pulling out Helm values and getting confused, because the precprocessor was falling back to HELM (this is what's happening now).

So the pragmatic reason for tables is: the existing OPAL/SCVH product is a table, and the problem was just confusing table support/fallback behavior that's not communicated to the user (so the free eos data files have a lot of fallback from MESA's other EOS'). Replacing the HELM fallback smoothly in preprocessing fixes the immediate issue without inventing a new runtime EOS boundary.  

Maybe your analytic runtime approach could be cleaner, but i figured this pr is smaller in what it actually changes. I'm happy with whatever is ultimately cleaner/better. I don't mind backing up and switching the approach. Also maybe it's not necessary to keep the ionization states, in which case an ideal eos blend might be better? What do you think, i defer to your judgement on the best path forward.

eos/eosFreeEOS_builder/src/free_eos_table.f90
Debraheem: after asking an ai to check this further: 

It affects every row marked `MESA = 1` in the FreeEOS tables, since those rows were written through the `mesa_eos_eval` fallback path. It only showed up at the low-T boundary because that was the place where runtime interpolation of the FreeEOS data was actually reaching those `MESA = 1` rows.

total numeric rows: 14,399,221
MESA fallback rows: 2,618,641

examples pulled from the tables comparing the old and new free eos tables: 

Here are a few matched v51/v52 fallback rows from `mesa-FreeEOS_02z70x.data`. In all of these rows, the v51 values are larger than the v52 values by `ln(10)`, which is the sign that v51 wrote natural-log values into fields that are read as base-10 logs.

| logT | T [K] | logRho | rho [g/cm^3] | v51 logP/logE/logS | v52 logP/logE/logS |
|---:|---:|---:|---:|---|---|
| 2.10 | 1.26e2 | -17.89 | 1.29e-18 | -10.0294 / 22.6406 / 20.9587 | -4.3557 / 9.8327 / 9.1022 |
| 2.50 | 3.16e2 | -9.20 | 6.31e-10 | 1.9463 / 23.8743 / 20.7341 | 0.8453 / 10.3685 / 9.0047 |
| 3.00 | 1.00e3 | -0.37 | 4.27e-1 | 23.4272 / 24.6846 / 19.4259 | 10.1743 / 10.7204 / 8.4366 |
| 3.50 | 3.16e3 | 0.18 | 1.51e0 | 29.2003 / 28.4440 / 19.7273 | 12.6815 / 12.3531 / 8.5675 |
| 4.00 | 1.00e4 | 1.09 | 1.23e1 | 33.3487 / 30.8600 / 19.6948 | 14.4831 / 13.4023 / 8.5533 |
| 5.00 | 1.00e5 | 1.65 | 4.47e1 | 35.7939 / 32.1696 / 20.1509 | 15.5451 / 13.9711 / 8.7514 |
|

eos/eosDT_builder/src/create_eos_files.f90
Debraheem: Some values overflowed the old fixed-width `f10.5` columns when I regenerated the eosDT tables as part of this PR. I still need to check why this overflow happened. It might be symptoms of another bug i might have introduced in this pr, so i'm looking closer.

Debraheem:
I had some AI help going through the differences, and I made plots comparing the tables in this PR to the tables currently on main. One useful detail: in git history, eos/eosDT_data.tar.xz is unchanged between the initial MESA r15140 import and this PR, so I cannot tell from git when the v51 tables were originally generated. They predate the public GitHub history.

The main difference I found is in the eosDT preprocessor path. The v51/main generation source uses:

    scvh_only = .true.

so the builder forces SCVH first across the eosDT output grid. It then applies the stricter check_results filter, including:

    gamma1_opalscvh <= 1d0 .or. gamma1_opalscvh >= 2d0 .or. grad_ad_opalscvh >= 50d0

If that forced-SCVH point fails the filter, the driver retries with OPAL; if that also fails, it falls back to HELM.

When I regenerated the tables, the OPAL/SCVH region looked noisier than the v51 tables in this region:
Screenshot 2026-05-27 at 6 02 07 PM



 The fix that made to make the regenerated tables look smooth again was to stop forcing SCVH everywhere and instead use:

    opal_scvh_only = .true.
    scvh_only = .false.

so opal_scvh_driver can use its internal OPAL/SCVH boundary/blend. I also relaxed the gamma/grad_ad part of check_results from the strict condition above to:

    gamma1_opalscvh <= 0d0

while still rejecting bad values. I feel like relaxing this only made the gamma1 plots smoother on the new branch as compared to main. 

So I think the difference right is:

    v51/main behavior: force SCVH, reject suspicious gamma1/grad_ad values, patch rejected points with OPAL
    PR behavior: use normal OPAL/SCVH selection/blending, reject only clearly broken OPAL/SCVH points

The high-density differences and f10.5 overflow are a separate side effect of changing that preprocessor path. Some fallback points that were probably OPAL-patched in the original v51-preprocessor path can become HELM fallback points in the new path. Current HELM returns finite but very large derivatives at a few of those high-density support points, which exposed the fixed width write format problem. THat occurs in the 

The derivative plots are sensitive to all of this because MESA builds bicubic interpolation coefficients from the regenerated table surface. So changing whether a patch is SCVH, OPAL, OPAL/SCVH blend, HELM, or the new low-P tail can change d/dlnT and d/dlnRho even when the value plots look broadly similar.


Here are some comparison plots to main. Somethings are improved, others maybe a little worse. curious for feedback.
[compare_branch_to_main.pdf](https://github.com/user-attachments/files/28324819/compare_branch_to_main.pdf)


Either way, the main lesson im learning is that the eos data tables haven't been regenerated in a long time, and some things have changed since then. I'm slowly going through all the noisy patches to see if i can quench those as well.

Debraheem:
comparing this branch to main, i have now tested four variations on the EOS:

Each folder contains the same comparison products:

- `compare_branch_to_main.pdf`
- `all_plots_contact.png`
- `manifest.csv`

The folder name gives the controls used to regenerate the current/PR-side eosDT tables.

| folder | OPAL/SCVH selection control | gamma/grad_ad check |
|---|---|---|
| `opal_scvh_only_relaxed_gamma_test` | `opal_scvh_only = .true.`, `scvh_only = .false.` | relaxed: `gamma1_opalscvh <= 0d0` |
| `opal_scvh_only_strict_gamma_grad_ad_test` | `opal_scvh_only = .true.`, `scvh_only = .false.` | strict: `gamma1_opalscvh <= 1d0 .or. gamma1_opalscvh >= 2d0 .or. grad_ad_opalscvh >= 50d0` |
| `scvh_only_relaxed_gamma_test` | `opal_scvh_only = .false.`, `scvh_only = .true.` | relaxed: `gamma1_opalscvh <= 0d0` |
| `scvh_only_strict_gamma_grad_ad_test` (main) | `opal_scvh_only = .false.`, `scvh_only = .true.` | strict: `gamma1_opalscvh <= 1d0 .or. gamma1_opalscvh >= 2d0 .or. grad_ad_opalscvh >= 50d0` |

https://www.dropbox.com/scl/fi/yjhcod42ddcnrqyoysmpu/EOS_regeneration_tests.zip?rlkey=oc78ls170aq8ybjmkxhexfvts&st=uxqak897&dl=0


It looks to me like :

| `opal_scvh_only_relaxed_gamma_test` | `opal_scvh_only = .true.`, `scvh_only = .false.` | relaxed: `gamma1_opalscvh <= 0d0` |

is the smoothest and most self consistent, but it's interesting nonetheless that regenerating the eos produces differences since the eos has changed substantially over time, but the data files and including their fallback to HELM(Helm as it was many years ago) significantly different. A major issue from regenerating the eos appears to be sudden checksum issues, as raised by @warrickball , see https://testhub.mesastar.org/bugfix%2FOPAL_SCVH_low_logP_extension/commits/head. I am looking into this at the moment. It's both interesting and concerning.


I've also made some comparison plots showing FreeEOS on this branch vs main. MESA=1 points are where FreeEOS was falling back to HELM, which was using the log base (corrected in this branch). It also shows orange where OPAL_SCVH was previously falling back to HELM, but is now falling back to ideal in this branch.

[freeeos_mesa_fallback_regions_compare.pdf](https://github.com/user-attachments/files/28534202/freeeos_mesa_fallback_regions_compare.pdf)


I've also corrected a major bug in the Z blend for OPAL/SCVH at Z=0.04. Previously MESA's blend was not turning on because the alpha for blending was set to 0. This caused major issues in some models, see https://github.com/MESAHub/mesa/issues/921 where I explained this further. This has been corrected in this branch as well:

before: [z_blend_scan_main.pdf](https://github.com/user-attachments/files/28534277/z_blend_scan_main.pdf)
after: [z_blend_scan_fixed_this_branch.pdf](https://github.com/user-attachments/files/28534299/z_blend_scan_fixed_this_branch.pdf)

eos/private/eosdt_eval.f90
```diff
@@ -992,26 +999,103 @@ subroutine get_opal_scvh_for_eosdt( &
...
+            d_dlnT_in(i_phase:i_latent_ddlnRho) = 0d0
+            d_dlnd_in(i_phase:i_latent_ddlnRho) = 0d0
+
+            res_in(i_frac:i_frac+num_eos_frac_results-1) = 0.0d0
+            d_dlnd_in(i_frac:i_frac+num_eos_frac_results-1) = 0.0d0
+            d_dlnT_in(i_frac:i_frac+num_eos_frac_results-1) = 0.0d0
+
+            res_in(i_frac_OPAL_SCVH) = 1.0d0
+
+         end subroutine mark_opal_scvh_result
```
VincentVanlaer, line 1098: Why not add an additional blending region to `get_opal_scvh_alfa_and_partials`, rather then reimplementing the blending infrastructure that already exists?

eos/private/eosdt_eval.f90
```diff
@@ -768,6 +768,12 @@ subroutine get_HELM_alfa( &
          ierr = 0
          ht => eos_ht
          helm_blend_width = 0.1d0
+         if (rq% use_FreeEOS .and. logT <= rq% logT_min_FreeEOS_lo) then
```
VincentVanlaer, line 771: Why is HELM force disabled below `logT_min_FreeEOS_lo` if FreeEOS is enabled. If HELM is not good here, should it not always be disabled in this region? It feels a bit of a violation to have EoS lower in the stack depend on what is happening higher in the stack.

eos/plotter/src/eos_plotter.f90
```diff
@@ -402,8 +404,8 @@ program eos_plotter
          end if
 
 
-         if (only_blend_regions .and. .not. in_eos_blend(res)) then
-            call set_nan(res1)
+         if (only_blend_regions) then
+            if (.not. in_eos_blend(res)) call set_nan(res1)
```
VincentVanlaer, line 408: Nit: what is the purpose of this change?

VincentVanlaer:
> In this pr, I also noticed when regenerating the eos tables that there was a HELM fallback in
> the high pressure region, and it produced noisy mu features
> near the OPAL/SCVH support boundary in regenerated tables. I think this fallback didn't exist,
> when the tables where generated the first time, so i turned it off for the table generation. see check_results :
> 
> was
> gamma1_opalscvh <= 1d0 .or. gamma1_opalscvh >= 2d0 .or. grad_ad_opalscvh >= 50d0) then
> now
> gamma1_opalscvh <= 0d0) then
> 
> This might not be correct, but it was the only way for me to make the new eos smooth like the old eos tables, but maybe this expression should be adjusted.

The current OPAL/SCVH tables stem from MESA revision 5596. All changes to the builder since then have not been actually been included in the actual tables. In particular, the only check for whether one should fall back to HELM was `is_bad_num(dE_dRho_opalscvh)`. This revision also had all `_only` flags set to false. There may have been other changes as well that cause issues.

<details>
  <summary>Receipts</summary>

- `eosDT_data.tar.xz` comes from [the initial commit](https://github.com/MESAHub/mesa/blob/fd396fd73d3f936da8063ffdf9d92361882eb557/eos/eosDT_data.tar.xz).
- This file was originally a bz2 compressed file rather than xz compressed (changed in rev 11558)
- The last meaningful change to this file was in rev 9935.
- Which is to restore the file to version of rev 5596 (same git hash 7a7048b53450aa97037ff59b0e110ccf31f838c0)

I can also mostly reproduce it bit-for-bit with the old mesa version (there are some very small changes in some of the output files, e.g. signed zeroes, a single least significant digit change)
</details>

eos/plotter/src/eos_plotter.f90
Debraheem: no good reason. just an artifact from using an llm on the plotter. It's just a safeguard that makes sure that "in_eos_blend" is skipped when only_blend_regions is false. I would call it an unnecessary change.

eos/private/eosdt_eval.f90
Debraheem: I added this local blend because the normal OPAL/SCVH fallback could still select HELM when we wanted a fallback to ideal, but it's clunky since it duplicates the normal blending logic as you pointed out. This could be moved into get_opal_scvh_alfa_and_partials instead of embedding it inside the opal_scvh evaluation itself, but I need to make sure there aren't any other unintended consequences of doing that.

eos/private/eosdt_eval.f90
Debraheem: yeah... this can be addressed with the comment above.

eos/eosDT_builder/src/scvh_core.f90
```diff
@@ -402,6 +402,11 @@ subroutine interp_vals_bicub(only_densities, search_for_SCVH, den_h, ener_h, ent
          max_logP = logPs(kP)
       end if
 
+      if (logP < logP_min) then
+         info = -1
+         return
+      end if
+
```
VincentVanlaer, line 409: When is this actually triggered? Afaict this guard already exists in `fscvh`.

eos/eosDT_builder/src/helm_opal_scvh_driver.f90
```diff
@@ -414,19 +418,14 @@ subroutine get_helmeos(include_radiation)
...
-            write (*, 1) 'Z', Z
-            write (*, 1) 'X', X
-            write (*, 1) 'abar', abar
-            write (*, 1) 'zbar', zbar
-            write (*, *) 'clip_to_table_boundaries', clip_to_table_boundaries
-            write (*, *) 'include_radiation', include_radiation
-            write (*, *) 'stop in get_helmeos'
-            stop 1
+            call helmeos2(temp, logT, den, logRho, abar, zbar, 1d6, 1d3, &
+                          helm_res, clip_to_table_boundaries, .true., .false., off_table, ierr)
```
VincentVanlaer, line 422: What is the reasoning for disabling electron/positron contributions. Is it a known problem in HELM that these can cause issues? Our main blend infrastructure does not do does, but rather just has a global switch for whether to include these contributions or not (this is also not exercised in the test suite, so I don't know whether that even has any practical use).

eos/eosDT_builder/src/helm_opal_scvh_driver.f90
```diff
@@ -460,6 +459,63 @@ subroutine get_helmeos(include_radiation)
 
       end subroutine get_helmeos
 
+      ! Used only when HELM cannot fill an eosDT support point.
+      subroutine get_ideal_helmeos(include_radiation)
```
VincentVanlaer, line 463: I am wondering what has changed compared to the 5596 version of the code that makes this necessary. Previously the fallback to HELM when OPAL/SCVH had problems appeared to be sufficient.

eos/defaults/eos.defaults
```diff
@@ -101,8 +101,8 @@
       logRho_max_FreeEOS_hi = 10.1d0
       logRho_max_FreeEOS_lo = 10.0d0
 
-      logT_min_FreeEOS_hi = 3.1d0
-      logT_min_FreeEOS_lo = 3.0d0
+      logT_min_FreeEOS_hi = 3.12d0
+      logT_min_FreeEOS_lo = 3.02d0
```
VincentVanlaer, line 105: This reduces the region we use for FreeEOS. I could not find in your comments why this needs to happen.

eos/eosDT_builder/src/eos_regions_defs.dek
```diff
@@ -33,7 +34,7 @@
          double precision, parameter :: logRho4 = logQ2 + 2*logT6 - 12
 
          double precision, parameter :: logT_HELM   = 3d0  ! smallest logT for HELM
-         double precision, parameter :: logQmin = -8.2d0
+         double precision, parameter :: logQmin = -8.05d0
          double precision, parameter :: logQmax = logQ1
```
VincentVanlaer, line 38: Where does this actually affect things if you are computing with `opal_scvh_only = .true.`? `eos_regions_defs.dek` is only used by `eos_regions_code.dek` which is used only if all `_only` flags are disabled. It is also only used to determine whether to blend to HELM in the generated tables, and not whether we got to FreeEOS or not.

VincentVanlaer (commented):
I have gone through the entire PR now, and I have some confusion that would need to be resolved. As far as I can tell, this PR does the following:

- Fix ln/log10 problems (seems good)
- Fix buggy Z interpolation (seems good)
- Region updates to FreeEOS blending (unsure why this is needed)
- Forced blend to ideal (which should be moved to the normal blending code)
- Added fallbacks to HELM without radiation and ideal (which seemed unnecessary before, so I wonder why it is needed now). Note that this is only triggered if the previous cases report errors which seems unchanged in general.
- A fix to `grad_ad` and `gamma1` computations (I am not well versed enough in these computations to check)
- Build system fixes (thanks! :P)

Other modifications just don't appear to do anything from my reading of the code?

eos/defaults/eos.defaults
Debraheem: at logT = 3d0, FreeEOS actually fallsback to helm in the FreeEOS table. We shift the blend boundary up 0.02d0, so it begins at 3.02d0 where FreeEOS actually begins returning values.

PR #1000, opened 2026-05-18
approx21: fix he4 dydt spurious a2

Debraheem:
There was a small bug in the derivative for he4 in approx21. `a2 = y(ihe3) * y(ihe4) * rate(irhe3ag)` is added twice on accident. This is just a derivative, so at most it would mildly affect convergence on the pre-ms when there is he3 available? Otherwise I think this is a relatively inconsequential bug.

### DIFF
--- a/net/private/net_approx21.f90
+++ b/net/private/net_approx21.f90
@@ -1110,7 +1110,7 @@
    ! cno cycles
          a1 = y(io16) * y(ih1) * rate(iropg)
 
-         qray(ihe4) =  qray(ihe4) + a1 + a2
+         qray(ihe4) =  qray(ihe4) + a1
 
          if (.not. deriva) then
             a1 = y(in14) * y(ih1) * rate(ifa) * rate(irnpg)


fxt44:
i'm not sure i agree with this. first, i think you are referring to the right hand side, the ODEs themselves. what seems to be missing is 

      a2 = -y(ihe4) * y(in14) * rate(irnag) * 1.5d0

then the original line 

      qray(ihe4) =  qray(ihe4) + a1 + a2

is correct. i'll look deeper at it later today.

fxt44:
i compared 26.4.1 with mine.  

the reaction being modeled is 14N(a,y)18F(e+ nu)18O(a,g)22Ne, important for being the first time a star goes neutron rich and for being a key source of neutrino emission. approx21 doesn't know about 18F, 18O, or 22Ne. it models this conversion as 14N + 1.5 * a -> 20Ne.

in mine, i put this reaction under a cno banner 

      a1 = y(io16) * y(ih1) * rate(iropg)
      a2 = -y(ihe4) * y(in14) * rate(irnag) * 1.5d0
      qray(ihe4) =  qray(ihe4) + a1 + a2

in mesa, this reaction was moved to be under the heavy ion reactions banner

         a4 = -y(ihe4) * y(in14) * rate(irnag) * 1.5d0  ! n14 + 1.5 alpha => ne20
         qray(ihe4) =  qray(ihe4) + a1 + a2 + a3 + a4

but in doing so the original a1 + a2 above was not reduced to just a1

         a1 = y(io16) * y(ih1) * rate(iropg)
         qray(ihe4) =  qray(ihe4) + a1 + a2

so i now agree this is a bug in mesa. its been double adding the  ppchain reaction 3He(a,g)4He. i also agree it probably has a small impact, likely producing a little ne20 before its time.  good catch!

Debraheem:
glad to see you worked it out and came around :)

warrickball:
`bluebear` has reported [all tests succeeded](https://testhub.mesastar.org/main/commits/5065b6d7eaae61388b63cad4f2ac4278ca0731f6).

Debraheem:
> `bluebear` has reported [all tests succeeded](https://testhub.mesastar.org/main/commits/5065b6d7eaae61388b63cad4f2ac4278ca0731f6).

thanks for testing this!

PR #1002, opened 2026-05-19
Revert "Minor Bugfix in Plasmon Neutrino Cooling Rates"

Debraheem:
Reverts MESAHub/mesa#1001


Debraheem:
https://testhub.mesastar.org/bugfix%2Fneu_plasmon_cooling/commits/d4cdbac @mathren two failing test cases. Would not be suprised if the failure is due to a test case sensitivity. We can look closer and correct tomorrow.

PR #1003, opened 2026-05-19
Composite overshoot

matthiasfabry:
add functionality to model a step + exponential overshoot.

you can enable with:
```fortran
      overshoot_scheme(1) = 'step+exponential'
      overshoot_zone_type(1) = 'burn_H'
      overshoot_zone_loc(1) = 'core'
      overshoot_bdy_loc(1) = 'top'
      overshoot_f(1) = 0.335   ! step part, slightly less than brott
      overshoot_f2(1) = 0.01  ! exponential part, replace outer part of brott with this
      overshoot_f0(1) = 0.01
```


Debraheem:
I just cleaned up some of the white space diff so this touches less files, and tweaked the docs slightly. Now we just need testing https://testhub.mesastar.org/composite_overshoot/commits/27af2e7.

wmwolf:
Running on Hercules: https://testhub.mesastar.org/composite_overshoot/commits/a84c8d0. It'll take a while to get through things since it's running serially, though.

wmwolf:
[Nearly done](https://testhub.mesastar.org/composite_overshoot/commits/a84c8d0). Looks like `make_pre_ccsn_13bvn` is stalling... currently at over 10k models and `log_dt` ~ –10 in `inlist_cc`.

wmwolf:
Hmm.. the more I watch `make_pre_ccsn_13bvn`, the more concerned I am. It seems to have no smooth failure mode like a maximum model number. I'll let it keep going, but I think you can probably count it as a failure. The remaining missing tests are from `binary` and `astero`, and thus haven't even started yet, so I doubt there are issues with them.

Debraheem:
It definitely needs a max_model_number. I think you should kill the test, if it's going on indefinitely. I'm a little surprised it is failing thought, given this branch did not touch it.

Debraheem:
@wmwolf  https://testhub.mesastar.org/main/commits/aa27a08/test_cases/star/make_pre_ccsn_13bvn, it should only do ~4k timesteps. I can't see the log since it doesn't update testhub until the run is complete, but maybe this is helpful.

wmwolf:
It surprised me, too. I killed it and the remaining tests should roll in soon.

Debraheem:
I'm surprised the BVN test is even failing, since it is untouched by this pr except via the change to 
https://github.com/MESAHub/mesa/blob/a84c8d00ec6d1c00bb9a78607a72da96dd7c071b/star/private/overshoot_utils.f90#L219

My lack of sleep from last week has turned into an illness, but when i'm back up to speed, i'll investigate and try to fix this while i go through all the other branches from the hackathon.

wmwolf:
Feel better soon, @Debraheem! Hercules has finished testing. Everything passed, except for `make_pre_ccsn_13bvn`, which stalled at tiny timesteps and had to be manually killed.

warrickball:
I've just started this on `bluebear`, which does full tests. `make_pre_ccsn_13bvn` hasn't been presenting any issues in the full tests, so it might just be that an intermediate model needs updating.

warrickball:
All clear [on `bluebear`](https://testhub.mesastar.org/composite_overshoot/commits/a84c8d0).

Debraheem:
Would you be open to updating the mod files from your test? It’s okay if
they’re deleted, i can re run it locally!

-EbF

warrickball:
Incidentally, does this not also allow one to refactor pure step/exponential overshoot into special cases of step+exponential? As in, is step overshoot the same as step+exp with the exponential coefficients set to zero? (I haven't yet had a chance to look closely at the code.)

Debraheem:
> Incidentally, does this not also allow one to refactor pure step/exponential overshoot into special cases of step+exponential? As in, is step overshoot the same as step+exp with the exponential coefficients set to zero? (I haven't yet had a chance to look closely at the code.)

I think in principle, yes. It might be much simpler as well. Something like just turning on the exponential tail if using step and f2 is non-zero.

Debraheem:
@warrickball following your comments, I've refactored the implementation into the step overshoot routine, and significantly reduced the diff. Thanks for the feedback! I've also regenerated the mod files from the bvn case, so it should pass normal testing now.

Debraheem (approved):

warrickball:
Sorry that I missed the last refactor. It looks like this has broken [`make_pre_ccsn_13bvn` on `bluebear`](https://testhub.mesastar.org/main/test_cases/star/make_pre_ccsn_13bvn). It was finishing everything within about 4000 steps but is now getting stopped by the new `max_model_number` of 6000.

I'm a bit strapped for MESA time for the near future but I'll try to have a quick look at the diff.

warrickball:
I suspect that the fact that this MR also makes the overshooting [use `s% alpha_mlt(k)` instead of `s% mixing_length_alpha` in `overshoot_utils.f90`](https://github.com/MESAHub/mesa/pull/1003/changes#diff-12023a476acec1108208fdf56258c281a12ebc7a9f3391aa2d09236a1dae3a9fL219) is the change that's leading to a failure. I don't know this test well enough to comment further right now.

Debraheem:
@warrickball , no worries, yes the bugfix to the overshooting length scale is likely the reason, but this test is just very very sensitive. In fact when I ran it locally on my arm mac and regenerated the mod files on my last refactor, it passed and finished, so it's quite surprising that it's now failing on bluebear after i regenerated the mod files. I'll look into it!

matthiasfabry:
the change to `s% alpha_mlt(k)` was because some test cases modify `alpha` in `run_star_extras`, which, when sticking with `mixing_length_alpha` would produce inconsistent overshooting lengths.

Debraheem:
exactly, and `make_pre_ccsn_13BVN` uses the `other_alpha_mlt` routine, so it is directly affected.

Debraheem:
I believe I know what happened. I did not merge main into the branch, so the when I tested the branch it was fine, but the last 15 commits on main perturbed the model slightly, including my approx21 bug fix to a derivative. The BVN model with the adjusted other_alpha_mlt routine now undergoes a shell merger at core collapse which makes the model stall. I'm going to fix the test case in a separate branch.

PR #1025, opened 2026-06-03
Add use const_def to eos_plotter.f90

sunnywong314:
Needs constants like crad and ln10


Debraheem:
Thanks sunny, I've actually fixed this inside of https://github.com/MESAHub/mesa/pull/997, but I'll merge your branch into there too, so your commit shows your contribution!

PR #1027, opened 2026-06-10
Pgplot residuals

matthiasfabry:
Based on @mathren initial work of bringing some diagnostic residual plots.

I had to refactor quite a lot to allow for our typical user controls.

I still need to do a few things, such as fix file printing, and possible allow for restarts.

Also, I'm not quite sure which branch we'd want to merge this into, because its based on the hackathon inlists. Maybe we want to rebase the top commits onto main first?


mathren:
Thanks @matthiasfabry! Is there a strong dependence on the inlist parameters? I'd expect only the `use_other_pgstar_plot=.true.`  to matter, so inlists shouldn't really matter for which branch to merge this on...

matthiasfabry:
I'm not sure what you mean?
I meant that this PR shouldn't be merging the hackathon massive star test cases into main.

mathren:
ok, sorry, I misunderstood. Indeed, I would maybe base this on main rather than the hackathon branch, and make it independent of the hackathon branch completely since nothing of the hackathon is necessary for these pgplots?

matthiasfabry:
That's right, I'll see if can do the rebasing.

@rhdtownsend I cannot get saving to files to work because apparently, in the PNG device driver, `pgimag` will always use a fixed color palette. LLMs suggest updating the pgplot libs that come with the SDK. Is this possible?

matthiasfabry:
I cherry-picked the relevant commits on top of main; making a new PR:

matthiasfabry:
closing to other PR with cherry picked commits #1028

PR #1028, opened 2026-06-10
Pgstar residuals take 2

matthiasfabry:
I cherry picked only the relevant commits (removing the hackathon changes)

See #1027 for original.

What remains to be done is checking if the device drivers can be updated and if that fixes plotting to pngs.


mathren:
@matthiasfabry regarding the color palette (from the other now closed PR), maybe it's easier to use the existing color palette (at least for the max residual across cells for each equation, not sure this would work easily for the colorbar)?
Those colors where picked by the LLM and are not even particularly good, so feel free to change them if it leads to issues with saving files

PR #1030, opened 2026-06-12
Remove copying binary history to star history

matthiasfabry:
I consider this a "legacy" setting due to how binary history needed to be plotted with `pgstar`.

I argue defaulting the copying/appending to false is a not-too-invasive change. Users will either need to reenable this in the inlists, or modify their data-analysis scripts that read in binary history data.

I also took the time to set up pgbinary use in the remaining (science) binary test_suite cases (creating a rough template for single star use in pgbinary).


Debraheem:
I think this branch can be merged after testing, however my concern is that users often do plot binary parameters in pgstars instead of pgbinary, and might not see this control in the future. Should we really default to off? I don't mind either way.

matthiasfabry:
I guess it comes down to opinion. I'm obviously very biased to using `pgbinary` over `pgstar`. I fact, I argue there's nothing pgbinary can't do that pgstar can (apart from plotting a binary column alongside a star column on the same history plot).

I feel changing the default is a good compromise between making MESA less bulky to the everyday and new user, and removing the capability altogether (which would break the power user's scripts and plots)

Debraheem:
passes testing, https://testhub.mesastar.org/remove-copying/commits/head can i merge?

PR #1032, opened 2026-06-18
Fix synthetic filter interpolation outside passbands

nialljmiller:
This fixes synthetic photometry for filters interpolated onto SED wavelength grids.

Previously, interpolated filter transmission could remain nonzero outside the tabulated filter wavelength range. 
This produced out-of-band flux leakage for hot/blue SEDs, especially WD and blackbody models, causing unphysical synthetic colors. I found this testing WD cooling tracks. 

The fix clamps interpolated filter transmission to zero outside the original filter support and clips negative interpolated transmission values to zero. 

Here is the function that has been added to `colors/private/synthetic.f90`  --

```
      fmin = minval(filt_wave)
      fmax = maxval(filt_wave)

      do i = 1, size(target_wave)
         if (target_wave(i) < fmin .or. target_wave(i) > fmax) then
            filt_on_target(i) = 0.0_dp
         else if (filt_on_target(i) < 0.0_dp) then
            filt_on_target(i) = 0.0_dp
         end if
      end do

```


Debraheem:
Nice find! I can't look at my computer for another couple days because im moving apts, just an fyi! But this is awesome!

Is there a reason the bolometric flux tests didn't capture this in the unit tests, maybe because i used too low log G and low teff? Do we need a high logG hot Teff bb flux unit test added to the others, so this would be caught? 

Is there a particular place where this comes? Perhaps we add a minor changelog entry for this to mark it and tweak the unit tests slightly to capture this scenario?

nialljmiller:
I have made this a draft while I verify the fix works across a few grids. Currently, the koester grid is giving a large offset at unpredictable points.

nialljmiller:
## Summary

This PR fixes numerical issues in the Colors synthetic-photometry machinery exposed by white-dwarf cooling tracks and hot/blue SED tests.

The main bug was that interpolated filter transmission curves could remain nonzero outside the tabulated filter passband. This allowed out-of-band flux to leak into synthetic magnitudes, especially for hot SEDs with substantial blue/UV flux. The fix enforces compact filter support after interpolation: transmission is set to zero outside the original filter wavelength range, and negative interpolated transmission values are clipped to zero. This is applied both to the model-star convolution and to the Vega zero-point calculation.

The PR also makes the SED interpolation machinery more robust. It adds a bounded Hermite interpolation path, fixes Hermite behavior for singleton grid axes, and simplifies the linear interpolation path so degenerate axes and grid-edge cases are handled consistently.

## Main changes

- Added `zero_filter_outside_support` in `colors/private/synthetic.f90`.
  This prevents both model-SED and Vega/reference-SED flux leakage outside the tabulated filter passband.

- Added `colors/private/hermite_interp_bounded.f90`.
  This provides `construct_sed_hermite_bounded`, supports both preloaded flux cubes and individual SED-file fallback, vectorizes interpolation over wavelength, and uses a local multilinear SED as a safety reference.

- Added bounded-Hermite safeguards.
  Hermite remains the preferred result, but invalid, nonpositive, NaN, or excessive local overshoot values fall back to multilinear interpolation. A whole-SED flux-scale check also falls back to multilinear interpolation when the Hermite SED differs too strongly from the multilinear reference.

- Fixed singleton-axis handling in both `hermite_interp.f90` and `hermite_interp_bounded.f90`.
  Axes with one grid point are now treated as valid degenerate dimensions, with zero interpolation extent and zero derivative contribution, rather than as out-of-grid failures. This is important for grids such as Koester white-dwarf atmospheres with a single metallicity axis.

- Made `bolometric.f90` default to `Hermite_bounded`.
  The default fallback also now uses bounded Hermite. `calculate_bolometric_phot` now cleans a local flux copy before integration instead of mutating the input SED arrays.

- Simplified `linear_interp.f90`.
  The old explicit 8-corner/log-space path was replaced with a shared `setup_axis` helper and a dimension-agnostic weighted loop. This handles singleton axes, edge clamping, and interpolation weights consistently.

- Updates `colors/Makefile` to compile `hermite_interp_bounded.f90`.

- Updated `star/test_suite/custom_colors/python_helpers/test_timings.py`.
  Removed the obsolete Newton-step Colors configurations and adjusted the timing benchmark/feature-breakdown plots to the remaining supported configurations.

- Updates `colors/README.rst` to include comments on interpolation issues.

colors/private/bolometric.f90
```diff
@@ -23,13 +23,25 @@ module bolometric
...
+   public :: calculate_bolometric, calculate_bolometric_phot
+
+
+
+   ! IAU 2015 Resolution B2 zero-point for the bolometric magnitude scale:
+   ! the irradiance corresponding to M_bol = 0. Defined so that a source with
+   ! L = L_sun observed from 10 pc has M_bol = 4.74. This is a fixed constant
+   ! and is unrelated to the Vega/AB/ST conventions in synthetic.f90, which
+   ! apply to bandpass magnitudes only.
+   real(dp), parameter :: f_bol_zero_point = 2.518d-5   ! erg s^-1 cm^-2
```
VincentVanlaer: This should go in `const/public/const_def.f90`. We have a bunch of these resolution defined constants there.

colors/private/bolometric.f90
```diff
@@ -63,28 +75,37 @@ subroutine calculate_bolometric(rq, teff, log_g, metallicity, R, d, bolometric_m
...
+      do i = 1, size(wavelengths)
+         if (wavelengths(i) <= 0.0d0 .or. clean_fluxes(i) < 0.0d0) then
+            clean_fluxes(i) = 0.0d0
          end if
       end do
 
-      call simpson_integration(wavelengths, fluxes, bolometric_flux)
+      call simpson_integration(wavelengths, clean_fluxes, bolometric_flux)
+
+      deallocate (clean_fluxes)
```
VincentVanlaer: Fortran will automatically deallocate, so this is not needed.

colors/private/bolometric.f90
```diff
@@ -99,11 +120,15 @@ end subroutine calculate_bolometric_phot
 
    real(dp) function flux_to_magnitude(flux)
       real(dp), intent(in) :: flux
+      logical, save :: warned = .false.
       if (flux <= 0.0d0) then
-         print *, "Error: Flux must be positive to calculate magnitude."
+         if (.not. warned) then
+            print *, "colors: non-positive bolometric flux; Mag_bol set to 99"
+            warned = .true.
+         end if
```
VincentVanlaer: Shouldn't flux be always positive since `bolometric_flux` in the function above is already checked for negative values?

PR #1038, opened 2026-07-31
keep failed test output

### LINKED ISSUE #1017: When a test fails, keep the check_results file
During the build process, when a test fails (via diff finding different output), the check_results file is deleted by make (which regards it as an intermediate file). Changing this behavior so that check_results is kept would be helpful in debugging test failures. I think this can be done using the .PRECIOUS target in the makefile.

VincentVanlaer:
Fixes #1017


Debraheem:
This looks fine to me, thanks!

PR #1040, opened 2026-08-09
Update GitHub action versions

warrickball:
While tinkering with re-using the action to install the SDK under Linux, I noticed we've been getting various warnings about deprecated versions of Node.js in actions (e.g. [this action run](https://github.com/MESAHub/mesa/actions/runs/30671411764)). I think the actual problematic action is `checkout@v4` but this PR goes ahead and updates all the actions to their latest major versions.

There might be reasons we don't want the latest versions, in which case I'm to roll back one or two.

### DIFF
--- a/.github/actions/install-mesa/action.yml
+++ b/.github/actions/install-mesa/action.yml
@@ -34,7 +34,7 @@ runs:
       shell: bash
 
     - name: Restore LFS cache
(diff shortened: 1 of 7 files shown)


VincentVanlaer:
Seems at least no issues have popped up for the checks on this PR. Let's see if anything breaks on main.

PR #1007, opened 2026-05-21
Bugfix/neu plasmon cooling2

Debraheem:
See https://github.com/MESAHub/mesa/pull/1001. This commit needs testing, and potential test_suite fixes before it can be merged.

PR #1041, opened 2026-08-10
Fix tidal deformation omega on restart and remesh

Debraheem:
I think the photo restart issue in https://github.com/MESAHub/mesa/pull/1007 comes from two related problems in models using tidal deformation.

With tidal deformation, the moment of inertia used by MESA is

```math
i_{\rm rot}(k)
=
f_{\rm tide}(k)\,i_{\rm rot,tidal}(k)
+
\left[1-f_{\rm tide}(k)\right]i_{\rm rot,single}(k).
```

Here, `f_tide` depends on how closely the shell is rotating at the synchronous angular velocity. By default,

```math
\begin{aligned}
\omega_{\rm sync} &= \frac{2\pi}{P_{\rm orb}}, \\[4pt]
f_{\rm sync} &=
\min\left(
\frac{|\omega|}{\omega_{\rm sync}},
\frac{\omega_{\rm sync}}{|\omega|}
\right), \\[4pt]
f_{\rm tide} &=
\frac{1}{
1+\exp\left[-w\left(f_{\rm sync}-f_0\right)\right]
}.
\end{aligned}
```

Here, `w` is `f_sync_switch_width` and `f_0` is `f_sync_switch_from_rot_defor`.

1. When restarting from a photo, `finish_load_model` is called before the binary tidal-deformation hooks are attached. The call to `use_xh_to_update_i_rot` therefore calculates only `i_rot,single`, rather than the blended `i_rot` given above. MESA then replaces the `omega` stored in the photo with

```math
\omega = \frac{j_{\rm rot}}{i_{\rm rot,single}}.
```

This changes the rotation state because the model before the restart used the blended tidal moment of inertia. The photo stores `omega`, `j_rot`, and `w_div_w_crit_roche`, but it does not store `i_rot` directly. On a photo restart, MESA now keeps the stored `omega` and `w_div_w_crit_roche` and recovers the effective moment of inertia from

```math
i_{\rm rot} = \frac{j_{\rm rot}}{\omega}.
```

This recovers the blended `i_rot` represented by the stored `j_rot` and `omega`, even though the tidal hooks have not yet been attached. Loading a `.mod` file keeps the existing behavior because saved models do not contain the complete state of the model.

2. During mesh adjustment, the tidal hooks are attached and MESA does calculate the blended `i_rot`. However, the old code reconstructed `j_rot` on the new mesh and evaluated `i_rot` before setting `omega` on that mesh. Since `f_tide` depends on `omega`, the tidal blend read an unset or leftover value of `omega`. This could produce different `i_rot` and `omega` values in uninterrupted and restarted runs.

MESA now reconstructs the old `omega` over the same cell overlaps used for `j_rot` and sets this temporary value before evaluating the tidal blend. It then calculates the final value from

```math
\omega = \frac{j_{\rm rot}}{i_{\rm rot}},
```

where `i_rot` is the blended tidal moment of inertia. Angular momentum remains the conserved quantity during remeshing. The reconstructed `omega` is used only to evaluate `f_tide` with a defined value.

Together, these changes preserve the blended tidal rotation state stored in a photo and prevent tidal remeshing from evaluating `i_rot` with an unset value of `omega`.

### DIFF
--- a/star/private/mesh_adjust.f90
+++ b/star/private/mesh_adjust.f90
@@ -231,7 +231,7 @@
 
          if (s% rotation_flag) then
             call adjust_omega(s, nz, nz_old, comes_from, &
-               xq_old, xq, dq_old, dq, xh, j_rot_old, &
+               xq_old, xq, dq_old, dq, xh, j_rot_old, omega_old, &
                xout_old, xout_new, dqbar_old, dqbar, ierr)
             if (failed('adjust_omega')) return
             if (s% D_omega_flag) then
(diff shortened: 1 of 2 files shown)


Debraheem:
testing passed, so i am going to merge this branch unless there are any concerns.

PR #1044, opened 2026-08-13
test reduced mass_fraction_limit

Debraheem:


evbauer:
It will be interesting to see what test timing looks like here, particularly for massive stars and larger nets. I think this could make the Skye EOS calls potentially 1-2 orders of magnitude more expensive, so at least some tests might show a noticeable slowdown. Have you done any local testing with timings reported?

I think there was some discussion about pushing this lower when Skye was being developed, but I can't remember specifically why we didn't go lower than 1d-4. Maybe a conservative choice. I think this was inherited from an older default `mass_fraction_limit_for_PC = 1d-3`, and we were already pushing an order of magnitude lower than that.

Debraheem:
I haven't noticed any timing difference, and i've been doing this locally for a quite a while. I think "fix_d_eos_dxa_partials = .true." is the real costly operation at the moment. But let's keep an eye on those tests!

PR #1059, opened 2026-09-03
Fix rotation and ST time smoothing state on restart

Debraheem:
Photo restarts could use an uninitialized w_div_w_crit_roche while reconstructing i_rot. They could also discard mstar_old and the saved ST coefficients before the first post-restart remeshing event.

The solution is to Initialize the rotation geometry before reconstructing i_rot, save mstar_old in photos, and preserve the ST time smoothing inputs through the first new_generation. Bump the star photo version to 19. Part of the solution uses the photo restart fixes to ST taken from https://github.com/MESAHub/mesa/pull/553/changes (originally found by @rjfarmer)

The three affected star test cases pass their restart and checksum checks with FPE checks enabled:
- accreted_material_j
- check_redo
- high_rot_darkening

The ST restart state was also verified with double_bh through a model-700 restart, but it also revealed a second fpe issue in the binary Roche potential routines. That problem is specific to nans and -infinity values inside the Roche potential tables and deserves its own separate pr.

### DIFF
--- a/star/private/alloc.f90
+++ b/star/private/alloc.f90
@@ -524,6 +524,12 @@
                if (failed('mlt_vc')) exit
                call do1(s% conv_vel, c% conv_vel)
                if (failed('conv_vel')) exit
+
+               ! These are persistent inputs needed for ST time smoothing.
+               call do1(s% D_ST_start, c% D_ST_start)
+               if (failed('D_ST_start')) exit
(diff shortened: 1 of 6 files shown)


Debraheem:
the double_bh fpe failure on my machine is a known issue related to fpe checks complaining about the nans in the roche tables. Otherwise, the tests are passing and the fpe checks passed on BOSE.

